Issue #25, opened 2018-03-26
Integrate with TravisCI


nishanth-kd:
Need a test clearance before the reviews. 

Capture our deployment steps here and then hook it with travis?

nishanth-kd:
#14

rathorevaibhav:
This is complete. Closing.

PR #27, opened 2018-03-26
(#25) testing

rathorevaibhav:
PR #25


rathorevaibhav:
@nishanth-kd Travis has been integrated. Right now, we just have a sample unit test case. I think we can add unit test cases in separate trackings. TravisCI is complete.

Travis CI URL:
https://travis-ci.org/ColoredCow/employee-portal

kd-csp:
@rathorevaibhav :+1:

.env.travis
```diff
@@ -0,0 +1,11 @@
+APP_ENV=testing
+APP_KEY=SomeRandomString
```
kd-csp: not the expert but is it supposed to be this?

tests/Unit/UserTest.php
```diff
@@ -0,0 +1,20 @@
...
+class UserTest extends TestCase
+{
+    /**
+     * A basic test example.
+     *
+     * @return void
+     */
+    public function testExample()
+    {
+        $this->assertTrue(true);
```
kd-csp: 👍

.env.travis
rathorevaibhav: We can leave it empty. The travis.yml contains `php artisan key:generate` so, we can make it just like .env.example

Issue #32, opened 2018-03-30
Finance module


rathorevaibhav:
1. Clients
2. Projects
3. Invoices

rathorevaibhav:
The foundation is complete. Targeting more detailed requirements separately. Closing.

PR #33, opened 2018-03-30
(#32) finance client, project, invoices stubs

rathorevaibhav:
migrations and stubs for models, controllers


rathorevaibhav:
@nishanth-kd PR for defining models, migrations, Controller stubs for the Finance Module. Ready for review.

app/Models/Finance/Invoice.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace App\Models\Finance;
+
+use Illuminate\Database\Eloquent\Model;
+
+class Invoice extends Model
+{
+    protected $table = 'finance_invoices';
+
+    protected $fillable = ['name', 'project_id', 'review_value', 'status', 'sent_on', 'sent_to', 'paid_on', 'file_path'];
```
nishanth-kd: I have a doubt on what should be fillable and what shouldn't on this one. @rudresh can you share some thoughts on this.

app/Models/Project.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace App\Models;
+
+use App\Models\Client;
+use Illuminate\Database\Eloquent\Model;
+
+class Project extends Model
+{
+    $fillable = ['name', 'started_on', 'invoice_email', 'status'];
```
nishanth-kd: Vaibhav the project_id and client_ids should be in the $gaurded array to avoid accidental updates.

app/Models/Finance/Invoice.php
```diff
@@ -0,0 +1,22 @@
...
+namespace App\Models\Finance;
+
+use App\Models\Project;
+use Illuminate\Database\Eloquent\Model;
+
+class Invoice extends Model
+{
+    protected $table = 'finance_invoices';
+
+    protected $fillable = ['name', 'project_id', 'project_invoice_id', 'review_value', 'status', 'sent_on', 'paid_on', 'file_path'];
```
nishanth-kd: project_invoice_id and project_id should be in gaurded.

app/Models/Project.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace App\Models;
+
+use App\Models\Client;
+use Illuminate\Database\Eloquent\Model;
+
+class Project extends Model
+{
+    $fillable = ['name', 'client_id', 'project_client_id', 'started_on', 'invoice_email', 'status'];
```
nishanth-kd: project_client id in gaurded. let's not keep this mass assignable.

nishanth-kd: Once created these ID would not be replaced or changed either

app/Models/Finance/Invoice.php
rathorevaibhav: Tracking this in #34

app/Models/Project.php
rathorevaibhav: Tracking this in #34

PR #35, opened 2018-03-31
32 finance views

rathorevaibhav:


app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -14,28 +19,44 @@ class InvoiceController extends Controller
...
      * Show the form for creating a new resource.
      *
-     * @return void
+     * @return \Illuminate\View\View
      */
     public function create()
     {
-        //
+        return view('finance.invoice.create')->with([
+            'projects' => Project::all(),
```
nishanth-kd: Do we need all here ? Or is there some kind of pagination that is supposed to be in place.

app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -14,28 +19,44 @@ class InvoiceController extends Controller
      */
     public function index()
     {
-        //
+        return view('finance.invoice.index')->with([
+            'invoices' => Invoice::with('project')->orderBy('sent_on', 'desc')->get(),
```
nishanth-kd: Same question on pagination. Am I putting a limit on how much data should be loaded, I think we should strategise on that.

app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -14,28 +19,44 @@ class InvoiceController extends Controller
...
     {
-        //
+        $validated = $request->validated();
+        $path = self::upload($validated['invoice_file']);
+        $invoice = Invoice::create([
+            'name' => $validated['name'],
+            'project_id' => $validated['project_id'],
+            'project_invoice_id' => $validated['project_invoice_id'],
+            'status' => $validated['status'],
+            'sent_on' => date("Y-m-d", strtotime($validated['sent_on'])),
```
nishanth-kd: Can we use Carbon as a standard in this application.
The problem with Date formats being hardcoded is that they are subject to change and can often be changed selectively. 

I would suggest maybe something like `(new Carbon('first day of January 2016'))->toFormattedDateString();`

Also if we're following certain date constants then define them in `config`. 
https://scotch.io/tutorials/easier-datetime-in-laravel-and-php-with-carbon#toc-formatting

app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -82,4 +114,30 @@ public function destroy(Invoice $invoice)
...
+
+    /**
+     * Upload invoice file.
+     *
+     * @param  \Illuminate\Http\UploadedFile  $file
+     * @return string
+     */
+    protected static function upload(UploadedFile $file)
+    {
+        $dir = date('Y') . '/' . date('m');
```
nishanth-kd: please put the logic to derive directory names in a separate method.

app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -82,4 +114,30 @@ public function destroy(Invoice $invoice)
...
+
+    /**
+     * Download invoice file.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return Symfony\Component\HttpFoundation\StreamedResponse
+     */
+    public function download($year, $month, $file)
+    {
+        $file_path = $year . '/' . $month . '/' . $file;
```
nishanth-kd: umm same for this path. and make it `funciton download($path)` if it shouldn't be of type `Request`.

app/Models/Client.php
```diff
@@ -7,7 +7,7 @@
 
 class Client extends Model
 {
-    $fillable = ['name', 'email', 'phone', 'address'];
+    protected $fillable = ['name', 'email', 'phone', 'address'];
```
nishanth-kd: 😆

app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -82,4 +114,30 @@ public function destroy(Invoice $invoice)
...
+    protected static function upload(UploadedFile $file)
+    {
+        $dir = date('Y') . '/' . date('m');
+        return $file->store($dir);
+    }
+
+    /**
+     * Download invoice file.
+     *
+     * @param  \Illuminate\Http\Request  $request
```
nishanth-kd: I see the DocBlock is reflecting a different `@param` from what is mentioned in the code.  

Should we still be in favour of using comments and duplicating information? @rathorevaibhav 

I personally see it as duplicate information that increases the `Cost of Change`. 

Hoping to resolve this soon before the codebase expands.

app/Http/Controllers/Finance/InvoiceController.php
rathorevaibhav: This is for the input field. to select the which project we're creating invoice for.

nishanth-kd: Oh ok, can we send selective information then instead of everything, I think only an ID and name is required here.

app/Http/Controllers/Finance/InvoiceController.php
nishanth-kd: send fields that we need, not all.

app/Http/Controllers/Finance/InvoiceController.php
rathorevaibhav: agreed.

app/Http/Controllers/Finance/InvoiceController.php
rathorevaibhav: tracking the pagination here #37

app/Http/Controllers/Finance/InvoiceController.php
rathorevaibhav: We're using year and month directory structure to store files in `storage/app` directory. Upon saving an invoice, the file_path gets stored is like `2018/04/abcd.pdf`. Now when this path is added to a URL (even after encoding), Laravel returns with a 404 not found. This is basically due to the slashes present in the file path. That's why I created a custom route with `invoices/download/{year}/{month}/{file}` for the download invoice.

Couldn't find any other way to resolve this. @nishanth-kd Let me know your thoughts.

app/Http/Controllers/Finance/InvoiceController.php
rathorevaibhav: @nishanth-kd I agree that the best code is self-documenting. However, since this is an open source project, I don't think we can assume that the other developers looking at the code will be familiar with the context in which it was written. I agree on increasing cost of change but really think comments make understanding easier for someone new to the code base. The cost of induction will be less. 

The target is to provide more information through the comments than what we can be retrieved from method signatures.

On the comments with mismatching information, I am taking care of them wherever I can find. This was a miss. Thanks for pointing out. Corrected.

app/Http/Controllers/Finance/InvoiceController.php
nishanth-kd: I think I'm fine with this. but the logic to derive file paths would be separate method

rathorevaibhav: yeah. did that. have a look.

PR #36, opened 2018-04-02
32 client project views

rathorevaibhav:
Changes include:

1. As per requirements, invoices don't need a name. Name was added to primarily describe the invoice as some projects can have different billing cycles (weekly, fortnightly, monthly). Name has replace by comments, if someone has to add any invoice meta.


app/Http/Controllers/ProjectController.php
```diff
@@ -53,23 +70,35 @@ public function show(Project $project)
...
      * Update the specified resource in storage.
      *
-     * @param  \Illuminate\Http\Request  $request
+     * @param  \App\Http\Requests\ProjectRequest  $request
      * @param  \App\Models\Project  $project
-     * @return void
+     * @return \Illuminate\Http\RedirectResponse
      */
-    public function update(Request $request, Project $project)
+    public function update(ProjectRequest $request, Project $project)
```
nishanth-kd: This is interesting. How does the Route for this update method look ? I can pass two objects to the controller via a PATCH request ?

.gitignore
```diff
@@ -17,3 +17,4 @@ yarn-error.log
 /public/build
 /public/css
 /public/js
+/public/images
```
nishanth-kd: What kind of images are we storing here ?

rathorevaibhav: oh, when I added jquery-ui for datepicker, it added some library assets, in the public folder. We're not storing anything here for now.

app/Helpers/DateHelper.php
```diff
@@ -0,0 +1,19 @@
...
+use Carbon\Carbon;
+
+if (!function_exists('format_date_to_save'))
+{
+	/**
+	 * Formats date string in 'd/m/Y' format to 'Y-m-d' format for storage
+	 *
+	 * @param  string $date
+	 * @return string
+	 */
```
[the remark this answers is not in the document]
rathorevaibhav: sure. I'll make changes.

app/Http/Controllers/ProjectController.php
rathorevaibhav: The route is a normal Resource route. `Route::resource('projects', 'ProjectController')`

Generally, when creating a controller using artisan, the default params for update are `(Request $request, $id)`. But when I ran `php artisan make:model Models/Project -mcr`, it replaced the $id param with Project object. Not very sure of how it's doing behind the hood, but looks interesting to dig into it.

rathorevaibhav:
@nishanth-kd changes applied. Have a look.

Issue #20, opened 2018-03-23
Server side validations


rathorevaibhav:
1. Need to add validation for current requests.
- [x] All Job requests
- [x] All Applicant requests

rathorevaibhav:
Done for both. Closing.

PR #42, opened 2018-04-03
(#20) hr form validations

rathorevaibhav:
PR for #20. Also includes some code refactoring.

### DIFF
--- a/app/Events/HR/JobCreated.php
+++ b/app/Events/HR/JobCreated.php
@@ -2,25 +2,25 @@
 
 namespace App\Events\HR;
 
-use App\Models\HR\Applicant;
-use Illuminate\Broadcasting\InteractsWithSockets;
-use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Broadcasting\Channel;
 use Illuminate\Queue\SerializesModels;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Broadcasting\InteractsWithSockets;
 
-class ApplicantCreated
+class JobCreated
 {
     use Dispatchable, InteractsWithSockets, SerializesModels;
 
-    public $applicant;
+    public $job;
 
     /**
      * Create a new event instance.
      *
      * @return void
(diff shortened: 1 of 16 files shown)


app/Events/HR/JobCreated.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace App\Events\HR;
+
+use App\Models\HR\Job;
+use Illuminate\Broadcasting\Channel;
+use Illuminate\Broadcasting\InteractsWithSockets;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Queue\SerializesModels;
+
+class JobCreated
```
nishanth-kd: Are these model events ? I see that the it's for the Model Job. 

Can you use the provision for Model events in Laravel. I think we won't need to create an Event separately for each of the model events. 

Check this out 
https://laravel.com/docs/5.6/eloquent#events

routes/api.php
```diff
@@ -18,5 +18,5 @@
 });
 
 
-Route::post('hr/jobs/create', 'HR\JobController@create');
-Route::post('hr/applicants/create', 'HR\ApplicantController@create');
+Route::post('hr/jobs/create', 'HR\JobController@store');
```
nishanth-kd: Hey vaibahv why do we need to define this route ? Can't we use a resource route.

nishanth-kd: yea i would suggest the latter because otherwise, i'm mapping create with the store method which is confusing in the Laravel and RESTfull API resources context

rathorevaibhav: ok. So I'll proceed with `Route:resource('hr/jobs', 'HR\JobController')->only(['store'])`.

nishanth-kd: 👍

app/Events/HR/JobCreated.php
nishanth-kd: Yea I think, did some research. We might just have to create events for passing this data.

Issue #44, opened 2018-04-04
Add weeklydose


rathorevaibhav:

rathorevaibhav:
@PrateekNarang This is ready for testing. Have a look.

rathorevaibhav:
This is complete and in use. Will track all other WeeklyDose related items separately.

PR #45, opened 2018-04-04
(#44) weeklydose

rathorevaibhav:
PR for #44

### DIFF
--- a/app/Http/Controllers/WeeklyDoseController.php
+++ b/app/Http/Controllers/WeeklyDoseController.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Http\Requests\WeeklyDoseRequest;
+use App\Models\WeeklyDose;
+
+class WeeklyDoseController extends Controller
+{
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\View\View
+     */
+    public function index()
+    {
+        return view('weeklydose')->with([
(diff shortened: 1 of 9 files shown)


rathorevaibhav:
Putting this on hold. Shifting to #55

rathorevaibhav:
@nishanth-kd We just migrated the WeeklyDose module in Employee Portal. There's no change in requirements as such. The only thing was user had to go to different URLs to access Employee Portal and WeeklyDose. Now, WeeklyDose is a module within Employee Portal.

The chrome plugin will now call a different URL that it used to. That is made configurable.

In the weeklydose listing page, I've also added a help guide for plugin setup.

resources/assets/js/app.js
```diff
@@ -68,3 +68,24 @@ function getProjectList(projects) {
 	}
 	return html;
 }
+
+$('#copy_weeklydose_service_url').tooltip({
```
nishanth-kd: check the formatting here

nishanth-kd:
Can you elaborate on migrated the module to employee portal. I see a migration and routes for weekly dose in the code. If we built it separately why are we writing this code again? Shouldn't all this be pulled from the composer package?

resources/assets/js/app.js
rathorevaibhav: Has been corrected in #66

nishanth-kd:
I think if the same code is being used in the service as well as the portal. It should've been packaged.

Issue #48, opened 2018-04-04
Notification Missing


PrateekNarang:
Once I create Project, I get the project creation step back. I don't get any notification or indication that my project has been created successfully. Although I see, See All Project, I need to click it to see the project.

rathorevaibhav:
This is complete. There's a success alert box once an entity is created/updated.

PR #102, opened 2018-04-11
(#48) session status

rathorevaibhav:
#48 

Changes include:

1. An alert box for success statuses whenever an entity is updated.
2. This box comes with a cross button that will hide the box with a fade-out animation.
3. Once an entity is created, the user is redirected to it's edit page with the success message.


resources/views/project/edit.blade.php
```diff
@@ -8,6 +8,7 @@
     <a class="btn btn-info" href="/projects">See all projects</a>
     <br><br>
     @include('errors', ['errors' => $errors->all()])
+    @include('status')
```
nishanth-kd: can error be merged into status ?

rathorevaibhav: Agreed. Will be cleaner.

rathorevaibhav:
@nishanth-kd done with the feedback. have a look.

Issue #105, opened 2018-04-12
Rich text editor for sending emails


rathorevaibhav:
When sending emails to an applicant after a round, the send mail button pops up a modal with a normal text box to add body. We need to change this to a rich text editor so that it gives the flexibility to add bold characters, lists, italics, etc.

PR #110, opened 2018-04-12
(#105) editor

rathorevaibhav:
#105 

Changes include:

1. Added TinyMCE as a rich text editor for mails
2. Using Mailable classes and views to send HTML emails


public/src/tinymce/js/tinymce/plugins/codesample/css/prism.css
```diff
@@ -0,0 +1,138 @@
+/* http://prismjs.com/download.html?themes=prism&languages=markup+css+clike+javascript */
```
nishanth-kd: Can these files be pulled in via package.json rather than manual ?

rathorevaibhav: I tried using npm but tinymce didn't work using `require('tinymce')`. A lot of more dependencies need to be added and the editor was not working in some cases. Moreover, for plugins ability like lists, auto-links etc, their individual plugin files had to be added separately to the npm route.

PR #112, opened 2018-04-13
cheque-status field

rathorevaibhav:
Changes include:

1. Added a cheque_status field for invoice that is active when payment type is cheque


app/Http/Requests/Finance/InvoiceRequest.php
```diff
@@ -33,6 +33,7 @@ public function rules()
             'paid_on' => 'nullable',
             'paid_amount' => 'nullable|numeric',
             'payment_type' => 'nullable|string',
+            'cheque_status' => 'nullable|string|required_if:payment_type,cheque',
```
nishanth-kd: +1

resources/assets/js/app.js
```diff
@@ -110,3 +110,11 @@ $('.status-close').on('click', function(){
     let wrapper = $(this).closest('.alert');
     wrapper.fadeOut(500);
 });
+
+$('.form-invoice').on('change', '#payment_type', function(){
```
nishanth-kd: would suggest you to use Vue over Jquery for this kind of presentation logic. Way more leaner and cleaner.

resources/assets/js/app.js
```diff
@@ -110,3 +110,11 @@ $('.status-close').on('click', function(){
     let wrapper = $(this).closest('.alert');
     wrapper.fadeOut(500);
 });
+
```
nishanth-kd: is there a component based JS or everything goes into main.js?

rathorevaibhav: It's just app.js right now. I'm planning to make it component based very soon. I think having a look at MMN codebase will help.

nishanth-kd: Umm I would suggest it to be a bit different from MMN. It should be isolated based on features, kike hr.js and finance.js. I think MMN currently follows a screen based thing much like how illuma does it.

Issue #121, opened 2018-04-15
Build Feedback on HR modules


rathorevaibhav:
- [x] An autoresponder mail should go to the applicants at the time of creation. This mail template should be editable.
- [x] Communication mail shouldn't be highlighted separately in the timeline. Merge it with the round status block in the timeline.
- [x] There can be an ability to move to next round. A user can take a decision what the next round can be from the previous round.
- [x] If an applicant is rejected, it should be highlighted as an applicant's meta.

PR #123, opened 2018-04-16
(#121) hr changes

rathorevaibhav:
Changes include:

1. Instead of calling `ApplicantController::update`, now for each round, the form will be submitted to `ApplicantRoundController::update`.
2. Renamed listener `CreateApplicantRound` to `CreateFirstApplicantRound` since its job is to create the first round when an applicant is created (Resume screening).
3. Deleted `ApplicantUpdated` event/listener as in actual the applicant round is being updated. The only attribute in Applicant that is updated is the status.
4. Cleaned status in constants file and added more attributes to be used from different areas.
5. Vue Component to decide which round will be conducted next. Based on the round selected, a corresponding applicant round is created for that applicant.


app/Http/Controllers/HR/ApplicantController.php
```diff
@@ -83,7 +83,7 @@ public function edit(Applicant $applicant)
             'job' => $applicant->job,
             'applicant' => $applicant,
             'rounds' => Round::all(),
-            'applicant_rounds' => $applicant->applicantRounds,
+            'unconductedApplicantRounds' => $applicant->getUnconductedRounds(),
```
nishanth-kd: casing change? intended?

app/Http/Controllers/HR/ApplicantController.php
```diff
@@ -92,18 +92,11 @@ public function edit(Applicant $applicant)
...
     {
-        $validated = $request->validated();
-        $round_status = $validated['round_status'];
-        $status = ($round_status === config('constants.hr.round.status.rejected')) ? $round_status : config('constants.hr.round.status.in-progress');
-        $applicant->_update([
-            'status' => $status
-        ]);
-
-        return redirect("/hr/applicants/$applicant->id/edit")->with('status', 'Applicant updated successfully!');
+        //
```
nishanth-kd: lol

app/Listeners/HR/CreateFirstApplicantRound.php
```diff
@@ -8,7 +8,7 @@
 use App\User;
 use Carbon\Carbon;
 
-class CreateApplicantRound
+class CreateFirstApplicantRound
```
nishanth-kd: First applicant Round? This sounds toooo specific.

app/Models/HR/ApplicantRound.php
```diff
@@ -21,12 +22,37 @@ public static function _create($attr)
...
+                'hr_applicant_id' => $applicant->id,
+                'hr_round_id' => $nextRound,
+                'scheduled_date' => Carbon::now()->addDay(),
+                'scheduled_person_id' => $scheduled_person ? $scheduled_person->id : config('constants.hr.defaults.scheduled_person_id'),
+            ]);
+        }
     }
 
-    public function _updateOrCreateReviews($reviews = [])
+    protected function _updateOrCreateReviews($reviews = [])
```
nishanth-kd: interesting, thoughts ?

app/Http/Controllers/HR/ApplicantController.php
rathorevaibhav: yes. I figured there were variations and a specific casing wasn't being followed. I figured it's best to stick with one. And I think camelCase will be best since Vue also converts kebab-case attributes to camelCase format.

app/Listeners/HR/CreateFirstApplicantRound.php
rathorevaibhav: Yes, it creates the first round for an applicant when an applicant is created. So if resume screening is the first round that needs to be conducted, it'll be created right after an applicant is created.

app/Models/HR/ApplicantRound.php
rathorevaibhav: Oh, it was being called only from the `_update` method. I don't think it will be called from somewhere else (in near future). Moved to `protected` to avoid direct access. We can move back to public if needed.

PR #124, opened 2018-04-16
(#75) Ui enhancements

rathorevaibhav:
Changes include:

1. Added a menu for HR module
2. Added modules in the primary navigation
3. Added font awesome icons in the module menus
4. Some cleanup on the UI

### DIFF
--- a/resources/views/layouts/app.blade.php
+++ b/resources/views/layouts/app.blade.php
@@ -25,6 +25,19 @@
                 @endguest
                         {{ config('app.name', 'Employee Portal') }}
                     </a>
+                <div class="collapse navbar-collapse" id="navbarSupportedContent">
+                    <ul class="navbar-nav mr-auto">
+                        <li class="nav-item">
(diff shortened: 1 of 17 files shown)


resources/views/layouts/app.blade.php
```diff
@@ -25,6 +25,19 @@
                 @endguest
                         {{ config('app.name', 'Employee Portal') }}
                     </a>
+                <div class="collapse navbar-collapse" id="navbarSupportedContent">
+                    <ul class="navbar-nav mr-auto">
+                        <li class="nav-item">
+                            <a class="nav-link" href="{{ url('/hr/applicants') }}">HR<span class="sr-only">(current)</span></a>
```
nishanth-kd: what's the significance of current ?

rathorevaibhav: Oh this was default text for screen readers only (.sr-only). Can be removed. I'll fix and merge.

Issue #54, opened 2018-04-04
More fields on Client Creation


PrateekNarang:
Client GST number.

rathorevaibhav:
GST will only be visible if the client's country is India. Otherwise, GST will be false for all the client's projects.

PR #125, opened 2018-04-16
(#54) Added GST No

pankaj-ag:
Pr for (#54 )


resources/assets/js/app.js
```diff
@@ -161,3 +163,12 @@ $('.hr_round_guide').on('click', '.save-guide', function(){
         },
     });
 });
+
+$('.client-form').on('change', 'select[name="country"]', function() {
```
nishanth-kd: We should use Vue here for these kind of presentation logic.

database/migrations/2018_04_16_084105_update_client_with_gst_number_column.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('clients', function (Blueprint $table) {
+            $table->string('gst_no')->nullable()->after('address');
```
nishanth-kd: I would suggest `gst_num` rather than `gst_no`.

app/Http/Controllers/ClientController.php
```diff
@@ -44,6 +44,7 @@ public function store(ClientRequest $request)
             'phone' => $validated['phone'],
             'country' => $validated['country'],
             'address' => $validated['address'],
+            'gst_num' => array_key_exists('gst_num', $validated) ? $validated['gst_num'] : null
```
pankaj-ag: Just a thought, I think instead of an array we should convert **$validated** to the laravel's collection object.  
After that, we can avoid all the general checks ( i.e array_key_exists, isset ) and use the predefined collection function. 

> $validated =  collect($validated);
'gst_num' => $validated->get('gst_num' )

pankaj-ag:
@nishanth-kd @rathorevaibhav this is ready for you review.

resources/views/client/edit.blade.php
```diff
@@ -38,18 +38,23 @@
...
+                                <option value="{{ $country }}" >{{ $country_name }}</option>
                             @endforeach
                         </select>
                     </div>
                 </div>
                 <br>
+                
+                <div class="form-row" v-if="country === 'india'">
+                    <div class="form-group col-md-5">
+                        <label for="phone">GST</label>
```
bohraji: Please update the label for attribute to 'gst'

resources/views/client/edit.blade.php
```diff
@@ -38,18 +38,23 @@
...
+                        <input type="text" class="form-control" name="gst_num" id="gst_num" placeholder="GST Number" value="{{ $client->gst_num }}">
+                    </div>
+                </div>
+                <br>
                 <div class="form-row">
                     <div class="form-group col-md-12">
                         <label for="name">Address</label>
```
bohraji: Please update the label for attribute.

app/Http/Controllers/ClientController.php
bohraji: Agreed.

 But I think it will need changing in all validate method usage to maintain consistency. Please create a new issue for this and track it seperately @rathorevaibhav @pankaj-ag

Issue #51, opened 2018-04-04
Invoice gets downloaded


PrateekNarang:
In invoice, on click invoice started to download. Ideally, it should open first and with an option to download.
Also, the name of invoice pdf shouldn't change from what was uploaded. The name pattern is 0190020000010418.pdf. where last 4 digits are MMYY and the early digits are ClientIDProjectIDInvoiceID.

PR #126, opened 2018-04-16
(#51) Invoice Files

pankaj-ag:
PR for (#51)

### DIFF
--- a/app/Http/Controllers/Finance/InvoiceController.php
+++ b/app/Http/Controllers/Finance/InvoiceController.php
@@ -167,8 +167,8 @@ class InvoiceController extends Controller
      * @param  string $year  uploaded year of the invoice file
      * @param  string $month uploaded month of the invoice file
      * @param  string $file  invoice file name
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -159,13 +167,25 @@ protected static function upload(UploadedFile $file)
...
     {
+        $headers = [
+            'content-type'=>'application/pdf'
+        ];
+
         $file_path = FileHelper::getFilePath($year, $month, $file);
-        if (Storage::exists($file_path)) {
-            return Storage::download($file_path);
+
+        if(!Storage::exists($file_path)) {
```
nishanth-kd: Move this logic to FileHelper::getFilePath

If the the file doesnt exist it should return false. and here we can check only `if ($file_path) { ... }`

pankaj-ag:
@nishanth-kd @rathorevaibhav This PR is ready for code review.

Issue #91, opened 2018-04-09
Job round interviewers selection should be from current employees instead of text field


rathorevaibhav:
Job round interviewers should be a select field. The options would be the current list of employees.

PR #134, opened 2018-04-22
(#91) interviewers dropdown

pankaj-ag:
PR for (#91 )


app/Models/HR/Job.php
```diff
@@ -54,6 +59,26 @@ public function getApplicantsByStatus($status = '')
 
     public function rounds()
     {
-        return $this->belongsToMany(Round::class, 'hr_jobs_rounds', 'hr_job_id', 'hr_round_id')->withPivot('hr_job_id', 'hr_round_id', 'hr_round_interviewer');
+        return $this->belongsToMany(Round::class, 'hr_jobs_rounds', 'hr_job_id', 'hr_round_id')->withPivot('hr_job_id', 'hr_round_id', 'hr_round_interviewer_id');
+    }
+
+    public function updateInterviewers($rounds = []) {
+        foreach($rounds as $roundID => $round) {
+            $this->assignInterviewer($roundID, $round['hr_round_interviewer_id']);
```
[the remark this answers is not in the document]
pankaj-ag: done.

app/Models/HR/Job.php
```diff
@@ -54,6 +59,10 @@ public function getApplicantsByStatus($status = '')
 
     public function rounds()
     {
-        return $this->belongsToMany(Round::class, 'hr_jobs_rounds', 'hr_job_id', 'hr_round_id')->withPivot('hr_job_id', 'hr_round_id', 'hr_round_interviewer');
+        return $this->belongsToMany(Round::class, 'hr_jobs_rounds', 'hr_job_id', 'hr_round_id')->withPivot('hr_job_id', 'hr_round_id', 'hr_round_interviewer_id');
+    }
+
+    public function updateInterviewers($rounds = []) {
+        return (empty($rounds)) ? false : $this->rounds()->sync($rounds);
```
rathorevaibhav: I don't think we need false as a return value. Even if the rounds are empty, it should update and delete all the existing rounds during sync. We'll need this functionality in future when we make the rounds assignable to a job.

The method `updateInterviewers` is, IMO, ambiguous as it just not updating interviewers but also updating job and round association. That means we just need `$this->rounds()->sync($rounds)`. Can we put it inside the if statement in line 40?

database/migrations/2018_04_22_133003_update_hr_jobs_rounds_table_by_replacing_hr_round_interviewer_user_id.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('hr_jobs_rounds', function (Blueprint $table) {
+            $table->dropColumn('hr_round_interviewer');
```
rathorevaibhav: 👍

Issue #133, opened 2018-04-18
More changes in Applicant screen


rathorevaibhav:
- [x] Remove the communication modal toggle link from applicant round box.
- [x] Applicant round should be visible on the applicant listing page.
- [ ] Ability to reopen a case.
- [ ] Delete a round. Only available for rounds for which mail didn't go.

PR #137, opened 2018-04-26
(#133) applicant changes

rathorevaibhav:
#133

Changes include:

1. Showing applicant's current round and previously completed round if the applicant is in progress.
2. Removed unconducted rounds for an applicant. Instead, the applicant can go to any round as per the requirement.

### DIFF
--- a/app/Models/HR/ApplicantRound.php
+++ b/app/Models/HR/ApplicantRound.php
@@ -33,11 +33,7 @@ class ApplicantRound extends Model
 
         $applicant = $this->applicant;
         if ($attr['round_status']) {
-            if ($attr['round_status'] == 'rejected') {
-                $applicantStatus = 'rejected';
-            } else {
-                $applicantStatus = sizeof($applicant->getUnconductedRounds()) ? 'in-progress' : 'completed';
-            }
+            $applicantStatus = ($attr['round_status'] == 'rejected') ? 'rejected' : 'in-progress';
             $applicant->update([ 'status' => $applicantStatus ]);
         }
 
--- a/app/Models/HR/ApplicantRound.php
+++ b/app/Models/HR/ApplicantRound.php
@@ -33,7 +33,10 @@ class ApplicantRound extends Model
 
         $applicant = $this->applicant;
         if ($attr['round_status']) {
-            $applicantStatus = ($attr['round_status'] == 'rejected') ? 'rejected' : 'in-progress';
+            $status = config('constants.hr.status');
+            $rejectedStatus = $status['rejected']['label'];
+            $inProgressStatus = $status['in-progress']['label'];
+            $applicantStatus = ($attr['round_status'] === $rejectedStatus) ? $rejectedStatus : $inProgressStatus;
             $applicant->update([ 'status' => $applicantStatus ]);
         }
 
--- a/resources/views/hr/applicant/edit.blade.php
+++ b/resources/views/hr/applicant/edit.blade.php
@@ -118,7 +118,7 @@
                         @if (! $applicantRound->round_status)
                             <div class="card-footer">
                                 <applicant-round-action-component
-                                :rounds="{{ json_encode($unconductedApplicantRounds) }}">
+                                :rounds="{{ json_encode($job->rounds) }}">
                                 </applicant-round-action-component>
                                 <button type="button" class="btn btn-outline-danger round-submit" data-status="rejected">Reject</button>
                             </div>
@@ -126,7 +126,7 @@
                             <div class="card-footer">
                                 @if ($applicantRound->round_status == 'rejected')
                                     <applicant-round-action-component
-                                    :rounds="{{ json_encode($unconductedApplicantRounds) }}">
+                                    :rounds="{{ json_encode($job->rounds) }}">
                                     </applicant-round-action-component>
                                 @endif
                                 @if (!$applicantRound->mail_sent)
--- a/resources/views/hr/applicant/edit.blade.php
+++ b/resources/views/hr/applicant/edit.blade.php
@@ -109,46 +109,44 @@
                                     <textarea name="reviews[feedback]" id="reviews[feedback]" rows="6" class="form-control">{{ $applicantReviewValue }}</textarea>
                                 </div>
                             </div>
-                        </div>
-                        <div class="card-footer">
-                        @if (! $applicantRound->round_status)
-                            <applicant-round-action-component
-                            :rounds="{{ json_encode($unconductedApplicantRounds) }}">
-                            </applicant-round-action-component>
-                            <button type="button" class="btn btn-outline-danger round-submit" data-status="rejected">Reject</button>
-                        @else
-                            <div class="d-flex align-items-center justify-content-between">
-                                <div>
-                                    <button type="button" class="btn btn-info round-update">Update</button>
-                                    @if ($applicantRound->round_status == 'rejected')
-                                        <applicant-round-action-component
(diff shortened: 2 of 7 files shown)


app/Models/HR/ApplicantRound.php
```diff
@@ -33,11 +33,7 @@ public function _update($attr, $type = 'new', $reviews = [], $nextRound = 0)
 
         $applicant = $this->applicant;
         if ($attr['round_status']) {
-            if ($attr['round_status'] == 'rejected') {
-                $applicantStatus = 'rejected';
-            } else {
-                $applicantStatus = sizeof($applicant->getUnconductedRounds()) ? 'in-progress' : 'completed';
-            }
+            $applicantStatus = ($attr['round_status'] == 'rejected') ? 'rejected' : 'in-progress';
```
pankaj-ag: Can we add the round status in the configuration file? (maybe constants.php). I think It will give us the flexibility to change these statuses name in the future. 

Also, can we use === instead of == for Identical match. 
https://stackoverflow.com/questions/6316784/what-is-the-difference-between-and-in-php

resources/views/hr/applicant/index.blade.php
```diff
@@ -9,30 +9,31 @@
...
         @foreach ($applicants as $applicant)
         <tr>
             <td>
                 <a href="/hr/applicants/{{ $applicant->id }}/edit">{{ $applicant->name }}</a>
             </td>
-            <td>{{ $applicant->email }}<br>{{ $applicant->phone }}</td>
+            <td>{{ $applicant->email }}</td>
             <td>{{ $applicant->job->title }}</td>
-            <td>{{ $applicant->created_at }}</td>
+            <td>{{ date(config('constants.display_date_format'), strtotime($applicant->created_at)) }}</td>
```
pankaj-ag: I think $applicant->created_at (timestamps) are already casted to carbon object in laravel, can we use those features instead of starttime.
Ie. $applicant->created_at->toFormattedDateString();

resources/views/hr/applicant/index.blade.php
```diff
@@ -9,30 +9,31 @@
...
+                <span class="d-flex justify-content-start">
+                    @if (in_array($applicant->status, ['in-progress', 'new']))
+                        <span class="badge badge-warning badge-pill">{{ $applicant->applicantRounds->last()->round->name }}</span>
+                        @if ($applicant->applicantRounds->count() > 1)
+                            <span class="badge badge-info badge-pill ml-1 px-2">Completed: {{ $applicant->applicantRounds->count() - 1 }}</span>
+                        @else
+                            <span class="badge badge-info badge-pill ml-1 px-2">New</span>
+                        @endif
+                    @else
+                        <span class="{{ config("constants.hr.status.$applicant->status.class") }} badge-pill">{{ config("constants.hr.status.$applicant->status.title") }}</span>
```
pankaj-ag: 👍

resources/views/hr/applicant/timeline.blade.php
```diff
@@ -24,11 +24,9 @@
                     <b><u>{{ date(config('constants.display_date_format'), strtotime($applicantRound->conducted_date)) }}</u></b><br>
                     {{ $applicantRound->round->name }} conducted by {{ $applicantRound->conductedPerson->name }}<br>
                     @if ($applicantRound->mail_sent)
-                        <span data-toggle="modal" data-target="#round_mail_{{ $applicantRound->id }}" class="modal-toggler-text text-primary">Communication mail</span><br>
+                        <span data-toggle="modal" data-target="#round_mail_{{ $applicantRound->id }}" class="{{ config("constants.hr.status.$applicantRound->round_status.class") }} modal-toggler">Communication mail</span><br>
```
pankaj-ag: what is $applicantRound->round_status is null. Because  Can we use laravel's  **optional**  function 
https://laravel.com/docs/5.6/helpers#method-optional.

resources/views/hr/applicant/edit.blade.php
```diff
@@ -109,46 +109,38 @@
...
-                                    @else
-                                        <button type="button" class="btn btn-primary" data-toggle="modal" data-target="#round_{{ $applicantRound->id }}">Send mail</button>
-                                    @endif
-                                </div>
-                            </div>
-                        @endif
                         </div>
+                        @elseif ($applicantRound->round_status == 'rejected' || !$applicantRound->mail_sent)
+                        <div class="card-footer">
+                            @if ($applicantRound->round_status == 'rejected')
```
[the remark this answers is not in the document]
pankaj-ag: Do you think we should add some validation for the Vue props? 
Please have a look at **Prop Validation** section in this https://vuejs.org/v2/guide/components-props.html.

app/Models/HR/ApplicantRound.php
pankaj-ag: Can we add the round status in the configuration file? (maybe constants.php). I think It will give us the flexibility to change these statuses name in the future. 

Also, can we use === instead of == for Identical match. 
https://stackoverflow.com/questions/6316784/what-is-the-difference-between-and-in-php

resources/views/hr/applicant/edit.blade.php
```diff
@@ -109,46 +109,38 @@
...
+                                    <button type="button" class="btn btn-info btn-sm round-update">Update feedback</button>
+                                </div>
+                            @endif
                         </div>
-                        <div class="card-footer">
                         @if (! $applicantRound->round_status)
+                        <div class="card-footer">
                             <applicant-round-action-component
-                            :rounds="{{ json_encode($unconductedApplicantRounds) }}">
+                            :rounds="{{ json_encode($job->rounds) }}">
```
pankaj-ag: Do you think we should add validations for the Vue props also?
Please have a look at Prop Validation section https://vuejs.org/v2/guide/components-props.html

resources/views/hr/applicant/edit.blade.php
pankaj-ag: Can we put all these kind of constants to the config files and avoid hardcoding?

resources/views/hr/applicant/index.blade.php
pankaj-ag: I think $applicant->created_at (timestamps) are already cast to carbon object in laravel, can we use those features instead of strtotime.
Ie`. $applicant->created_at->toFormattedDateString();`

resources/views/hr/applicant/timeline.blade.php
pankaj-ag: what if $applicantRound->round_status is null. Because  Can we use laravel's  **optional**  function 
https://laravel.com/docs/5.6/helpers#method-optional.

resources/views/hr/applicant/edit.blade.php
rathorevaibhav: Agreed.

resources/views/hr/applicant/edit.blade.php
rathorevaibhav: I think that would be better. Let's target this separately. This can be a `good first issue` for anyone new to the product development.

rathorevaibhav: Tracking this in #138

resources/views/hr/applicant/timeline.blade.php
rathorevaibhav: Just thinking over it. This code runs when the mail is sent to the applicant for that round. Mail can only be send if there's a status set for the round. Maybe this is not the correct place to check if the round status exists? Thoughts?

Issue #135, opened 2018-04-26
Develop Library Module 


pankaj-ag:

pankaj-ag:
We have set up the module. Closing this one as complete.

PR #141, opened 2018-04-28
(#135) save book

pankaj-ag:


app/Http/Requests/KnowledgeCafe/Library/BookRequest.php
```diff
@@ -24,8 +24,15 @@ public function authorize()
...
-            'add_method' => 'required|string',
-        ];
+            'add_method' => 'filled|string',
+            'title' => 'filled|string',
+            'author' => 'nullable|string',
+            'readable_link' => 'filled|string',
+            'categories' => 'nullable|string',
+            'thumbnail' => 'filled|string',
+            'isbn' => 'filled|string',
+            'selfLink' => 'nullable|string'
```
rathorevaibhav: follow the same naming convention. self_link

app/Models/KnowledgeCafe/Library/Book.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace App\Models\KnowledgeCafe\Library;
+
+use Illuminate\Database\Eloquent\Model;
+
+class Book extends Model
+{
+    protected $table = 'library_books';
+    protected $fillable = ['title', 'author', 'categories', 'isbn', 'thumbnail', 'readable_link', 'selfLink'];
```
rathorevaibhav: same here and at all other places. self_link

resources/assets/js/app.js
```diff
@@ -192,18 +193,24 @@ $('.hr_round_guide').on('click', '.save-guide', function(){
...
     const bookForm = new Vue({
-        el: '#book_form',
+        el: '#show_and_save_book',
         data: {
-            addMethod: 'from_image'
+            addMethod: 'from_isbn',
+            showInfo: false,
+            book: {},
+            routes: {
+                index:$('#show_book').attr('data-index-route'),
```
rathorevaibhav: I think it'll better to avoid jQuery whenever we can. Let's use the native document.getElementByID and the dataset.
It'll be faster and we'll also follow the same convention. I think we should only use jQuery where we're unclear of it's alternative.

resources/assets/js/app.js
```diff
@@ -221,40 +228,58 @@ if (document.getElementById('book_form')) {
...
                 if(this.compressedFile) {
                     formData.append('book_image', compressedFile, compressedFile.name);
                 }
-                $('#show_book').html('');
-                this.bookData = null;
-                axios.post('/knowledgecafe/library/book/fetchinfo', formData).then(
+
+                this.book = {};
+                let button =  $('#submit_book_form_btn');
+                button.prop('disabled', true).find('.item').toggleClass('d-none');
```
rathorevaibhav: can we use v-show at the button?

resources/assets/js/app.js
pankaj-ag: @rathorevaibhav I have made the changes. I agree that we should follow the conventions. 
Just curious why we are avoiding jquery.

app/Http/Controllers/KnowledgeCafe/Library/BookController.php
```diff
@@ -94,41 +98,65 @@ public function destroy($id)
...
         $method = $validated['add_method'];
-        $ISBN = null;
-        
+
         if($method === 'from_image' && $request->hasFile('book_image')) {
             $file = $request->file('book_image');
             $ISBN = BookServices::getISBN($file);
-        } else if ($method ==='from_isbn') {
+        } else if($method ==='from_isbn') {
             $ISBN = $request->input('isbn');
```
rathorevaibhav: use $validated

resources/views/knowledgecafe/library/books/index.blade.php
```diff
@@ -7,18 +7,37 @@
...
+        @foreach($books as $book)
+            <tr>
+                <td> 
+                    <a href="{{ route('books.show', $book->id) }}"> 
+                        {{ str_limit($book->title, 25) }} </a>
+                    </td>
+                <td> {{ str_limit($book->author, 20) }} </td>
+                <td> {{ str_limit($book->categories, 20) }} </td>
+                <td> 
+                    <div style = "width:75px; height:150px">
```
rathorevaibhav: external CSS.

resources/views/knowledgecafe/library/books/info.blade.php
```diff
@@ -1,58 +1,71 @@
...
+                <div class="col-6">
                     <div class="row">
-                        <div class="col-3">
-                            <p>
-                                <b>Category : </b>
-                            </p>
-                        </div>
                         <div class="col-6">
-                            <p> {{ implode ($book['volumeInfo']['categories']) }} </p>
+                                <img v-bind:src="book.thumbnail" />
```
rathorevaibhav: this is fine. Just an additional thing that in Vue2.0, you can also directly bind attributes using just `:src`.

Issue #74, opened 2018-04-06
In-active state should be automatic


PrateekNarang:
In-Active state for a client should be automatic. Otherwise, there is a lot of clutter in the drop-down. Something like if we have not sent them an invoice for more than three months, they should be qualified for inactive state.

rathorevaibhav:
@PrateekNarang I think for the initial version, we can add a flag to the client edit screen to set the state for the client. Basically "active" or "inactive". On the projects and invoices screens, we can then only show the clients with the state set to active.

PrateekNarang:
Agree.

PR #144, opened 2018-04-30
(#74) client status

rathorevaibhav:
#74 

Changes include:

1. Active flag for client
2. Showing only active client list in create project/invoice screens


pankaj-ag:
Looks good to me.

PR #176, opened 2018-05-05
(#168) GST and TDS fields cleanup

rathorevaibhav:


app/Models/Client.php
```diff
@@ -24,6 +24,6 @@ public function projects()
      */
     public static function getActiveClients()
     {
-        return self::select('id', 'name')->where('is_active', true)->get();
+        return self::where('is_active', true)->get();
```
nishanth-kd: getting all fields instead of just `name` and `id` now?

resources/assets/js/app.js
```diff
@@ -53,7 +53,8 @@ if (document.getElementById('form_invoice')) {
             chequeStatus: document.getElementById('cheque_status').dataset.chequeStatus || null,
             selectedClient: '',
             activeClient: [],
-
+            countries: document.getElementById('client_id').dataset.countries || [],
+            activeClientCurrency: 'INR'
```
nishanth-kd: I presume this is the default.

rathorevaibhav: Yes

PR #177, opened 2018-05-07
(#161) invoice listing

rathorevaibhav:
#161

### DIFF
--- a/app/Http/Controllers/Finance/InvoiceController.php
+++ b/app/Http/Controllers/Finance/InvoiceController.php
@@ -9,6 +9,7 @@ use App\Http\Requests\Finance\InvoiceRequest;
 use App\Models\Client;
 use App\Models\Finance\Invoice;
 use App\Models\ProjectStageBilling;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -22,6 +23,19 @@ class InvoiceController extends Controller
      */
     public function index()
     {
+        $request = request();
+
```
nishanth-kd:
```
if (...) { 
$startDate = $request->get('start');
$endDate = $request->get('end');
$attr = [
               'invoices' => Invoice::filterByDates($startDate, $endDate)->appends(Input::except('page')),
               'startDate' => $startDate,
               'endDate' => $endDate,
];
} else {
$attr = [  'invoices' => Invoice::getList()];
}
return view('finance.invoice.index')->with($attr)
```

nishanth-kd: avoids duplicating the view file being referenced.

Issue #166, opened 2018-05-03
Reports in Finance


rathorevaibhav:
- [ ] Total monthly income
- [ ] Monthly due
- [ ] Monthly TDS

PR #178, opened 2018-05-07
(#166) finance reports

rathorevaibhav:
#166 

Needs to be reviewed and merged after #177


app/Http/Controllers/Finance/ReportsController.php
```diff
@@ -0,0 +1,77 @@
...
+            return view('finance.reports.index')->with([
+                'invoices' => $invoices,
+                'report' => self::getReportDetails($invoices),
+                'startDate' => $startDate,
+                'endDate' => $endDate,
+            ]);
+        }
+
+        $invoices = Invoice::all();
+        return view('finance.reports.index')->with([
```
nishanth-kd: A similar feedback here. I would suggest you look at this again. I see this code can be optimised in terms of number of lines.

app/Http/Controllers/Finance/ReportsController.php
```diff
@@ -0,0 +1,77 @@
...
+     * @return array
+     */
+    public static function getReportDetails($invoices)
+    {
+        $report = [];
+        $report['tds'] = 0;
+        $report['gst'] = 0;
+        foreach ($invoices as $invoice) {
+            $report['gst'] += $invoice->gst;
+            if (!isset($report['sentAmount'][$invoice->currency_sent_amount])) {
```
nishanth-kd: Can these Calculations be done via MYSQL instead by using MYSQLs Caculation functions to aggregate values in particular columns. 

I think you should give it a shot :)

app/Http/Controllers/Finance/ReportsController.php
```diff
@@ -0,0 +1,86 @@
...
+        $report['tds'] = 0;
+        $report['gst'] = 0;
+        foreach (config('constants.currency') as $currency => $currencyMeta) {
+            $report['sentAmount'][$currency] = 0;
+            $report['paidAmount'][$currency] = 0;
+            $report['transactionCharge'][$currency] = 0;
+            $report['transactionTax'][$currency] = 0;
+        }
+        foreach ($invoices as $invoice) {
+            $report['gst'] += $invoice->gst;
```
nishanth-kd: couldn't sum up the GST via SQL ?

nishanth-kd: It appears my previous comments haven't been address. Please don't assign for review in future unless previous feedbacks haven't been addressed.

rathorevaibhav: @nishanth-kd I had read your previous comment on getting report calculations from the SQL but missed addressing.

Right now, we're not very sure what exactly we need in reporting and how exactly that'll be used. Considering that and also we have to make these calculations in SQL, the code is wrapped within a single method which we can remove later when we move to SQL.

I've created a tracking for this #182 . Once we test and use the reports feature, I think then it'll be better to move to SQL for further performance enhancements.

Issue #184, opened 2018-05-08
Usability feedback on Reports


rathorevaibhav:
- [x] Received amount should be total INR (Sum of INR and USD)
- [x] In the default card-header, can we print "Showing results for **April 2018**" instead of showing the date range?
- [x] Bank charges label to "Bank charges on Fund Transfer"
- [x] Bank charges currency default to USD
- [x] Conversion rate should be mandatory if invoice status is paid and currency is non-INR
- [x] Due balance field in the invoice
- [x] Tabs to show sent invoices and received invoices (table) in reports
- [x] Sorting in received and sent invoices table.
- [x] Receivable section (next to invoiced amount). Should show total amount yet to be received.
- [x] A conversion rate input in "invoiced amount" section to convert and show total estimated invoiced in INR.

PR #191, opened 2018-05-09
(#184) reports test feedback

rathorevaibhav:
#184 

Changes include:

1. "Receivable" section for the amount yet to be received.
2. The ordering of invoices in the report tables.
3. An input to enter estimated conversion rate and amount to get an idea of the total estimated sent amount.

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -129,7 +129,19 @@ if (document.getElementById('finance_report')) {
     const invoiceForm = new Vue({
         el: '#finance_report',
         data: {
-            showReportTable: 'received'
+            showReportTable: 'received',
+            sentAmountINR: document.getElementById('sent_amount_INR').dataset.sentAmount || 0,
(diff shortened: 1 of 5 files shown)


resources/assets/js/app.js
```diff
@@ -129,7 +129,19 @@ if (document.getElementById('finance_report')) {
     const invoiceForm = new Vue({
         el: '#finance_report',
         data: {
-            showReportTable: 'received'
+            showReportTable: 'received',
+            sentAmountINR: document.getElementById('sent_amount_INR').dataset.sentAmount || 0,
+            sentAmountUSD: document.getElementById('sent_amount_USD').dataset.sentAmount || 0,
+            conversionRateUSD: 65,
```
bohraji: Conversion rate should not be Hard Coded. It can be fetched from some config file or some external service if we want to account for varying rates.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

Issue #192, opened 2018-05-09
Login for organisation's CA


rathorevaibhav:
The Chartered Accountant for an organization should be able to log in and see the only finance reports (for now).

PR #194, opened 2018-05-09
(#192) ACL implementation (for CA)

rathorevaibhav:
#192

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -67,7 +67,7 @@ class RegisterController extends Controller
             'name' => $data['name'],
             'email' => $data['email'],
             'password' => Hash::make($data['password']),
-            'provider' => 'custom',
+            'provider' => 'default',
             'provider_id' => '',
         ]);
     }
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -67,6 +67,8 @@ class RegisterController extends Controller
             'name' => $data['name'],
             'email' => $data['email'],
             'password' => Hash::make($data['password']),
+            'provider' => 'custom',
+            'provider_id' => '',
         ]);
     }
 }
--- a/app/Http/Controllers/ClientController.php
+++ b/app/Http/Controllers/ClientController.php
@@ -20,7 +20,7 @@ class ClientController extends Controller
      */
     public function index()
     {
-        //
+        $this->authorize('list', Client::class);
 
         return view('client.index')->with([
             'clients' => Client::select('id', 'name', 'emails', 'phone', 'is_active')->orderBy('id', 'desc')->paginate(config('constants.pagination_size')),
--- a/app/Http/Controllers/ClientController.php
+++ b/app/Http/Controllers/ClientController.php
@@ -14,6 +14,8 @@ class ClientController extends Controller
      */
     public function index()
     {
+        $this->authorize('view', new Client);
+
         return view('client.index')->with([
             'clients' => Client::select('id', 'name', 'emails', 'phone', 'is_active')->orderBy('id', 'desc')->paginate(config('constants.pagination_size')),
         ]);
@@ -26,6 +28,8 @@ class ClientController extends Controller
      */
     public function create()
     {
+        $this->authorize('create', Client::class);
+
         return view('client.create');
     }
 
@@ -37,6 +41,8 @@ class ClientController extends Controller
      */
     public function store(ClientRequest $request)
     {
+        $this->authorize('create', Client::class);
+
         $validated = $request->validated();
         $client = Client::create([
             'name' => $validated['name'],
@@ -59,7 +65,7 @@ class ClientController extends Controller
      */
     public function show(Client $client)
     {
-        //
+        $this->authorize('view', $client);
     }
 
     /**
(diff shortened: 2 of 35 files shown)


app/Http/Controllers/ClientController.php
```diff
@@ -84,6 +92,8 @@ public function edit(Client $client)
      */
     public function update(ClientRequest $request, Client $client)
     {
+        $this->authorize('update', $client);
```
nishanth-kd: Can these be the in the request validation? Acting like a middleware

app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -23,6 +23,8 @@ class InvoiceController extends Controller
      */
     public function index()
     {
+        $this->authorize('view', new Invoice);
```
nishanth-kd: We're sending an object here but not in other scenarios.

app/Http/Controllers/KnowledgeCafe/Library/BookController.php
```diff
@@ -46,46 +51,48 @@ public function store(BookRequest $request)
...
-     * @param  int  $id
-     * @return @return void
+     * @param  \App\Models\KnowledgeCafe\Library\Book  $book
+     * @return void
      */
-    public function update(Request $request, $id)
+    public function update(Request $request, Book $book)
     {
-        //
+        $this->authorize('update', $book);
```
nishanth-kd: Have we given thought to `authorizeResource()`

database/seeds/RoleHasPermissionsTableSeeder.php
```diff
@@ -0,0 +1,25 @@
...
+     *
+     * @return void
+     */
+    public function run()
+    {
+        $superAdmin = Role::where(['name' => 'super-admin'])->first();
+        $superAdmin->syncPermissions(Permission::all());
+
+        $admin = Role::where(['name' => 'admin'])->first();
+        $admin->syncPermissions(Permission::all());
```
nishanth-kd: I believe these migrations were added by the spatie package.

routes/web.php
```diff
@@ -51,3 +48,6 @@
     Route::post('/knowledgecafe/library/book/fetchinfo', 'KnowledgeCafe\Library\BookController@fetchBookInfo')->name('books.fetchInfo');
     Route::get('/finance/reports', 'Finance\ReportsController@index');
 });
+
+// Route::middleware(['auth', 'permission:view.finance_reports'])->group(function(){
```
nishanth-kd: what's with this?

app/Http/Controllers/KnowledgeCafe/Library/BookController.php
nishanth-kd: More thoughts and opinions here https://laracasts.com/discuss/channels/laravel/how-to-apply-policy-to-a-resource-controller?page=1

nishanth-kd:
@rathorevaibhav did my first round of review for this. I believe most of the architectural questions i had on this are resolved, because we're using the @spatie package for roles and permission. we may not need to sit with @rudresh and @bohraji on this given how this implementation has panned out. We can just request them to go over this PR, it's comments etc for opinions. 

I've checked this PR: 
- [x] For duplication.
- [x] For performance. 
- [x] For coupling

routes/web.php
rathorevaibhav: My bad. I had added authorization checks on the routes before (only for reports) but since we are doing authorization checks at the controller level, this code was commented out.

We're just using auth middleware check on the routes (for now). Checking on the route adds an extra layer of security and is maybe better because it happens at the entry point. But given we require different types of authorization checks for different requests within a (resource) controller, I added all checks to the controller methods itself.

rathorevaibhav: Have removed this commented code.

app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -67,6 +67,8 @@ protected function create(array $data)
             'name' => $data['name'],
             'email' => $data['email'],
             'password' => Hash::make($data['password']),
+            'provider' => 'custom',
```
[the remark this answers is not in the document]
rathorevaibhav: We're using Laravel's socialite (only Google login). It requires a provider (google/facebook/twitter etc) and a provider id (retrieved from OAuth handshake). 

These provider details are mandatory in database. To differentiate, we're using "custom" for accounts generated via Laravel's default RegisterController and an empty provider id.

app/Http/Controllers/ClientController.php
rathorevaibhav: Based on our discussion, it won't be required if we use authorizeResource

database/seeds/RoleHasPermissionsTableSeeder.php
rathorevaibhav: No, these were added by us to add roles and permissions and their associations

rathorevaibhav:
Based on the discussion:

1. `provider_type` will be `default` if created using Laravel's register controller.
2. Renamed permissions to `module_name.permission`
3. `authorizeResource` is in place for each ResourceController
4. A custom `list` method was created in each resource policy for authorization the Controller's index method.

@nishanth-kd ready for your review again.

rathorevaibhav:
Commands to run during deployment:

```
$ php artisan migrate
$ php artisan db:seed --class=RolesTableSeeder
$ php artisan db:seed --class=PermissionsTableSeeder
$ php artisan db:seed --class=RoleHasPermissionsTableSeeder
```

nishanth-kd:
Good to go.

Issue #193, opened 2018-05-09
Details in reporting tables


rathorevaibhav:
- [ ] Invoice Number
- [x] Client name
- [x] Client's GST (if applicable)
- [ ] Showing foreign and local clients separately
- [x] Invoice file download icon
- [x] Option to see reports for previous invoices

PR #199, opened 2018-05-11
(#193) reporting details

rathorevaibhav:
#193 

Changes include:

1. Client name in tables in reports - received and sent
2. Client's GST (if applicable)
3. Invoice file download icon
4. Option to see reports for previous invoices - A drop-down to filter reports based on previous 6 months.

### DIFF
--- a/app/Http/Controllers/Finance/ReportsController.php
+++ b/app/Http/Controllers/Finance/ReportsController.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers\Finance;
 
+use App\Helpers\DateHelper;
 use App\Http\Controllers\Controller;
 use App\Models\Finance\Invoice;
 use Carbon\Carbon;
@@ -18,9 +19,19 @@ class ReportsController extends Controller
     {
         $request = request();
 
-        if ($request->get('show') && $request->get('show') == 'default') {
-            $startDate = new Carbon('first day of last month');
-            $endDate = new Carbon('last day of last month');
+        if ($request->get('type') == 'monthly') {
+
+            if ($request->get('month') && $request->get('year')) {
+                $month = $request->get('month');
+                $year = $request->get('year');
+                $date = "$year-$month-01";
+                $startDate = new Carbon($date);
+                $endDate = (new Carbon($date))->endOfMonth();
+            } else {
+                $startDate = new Carbon('first day of this month');
+                $endDate = new Carbon('last day of this month');
+            }
+
             $formattedStartDate = $startDate->format(config('constants.date_format'));
             $formattedEndDate = $endDate->format(config('constants.date_format'));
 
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/Finance/ReportsController.php
```diff
@@ -18,9 +19,17 @@ public function index()
...
+                $date = "$year-$month-01";
+                $startDate = new Carbon($date);
+                $endDate = $startDate->copy()->endOfMonth();
+            } else {
+                $startDate = new Carbon('first day of this month');
+                $endDate = new Carbon('today');
+            }
             $formattedStartDate = $startDate->format(config('constants.date_format'));
             $formattedEndDate = $endDate->format(config('constants.date_format'));
             $invoices = Invoice::filterByDates($formattedStartDate, $formattedEndDate);
```
bohraji: In all the reports, we are querying the `Invoice` model to get invoices and then arrange them. How we query `Invoice` model is dependent on the `start` and `end` date, which is specific to report.

Can we have an instance variable `$start_date` and `$end_date`, set to `null` and have them filled up  according to type of the report and the processing required.  Thereafter, we can call a common function `getInvoiceReportDetails` and have it handle the fetching of invoice info, arranging, etc.

Also, we should add an instance variable $attr, which will be returned. Since a lot of info is common across all types of reports, the above `getInvoiceReportDetails` function should add the common info to the variable as well.

The rest of report specific info can be added inside the switch case, after the report info has been added.

app/Models/Finance/Invoice.php
```diff
@@ -64,6 +64,13 @@ public static function filterByDates($start, $end, $paginated = false)
             })->orderBy('sent_on', 'desc')
             ->orderBy('paid_on', 'desc');
 
+        $invoices->with('projectStageBillings.projectStage.project.client');
+
         return $paginated ? $invoices->paginate(config('constants.pagination_size')) : $invoices->get();
     }
+
+    public function getClientAttribute()
```
bohraji: Unused addition.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

app/Http/Controllers/Finance/ReportsController.php
```diff
@@ -18,9 +19,17 @@ public function index()
     {
         $request = request();
 
-        if ($request->get('show') && $request->get('show') == 'default') {
-            $startDate = new Carbon('first day of this month');
-            $endDate = new Carbon('today');
+        if ($request->get('type') == 'monthly') {
```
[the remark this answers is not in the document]
nishanth-kd: I would kind on build on @bohraji's thought and suggest that the logic for which report should be generated be delegated to another method.

app/Http/Controllers/Finance/ReportsController.php
nishanth-kd: Interesting approach towards this abstraction. @rathorevaibhav thoughts?

app/Models/Finance/Invoice.php
nishanth-kd: Good catch @bohraji

app/Http/Controllers/Finance/ReportsController.php
rathorevaibhav: Based on some refactoring, I think @bohraji's approach made sense. Though I'm not sure if I've implemented exactly what he meant, I think the updated codebase is far cleaner without any duplicacy.

rathorevaibhav:
@bohraji done with the requested changes. Ready for review again.

Issue #200, opened 2018-05-12
Add a new stage: "Trial Program"


rathorevaibhav:
Need to add a new stage with title "Trial Program". This basically means an applicant is looking good and the company can work with him/her out on a trial period.

Note: please make sure it gets attached to all existing jobs.

PR #201, opened 2018-05-12
(#200) Trial Program Round

rathorevaibhav:
#200 

Changes include

1. Seeder to add a new job round: Trial Program.
2. Also makes sure it gets attached to existing rounds.

Once deployed, the following command needs to be run:

```
$ php artisan db:seed --class=HRRoundsTrialProgramSeeder
```

### DIFF
--- a/database/seeds/HRRoundsTrialProgramSeeder.php
+++ b/database/seeds/HRRoundsTrialProgramSeeder.php
@@ -1,0 +1,24 @@
+<?php
+
+use App\Models\HR\Job;
+use App\Models\HR\Round;
+use Illuminate\Database\Seeder;
+use Illuminate\Support\Carbon;
+
+class HRRoundsTrialProgramSeeder extends Seeder
+{
+    /**
+     * Run the database seeds.
+     *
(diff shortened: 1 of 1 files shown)


database/seeds/HRRoundsTrialProgramSeeder.php
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
+        $round = Round::create([ 'name' => 'Trial Program' ]);
+
+        $jobs = Job::all();
```
bohraji: Please Eager loading rounds with jobs. It will prevent additional db fetch to load rounds of each job one by one with each iteration.

bohraji: We are only accessing the relationship to add more values, and not to access these properties. No need to eager load.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

Issue #204, opened 2018-05-12
Duplicate applicants


rathorevaibhav:
- [x] We should track applicant as a single entity identified by their email.
- [x] The applications for an applicant can be tracked separately using applicant_id as the reference.
- [x] Need to see how the UI will look when an applicant applies for multiple job.
- [x] A single interview schedule can be tracked as the reference of progress if an applicant has applied for multiple jobs.
- [x] Applicant may or may not get rejected for one or all jobs he has applied for.
- [x] What'll happen if an applicant comes after a certain amount of time (after a year or more). His resume and skillset may differ. Will he be allowed to add a different resume and answers to queries?

PR #210, opened 2018-05-15
(#204) applicant application

rathorevaibhav:
#204 

Contains changes included in #207 along with following additions:

1. A new applicant won't be created if he/she is applying again (for the same or different job). Instead, a new application will be created.
2. Reason for eligibility, resume, and some other fields can be different for each application. Hence these details will be stored in `hr_applications` table.
3. `hr_applicant_round` will be renamed to `hr_application_round`. Will reference application_id instead of applicant_id.
4. `hr_applicant_reviews` will be renamed to `hr_application_review`. Will reference application_round_id instead of applicant_round_id.
5. The applicant edit screen will be replaced by application edit screen.


rathorevaibhav:
@bohraji since these are heavy changes, let me know if you want to sit together and discuss what I've done in this PR.

nishanth-kd:
I see these are a lot of changes for one reviewer to review, @bohraji are you planning to review it later or we're deliberately holding off reviews on this one. Let's follow a daily review approach please @rathorevaibhav.

rathorevaibhav:
Based on discussion with @bohraji, there are various questions:

1. The heavy changes are due to the result of coupling between modules. Need to talk with the team (looping in @rudresh) if this level of coupling is necessary or we can reduce it.
2. @nishanth-kd This PR is a fork from #207 and some of the changes in this PR were reviewed yesterday. While we targeted shipping this piece into two smaller PRs but figured that won't be efficient for both reviews and testing. This is just one time where we couldn't make it into smaller chunks. We're sticking with the daily review and build plan.

PR #222, opened 2018-05-16
(#204) applicant changes

rathorevaibhav:
#204 

Changes include:

1. Migration for deleting old columns `hr_applicant_id` from tables. After changes implemented in #210. These columns were needed to migrate data only and are no longer required.
2. Check for an applicant if it already exists.
3. Modified the workflow around applicant and application generation.
4. Modified the workflow around application round actions.
5. In case an applicant applies for multiple jobs, the first job is marked as in-progress. All the following will be marked as on-hold.
6. In case an applicant applies for multiple jobs, the reject action will open up a modal. That modal will have options either to reject for this application but refer for another appliction or to reject for all applications at once.

### DIFF
--- a/app/Http/Controllers/HR/ApplicationController.php
+++ b/app/Http/Controllers/HR/ApplicationController.php
@@ -77,12 +77,13 @@ class ApplicationController extends Controller
      */
     public function edit(Application $application)
     {
-        $application->load(['job.rounds', 'applicant', 'applicant.applications', 'applicationRounds']);
+        $application->load(['job.rounds', 'applicant', 'applicant.applications', 'applicationRounds', 'applicationRounds.round']);
 
         return view('hr.application.edit')->with([
             'applicant' => $application->applicant,
             'application' => $application,
             'rounds' => Round::all(),
+            'applicantOpenApplications' => $application->applicant->openApplications(),
         ]);
     }
 
--- a/app/Http/Controllers/HR/ApplicationController.php
+++ b/app/Http/Controllers/HR/ApplicationController.php
@@ -6,6 +6,7 @@ use App\Http\Controllers\Controller;
 use App\Models\HR\Application;
 use App\Models\HR\Round;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Input;
 
 class ApplicationController extends Controller
 {
@@ -17,7 +18,7 @@ class ApplicationController extends Controller
     public function index()
     {
         return view('hr.application.index')->with([
-            'applications' => self::getApplicationsList(),
+            'applications' => Application::filterByStatus(request()->get('status'))->appends(Input::except('page')),
             'status' => request()->get('status'),
         ]);
     }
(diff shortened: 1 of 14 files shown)


app/Http/Controllers/HR/ApplicationController.php
```diff
@@ -109,4 +109,24 @@ public function destroy(Application $application)
     {
         //
     }
+
+    /**
+     * get list of applications based on their show status
+     * @return \Illuminate\Database\Eloquent\Collection
+     */
+    public function getApplicationsList()
```
bohraji: This piece is pretty specific to the `Application` model. Can we move it to the model and have it called here. Any specific changes that is required to be done to the data that has been retrieved, before it is passed to the controller method to work with views, can be done in this method.

Also, there is a possibility that we'll be by default fetching the non reject application, so we can have it as default.

app/Http/Controllers/HR/ApplicationController.php
```diff
@@ -109,4 +109,24 @@ public function destroy(Application $application)
...
+    {
+        $request = request();
+        $applications = Application::with('applicant', 'job');
+        $rejected = config('constants.hr.status.rejected.label');
+        switch ($request->get('status')) {
+            case $rejected:
+                $applications = $applications->where('status', $rejected);
```
bohraji: I see the possibility of using this filter a lot. We have a local scope for it. That would make it much more readable. Thoughts?

app/Http/Controllers/HR/ApplicationController.php
```diff
@@ -109,4 +109,24 @@ public function destroy(Application $application)
...
+                break;
+            default:
+                $applications = $applications->where('status', '!=', $rejected);
+                break;
+        }
+        return $applications->orderBy('id', 'desc')->paginate(config('constants.pagination_size'));
```
bohraji: Also, the list that we retrieve is mostly paginated and ordered. We can add that as a global or local scope. Thoughts?

app/Models/HR/Applicant.php
```diff
@@ -23,39 +20,35 @@ class Applicant extends Model
...
-            'reviews' => $request->input('reviews'),
-        ]));
-        return $updated;
-    }
-
-    public function getApplicantRound($round_id)
+    public function openApplications()
     {
-        return $this->applicantRounds->where('hr_round_id', $round_id)->first();
+        return $this->applications->where('status', '!=', config('constants.hr.status.rejected.label'));
```
bohraji: We can use the model scope defined above here.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

app/Http/Controllers/HR/ApplicationController.php
rathorevaibhav: Agreed. It makes sense. Will proceed with the changes.

rathorevaibhav: I think a global scope will probably make more sense here. Considering that, the changes will happen in other modules as well. I think we should target it separately. @bohraji thoughts?

Issue #223, opened 2018-05-16
Application round cards should be collapsible


rathorevaibhav:
Right now, the cards for rounds in application edit are not collapsible and take full height. The window scroll height is too large for application with many rounds. We need to make them collapsible except the latest round card.

Issue #214, opened 2018-05-16
Each application round should show the interviewer set for that round. Conducted rounds should show who conducted that round.


rathorevaibhav:
In the application screen (below), there's no way to see which interviewer is assigned or which person conducted this round. There should be a way to see that.

![image](https://user-images.githubusercontent.com/12053186/40097925-80a7f1b6-58f6-11e8-850f-e2001b901b9b.png)

PR #224, opened 2018-05-16
(#223) and (#214)

rathorevaibhav:
#223, #214

Changes include:

1. Made application round card as collapsible
2. Showing conducted/scheduled person/date accordingly for each application round

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -281,3 +281,12 @@ input:focus + .slider {
     top: 20%;
     font-size: 12px;
 }
+
+#page_hr_applicant_edit {
+    .card-body,
+    .card-footer {
+        &.collapse {
+            display: none;
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


resources/assets/sass/app.scss
```diff
@@ -281,3 +281,12 @@ input:focus + .slider {
     top: 20%;
     font-size: 12px;
 }
+
+#page_hr_applicant_edit {
+    .card-body,
+    .card-footer {
+        &.collapse {
```
bohraji: Are we not using the the bootstrap accordion component to toggle cards? If that is so, then we do not need to write custom css to show and hide components.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

resources/assets/sass/app.scss
rathorevaibhav: @bohraji yes I tried to implement Bootstrap's accordion but it seemed like too many changes. But after giving it a fresh look, I was able to accommodate it successfully. Have implemented the accordion functionality. Have a look.

Issue #61, opened 2018-04-05
Make sure for quick response


PrateekNarang:
Most of the companies just reply with automated email ones the candidate applies. It would be great if we can connect, personally, within 24 hours with that candidate.

For this, there will be PUSH to the responsible person to contact the candidate, in person. A system email stating that He should call or connect within that time frame.

There can be a link to the 'suggestive talk' which we need to do with the candidate.

PR #231, opened 2018-05-18
(#61) Ability to schedule datetime and person

rathorevaibhav:
#61 

Changes include:

1. Added scheduled date and scheduled person fields in order to change/update. This'll be used to send reminders in the next PR.

### DIFF
--- a/app/Models/HR/ApplicationRound.php
+++ b/app/Models/HR/ApplicationRound.php
@@ -25,14 +25,6 @@ class ApplicationRound extends Model
 
     public function _update($attr)
     {
-        if ($attr['action'] == 'schedule-update')  {
-            $this->update([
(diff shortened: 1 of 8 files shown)


app/Models/HR/ApplicationRound.php
```diff
@@ -25,6 +25,14 @@ public static function _create($attr)
 
     public function _update($attr)
     {
+        if ($attr['action'] == 'schedule-update')  {
```
bohraji: This should be part of the switch case for all the actions that we have defined below.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

rathorevaibhav:
@bohraji done with the changes. I also found a bug in the date format. fixed that also. Assigning back for review.

PR #238, opened 2018-05-19
(#61) Sending notification to the scheduled user

rathorevaibhav:
For #61 

Duplicate of #236 which was mistakenly merged. #236 had been reverted.

Changes include:

1. Using Laravel's default notification to send mail to interviewer set for the new round. This is only triggered if an application round is marked as confirmed.


app/Models/HR/ApplicationRound.php
```diff
@@ -46,13 +47,14 @@ public function _update($attr)
                 $fillable['round_status'] = 'confirmed';
                 $application->markInProgress();
                 $nextApplicationRound = $application->job->rounds->where('id', $attr['next_round'])->first();
-                $scheduledPersonId = $nextApplicationRound->pivot->hr_round_interviewer_id;
+                $scheduledPersonId = $nextApplicationRound->pivot->hr_round_interviewer_id ?? config('constants.hr.defaults.scheduled_person_id');
```
bohraji: Can we have the default user for the round set in the database, maybe in the rounds table. Two advantages of doing so-
- Since we are storing the person (user) ID against the round ID, it would be very easy to see who is scheduled to take the interview of which round.
- Having it in the database makes it intuitive to add a default for the round. Also, when the software will be run by other organisation, they can have their own config in the database.

app/Models/HR/ApplicationRound.php
```diff
@@ -46,13 +47,14 @@ public function _update($attr)
...
                 $applicationRound = self::_create([
                     'hr_application_id' => $application->id,
                     'hr_round_id' => $attr['next_round'],
                     'scheduled_date' => Carbon::now()->addDay(),
-                    'scheduled_person_id' => $scheduledPersonId ?? config('constants.hr.defaults.scheduled_person_id'),
+                    'scheduled_person_id' => $scheduledPersonId,
                 ]);
+                User::find($scheduledPersonId)->notify(new ApplicationRoundScheduled($applicationRound));
```
bohraji: We are sending an email to the scheduled user, every-time an application round database entry is created. We can put this `notification` inside the database model observer while listening to the `created` event.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

app/Models/HR/ApplicationRound.php
rathorevaibhav: I agree with the above two pointers. It should have been better if the default round interviewers are set in the database.

However, recalling the time when this config was added to the system, this is supposed to be a temporary config. The jobs are targeted to be created from the portal itself and at the time of job creation, the interviewers for each round will be required. That means we won't have any `job_round` with a `NULL` round_interviewer.

But in any case, I think having a default interviewer set for a round is better. Since that is not directly linked to the notification issue, I think we can target this separately. Thoughts? @bohraji

app/Models/HR/ApplicationRound.php
rathorevaibhav: Great! I almost forgot we also have a model observer. I was targeting to implement a custom event-listener for this today.

rathorevaibhav: I'll proceed with the changes.

rathorevaibhav:
@bohraji we don't have queue enabled on Production. I think we will go for AWS SQS. Let me know if you have any preference on this.

To keep track of enabling queues, I've created an issue #247

app/Models/HR/ApplicationRound.php
rathorevaibhav: Created tracking #248

bohraji:
@rathorevaibhav Amazon SQS is good, but will come at extra cost. To start with we can go with REDIS or Database based queues, and scale as need arises. Thoughts?

rathorevaibhav:
@bohraji I think Redis will require some installation effort on our primary servers. Not sure if it is already installed. If it is, that would be helpful.

In any case, I'll have to compare the installation effort first and then will proceed with whichever is more efficient for now.

rathorevaibhav:
@bohraji review feedback implemented. ready for review again.

app/Providers/ApplicationRoundModelServiceProvider.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace App\Providers;
+
+use Illuminate\Support\ServiceProvider;
+use App\Models\HR\ApplicationRound;
+use App\Observers\HR\ApplicationRoundObserver;
+
+class ApplicationRoundModelServiceProvider extends ServiceProvider
```
bohraji: I think this is very specific to the ApplicantRoundModel, for adding only a model observer. I would recommend having something generic, such as a `DatabaseServiceProvider`, which can be used to add other observers also in the same file.

rathorevaibhav:
@bohraji changed the service provider name to DatabaseServiceProvider. Let me know if there's anything else.

PR #259, opened 2018-05-22
#254 inline resume

bohraji:


resources/views/hr/application/inline-resume.blade.php
```diff
@@ -0,0 +1,16 @@
+<a href="#" data-trigger="resume"><i class="fa fa-file fa-2x"></i></a>
```
rathorevaibhav: Not completely sure, but I don't think to have an anchor tag with href=# fits well here.
If we only need to show/hide the iFrame, I think a span/div would do the work. And that could be the reason why Bootstrap uses div for modal toggles.

bohraji:
@rathorevaibhav i have updated the feedbacks. Please have a look.
Thanks for the help 😄

resources/views/hr/application/inline-resume.blade.php
```diff
@@ -0,0 +1,16 @@
+<span class="c-pointer text-primary" @click="toggleResume()"><i class="fa fa-file fa-2x"></i></span>
+<div class="card border-dark inline-resume" id="sliding_resume" v-bind:class="{ hidden: isHidden }">
```
rathorevaibhav: `v-bind:class="{ hidden: isHidden }` can be replaced by `v-show="!isHidden"`

resources/views/hr/application/inline-resume.blade.php
```diff
@@ -0,0 +1,16 @@
...
+    <div class="card-header bg-warning">
+        <ul class="nav justify-content-end">
+            <li class="nav-item">
+                <b><a class="nav-link py-0" href="{{ $resume }}" target="_blank"><i class="fa fa-external-link" aria-hidden="true"></i> Open in a New Tab</a></b>
+            </li>
+            <li class="nav-item">
+                <b><span class="c-pointer text-primary" href="#" @click="toggleResume()">Close</span></b>
```
rathorevaibhav: remove href

resources/assets/js/app.js
```diff
@@ -28,11 +28,15 @@ if (document.getElementById('page_hr_applicant_edit')) {
     const applicantEdit = new Vue({
         el: '#page_hr_applicant_edit',
         data: {
+            isHidden: true,
```
rathorevaibhav: can we rename this? This element `page_hr_applicant_edit` contains multiple data and elements. It'll be better if we make this variable name and methods more specific to their use. something like `hideResumeFrame` and `toggleResumeFrame`

Issue #250, opened 2018-05-21
Builds failing


rathorevaibhav:
Looking at the TravisCI logs,

![image](https://user-images.githubusercontent.com/12053186/40295394-b7e98370-5cf6-11e8-9e08-f3c1f9e7aebc.png)

There is something wrong in one of the migrations for `application_meta` table. Not sure how it did not give any error on dev.

The builds started failing from here.
https://travis-ci.org/ColoredCow/employee-portal/builds/381012397

PR #260, opened 2018-05-22
#250 update json column to text for support across mysql versions

bohraji:


bohraji:
@rathorevaibhav Found the bug for failed builds Solution based on [link](https://github.com/spatie/laravel-medialibrary/issues/563)

rathorevaibhav:
Great @bohraji 🙌

rathorevaibhav:
To clarify, we don't have to run the migration on our environments, right? @bohraji

bohraji:
@rathorevaibhav I changed the existing migration and the type of the column. We'll need to perform a rollback and then migrate again. That will bring all the database in a consistent state. If a lot of users have run it, then we can create a new migration.

bohraji:
@rathorevaibhav Added new migration and update existing one. This was done as the Travis check was breaking in creating the original JSON type column. Also, new migration added so that systems that have data in the system can also update their column and keep the project consistent.

rathorevaibhav:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] Usage of recursive methods, loops, and algorithms.
- [ ] For coupling

PR #267, opened 2018-05-22
(#266) Reject button not working

rathorevaibhav:
#266 

Changes include:

1. Checking validation for `next_scheduled_date` and `next_scheduled_person_id` only if action is confirm.


rathorevaibhav:
@bohraji going ahead with the hotfix merge. Please take a look at the changes when you have time. Will target feedback separately.

PR #274, opened 2018-05-23
(#268) job to internship

rathorevaibhav:
#268 

Changes include:

1. Change schema for the application_meta table. Update to a key-value structure.
2. For any job application, the system will suggest if the application can be moved to the internship. On taking this action, the job id changes and corresponding meta is saved in application meta table for track history.
3. Showing the job changed event in the applicant (application) timeline.


app/Mail/HR/Application/JobChanged.php
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Create a new message instance.
+     *
+     * @param Application $application
+     * @param ApplicationMeta $changeJobMeta
+     */
+    public function __construct(Application $application, ApplicationMeta $changeJobMeta)
+    {
+        $this->application = $application;
```
bohraji: Please eager load application's applicant info, as it is used below.

app/Models/HR/Applicant.php
```diff
@@ -87,4 +89,14 @@ public function timeline()
         }, $timeline), SORT_ASC, $timeline);
         return $timeline;
     }
+
+    /**
+     * Suggests if applicant can be set for internship if the applicant is not graduating this year
+     *
+     * @return boolean
+     */
+    public function suggestInternship()
```
bohraji: Can we update the name of this function to something generic like `inCollege`, or `graduated`, etc. That would enable us to use this comparison in other places and not limit to checking internship.

app/Models/HR/Application.php
```diff
@@ -169,7 +173,7 @@ public function markInProgress()
      */
     public function timeline()
     {
-        $this->load('applicationRounds', 'applicationRounds.round');
+        $this->load('applicationRounds', 'applicationRounds.round', 'applicationMeta');
```
bohraji: 👍

app/Models/HR/Application.php
```diff
@@ -181,6 +185,45 @@ public function timeline()
                 ];
             }
         }
+
+        // adding change-job events in the application timeline
+        $jobChangeEvents = $this->applicationMeta()->jobChanged()->get();
```
bohraji: This will again make a request to the database to get the values of the required `applicationMeta`. In this case then, we do not need to `eager load` the  whole `applicationMeta` relationship, as done above on line `176`. This constrained eager loading is fine.

app/Models/HR/Application.php
```diff
@@ -169,7 +173,7 @@ public function markInProgress()
      */
     public function timeline()
```
bohraji: The way this `applicant` timeline is getting generated is currently not very optimised. We start with an applicant, go over each of the application and capture the events. Now we have the job change events also. All this bring in a lot of loops. Then the sorting of all the data is also performance intensive task.

I would suggest exploring the option of having a `ApplicantTimeline` model, that captures all the actions that are taken on an applicant in the order they happen. This model can have a polymorphic relationship with application, application rounds, application meta, etc. stored in the order the events occur. I see the following advantages-
- The information will already be sorted. So no need to sort info again.
- All the info can be eager loaded at once.
- We will have all the applicant activity tracked in a single place, whose detail is spread across different specialised table.
- The timeline can be filtered, sorted, etc. if required.

Let me know your thoughts on this. We can track it in a later build, but I see this update required as we are generating a lot of data, and deferring it for much later will increase the data migration effort.

app/Models/HR/Application.php
```diff
@@ -181,6 +185,45 @@ public function timeline()
...
+            'new_job' => $attr['hr_job_id'],
+            'job_change_mail_subject' => $attr['job_change_mail_subject'],
+            'job_change_mail_body' => ContentHelper::editorFormat($attr['job_change_mail_body']),
+            'user' => Auth::id(),
+        ];
+
+        $this->update(['hr_job_id' => $attr['hr_job_id']]);
+        return ApplicationMeta::create([
+            'hr_application_id' => $this->id,
+            'key' => 'change-job',
```
bohraji: Can we have the different key's that we now have stored in a `config` file.

app/Models/HR/ApplicationMeta.php
```diff
@@ -7,12 +7,22 @@
...
     protected $table = 'hr_application_meta';
 
     public function application()
     {
     	return $this->belongsTo(Application::class, 'hr_application_id');
     }
+
+    public static function scopeFormData($query)
+    {
+        return $query->where('key', 'form-data');
```
bohraji: 👍 for scopes.
Lets add these values form `Config`

resources/views/hr/application/edit.blade.php
```diff
@@ -66,17 +66,20 @@
                         </div>
                         <div class="form-group offset-md-1 col-md-5">
                             <b>Graduation Year</b>
-                            <div>{{ $applicant->graduation_year ?? '-' }}</div>
+                            <div>
+                                {{ $applicant->graduation_year ?? '-' }}&nbsp;
+                                @if (isset($suggestInternship) && $suggestInternship)
```
bohraji: We can move the trigger for the model, also in the blade file that is included. And then we can use `@includeIf` directive to reduce these lines to a single one. The included file will come with the model and its trigger as they both are dependent on each other and having them included together can help.

Thoughts?

resources/views/hr/application/timeline.blade.php
```diff
@@ -22,10 +22,39 @@
                         <b><u>{{ date(config('constants.display_date_format'), strtotime($applicationRound->conducted_date)) }}</u></b><br>
                         {{ $applicationRound->round->name }} for {{ $application->job->title }} conducted by {{ $applicationRound->conductedPerson->name }}<br>
                         @if ($applicationRound->mail_sent)
-                            <span data-toggle="modal" data-target="#round_mail_{{ $applicationRound->id }}" class="{{ config("constants.hr.status.$applicationRound->round_status.class") }} modal-toggler">Communication mail</span><br>
-                            @include('hr.round-review-sent-mail-modal', [ 'applicationRound' => $applicationRound ])
+                            @php
+                                $data = [
```
bohraji: Can we have this all inside a function in the `ApplicationRound` model. That will be directly accessed here and the info assigned to the variable or directly passed to the model.
`
$applicationRoundMailRender = $applicationRound->mailData;
`

resources/views/hr/application/timeline.blade.php
```diff
@@ -22,10 +22,39 @@
...
                         @break
+                    @case('job-changed')
+                        @php
+                            $jobChangeEvent = $item['jobChangeEvent'];
+                        @endphp
+                        <b><u>{{ date(config('constants.display_date_format'), strtotime($item['date'])) }}</u></b><br>
+                        Moved from {{ $jobChangeEvent->value->previous_job }} to {{ $jobChangeEvent->value->new_job }}<br>
+                        @php
+                            $data = [
+                                'modal-id' => 'job_change_' . $jobChangeEvent->id,
```
bohraji: Move this to a model or a helper. We should be only concerned with rendering the data in the view files.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

app/Models/HR/Application.php
rathorevaibhav: Thanks. Will proceed with the changes.

app/Http/Controllers/HR/Applications/ApplicationController.php
```diff
@@ -46,21 +49,46 @@ public function index()
...
+     * @param integer $id
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function update(ApplicationRequest $request, int $id)
+    {
+        $validated = $request->validated();
+        $application = Application::findOrFail($id);
+
+        switch($validated['action']) {
+            case 'change-job':
```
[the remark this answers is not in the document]
rathorevaibhav: Agreed. Created tracking #276

app/Http/Controllers/HR/Applications/ApplicationController.php
```diff
@@ -46,21 +49,46 @@ public function index()
...
+                $changeJobMeta = $application->changeJob($validated);
+                Mail::send(new JobChanged($application, $changeJobMeta));
+                return redirect()->route('applications.internship.edit', $id)->with('status', 'Application successfully moved to internship!');
```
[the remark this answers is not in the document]
rathorevaibhav: 👍

app/Models/HR/Application.php
rathorevaibhav: I agree. I was having similar thoughts as the timeline is now having multiple sources and the current implementation is not optimized. Moreover, I was thinking to create a vertically growing table but wasn't sure if that would be great. Maybe we can start our discussion there. haven't looked at the polymorphic relationship but will take a look at it first. Thanks for the suggestion.

rathorevaibhav: Created tracking #277

resources/views/hr/application/edit.blade.php
rathorevaibhav: Aah yes. Thanks. will do that. That's an easy change.

Issue #265, opened 2018-05-22
Default template for round emails.


rathorevaibhav:
After an interviewer conducts a round, he needs to send a communication mail to the applicant. Now, for each type of round, for example resume screening, a similar mail goes to all the applicants who cleared resume screening.

We can create a template as default round communication template. This will get prepopulated when someone clicks on the "send mail" button.

PR #283, opened 2018-05-24
(#265) round default template

rathorevaibhav:
#265

Changes include:

1. Added migration to add two new columns in `hr_rounds` table. To store confirmed and rejected mail template.
2. View changes to save these default templates from the settings tab.
3. Showing template when the user clicks on send mail button from the application round screen.


app/Models/HR/Round.php
```diff
@@ -7,10 +7,20 @@
...
+    protected $fillable = ['name', 'guidelines', 'confirmed_mail_template', 'rejected_mail_template'];
 
     protected $table = 'hr_rounds';
 
+    /**
+     * The attributes that should be cast to native types.
+     *
+     * @var array
+     */
+    protected $casts = [
```
bohraji: 👍

app/Http/Controllers/SettingController.php
```diff
@@ -19,7 +20,14 @@ public function index(String $module)
         $this->authorize('view', Setting::class);
 
         $settings = Setting::where('module', $module)->get()->keyBy('setting_key');
-        return view("settings.$module")->with(['settings' => $settings]);
+        return view("settings.$module")->with([
+            'settings' => $settings,
+            'rounds' => Round::all(),
+            'roundMailTypes' => [
```
bohraji: This setting controller is very generic and deals with different modules of the whole application. Returning `rounds` info and also `round mail type` does not appear very extendible and generic.

Can we have a separate route, or route parameter or filters, etc that connects or specifies the need to return `roundMailTypes`.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

bohraji:
Also, @rathorevaibhav please ensure that the `travis` check have been approved before merging.

rathorevaibhav:
@bohraji done with the changes.

PR #288, opened 2018-05-24
282 checklist round review

bohraji:


app/Models/HR/ApplicationRoundReviewEvaluation.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace App\Models\HR;
+
+use App\Models\HR\ApplicationRound;
```
rathorevaibhav: unused `use`. I think this was meant to be `ApplicationRoundReview`

app/Models/HR/EvaluationParameterOption.php
```diff
@@ -0,0 +1,24 @@
...
+namespace App\Models\HR;
+
+use App\Models\HR\EvaluationParameter;
+use App\Models\HR\Round;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
+
+class EvaluationParameterOption extends Model
+{
+	use SoftDeletes;
```
rathorevaibhav: fix the indentation.

database/migrations/2018_05_24_093123_add_evaluation_parameter_table.php
```diff
@@ -0,0 +1,84 @@
...
+            $table->softDeletes();
+            $table->timestamps();
+        });
+
+        Schema::create('hr_round_evaluation', function (Blueprint $table) {
+            $table->integer('evaluation_id')->unsigned();
+            $table->integer('round_id')->unsigned();
+        });
+
+        Schema::create('hr_application_round_review_evaluations', function (Blueprint $table) {
```
rathorevaibhav: I see that an evaluation is directly associated to a round (`hr_round_evaluation`). Since an application round points to a round, I think that evaluation should also be directly related to an application round and not to an application round review.

The table `application_round_reviews` was originally created keeping in mind multiple evaluation parameters that an application round can have. That's why it is vertically growing with a key-value pair. (key is the evaluation criteria and value is the decision taken). I see that now we can completely remove this and use `hr_application_round_evaluations` instead.

bohraji: @rathorevaibhav 
Elaborating my understanding below-
The `hr_application_round_reviews` table was designed to store multiple types of reviews (currently only of type `feedback`). With the new table `hr_application_round_review_evaluations `, where we are going to store the different parameters related to a particular application stage, the original `hr_application_round_reviews ` table is now redundant and we should move its data to `hr_application_round_review_evaluations `, delete the previous table and also rename the new table to `hr_application_round_evaluations `.

Let me know if the understanding is correct.

rathorevaibhav: That is completely correct. 👍

rathorevaibhav:
@bohraji so we agree on the migrating data from `hr_application_round_reviews` to `hr_application_round_review_evaluations ` (which needs to be renamed to `hr_application_round_evaluations`, I think the changes should be included in this PR. Thoughts?

Issue #293, opened 2018-05-25
Reports link is incorrect


rathorevaibhav:
- [x] fix the link in Primary Navigation
- [x] fix the link in home.blade.php

PR #294, opened 2018-05-25
(#293) reports link correction

rathorevaibhav:


rathorevaibhav:
@bohraji As this is a critical one, going ahead with the merge. Though the changes are straightforward, please have a look at these anyway.

Issue #290, opened 2018-05-25
On hold applications should show up in a separate tab.


rathorevaibhav:
In the application listing page, there needs to be another tab for `On-hold` applications. This will open up a new page and will show only the on-hold applications. That means that the `Active` tab will only show `in-progress` applications.

![image](https://user-images.githubusercontent.com/12053186/40532786-3d947046-601e-11e8-9588-4af93d556236.png)

Also, rename the following texts in that menu:

- [x] `Active` to `Open`
- [x] `Rejected` to `Closed`

rathorevaibhav:
This is complete. Closing.

PR #305, opened 2018-05-28
On hold applications tab

wevekamola:
In PR #290 `On hold` tab has been added to the application dashboard and also `Active` tab has been renamed to `Open` and `Rejected` tab has been renamed to `Closed`


app/Models/HR/Application.php
```diff
@@ -150,6 +153,22 @@ public function scopeNonRejected($query)
         return $query->where('status', '!=', config('constants.hr.status.rejected.label'));
     }
 
+    /**
+     * get applications where status is new and in-progress
+     */
+    public function scopeOpenApplication($query)
```
rathorevaibhav:
```
Application::with('applicationRounds')->openApplication()->get();
```
has repetitive term `application`. 

If we rename the scope to scopeIsOpen then the following query is more readable and meaningful.
```
Application::with('applicationRounds')->isOpen()->get();
```

app/Models/HR/Application.php
```diff
@@ -93,8 +93,11 @@ public function scopeFilterByStatus($query, $status)
             case config('constants.hr.status.rejected.label'):
                 $query->rejected();
                 break;
+            case config('constants.hr.status.on-hold.label'):
+                $query->onHold();
+                break;    
             default:
-                $query->nonRejected();
```
rathorevaibhav: Now that we don't need nonRejected scope anymore, I think we can remove it altogether?

app/Models/HR/Application.php
```diff
@@ -150,6 +153,22 @@ public function scopeNonRejected($query)
...
+     */
+    public function scopeOpenApplication($query)
+    {
+        return $query->whereIn('status', array(config('constants.hr.status.new.label'), config('constants.hr.status.in-progress.label')));
+    }
+
+    /**
+     * get applications where status is on-hold
+     */
+    public function scopeOnHold($query)
```
rathorevaibhav: 👍

app/Models/HR/Application.php
wevekamola: @rathorevaibhav We are using `nonRejected` scope in other files also `Applicant.php` and `ApplicationController.php` i don't think it's a good idea to remove it. Let me know your thoughts.

app/Models/HR/Application.php
wevekamola: 👍

app/Models/HR/Application.php
wevekamola: @rathorevaibhav Yeah we should only count `new ` and `in-progress` as open application. Is that what you are saying?

rathorevaibhav: Yes. openApplications = `new` and `in-progress` applications

That means, there's no section where we need to show nonRejected (earlier openApplications = `nonRejected`). So we can remove nonRejected from everywhere.

PR #319, opened 2018-05-30
(#311) no show

rathorevaibhav:
#311 

Changes include:

1. Basically renamed `round-not-conducted` to `no-show`. Upcoming changes will be following this.


app/Models/HR/Application.php
```diff
@@ -202,14 +202,14 @@ public function timeline()
         }
 
         // adding round-not-conducted events in the application timeline
```
bohraji: Update comment, to show the updated action

PR #327, opened 2018-05-31
(#311) no show scheduling

rathorevaibhav:
#311 

Changes include:

1. Artisan command that changes application status to no-show if an application round that was scheduled 2 hours ago is still not conducted.
2. Task scheduler that calls the above artisan command every 30 minutes.
3. Changes in applicant timeline when the cron runs.

Steps for deployment:
1. Add the following cronjob
```
* * * * * php path_to_employee_portal/artisan schedule:run >> /dev/null 2>&1
```

### DIFF
--- a/app/Console/Commands/HR/ApplicationNoShow.php
+++ b/app/Console/Commands/HR/ApplicationNoShow.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace App\Console\Commands\HR;
+
+use Illuminate\Console\Command;
+use Carbon\Carbon;
+use App\Models\HR\ApplicationRound;
+use App\Models\HR\ApplicationMeta;
+
+class ApplicationNoShow extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'application:no-show';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Set application status to no-show if an application round is not conducted 2 hours after scheduled time';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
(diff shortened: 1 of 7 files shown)


app/Console/Commands/HR/ApplicationNoShow.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $applicationRounds = ApplicationRound::with('application')
+                            ->whereNull('round_status')
+                            ->where('scheduled_date', '<=', Carbon::now()->subHours(config('constants.hr.no-show-hours-limit'))->toDateTimeString())
+                            ->get();
+
+        foreach ($applicationRounds as $applicationRound) {
+            $application = $applicationRound->application;
+            if ($application->status != config('constants.hr.application-meta.keys.no-show')) {
+                $application->update([
```
bohraji: This part is specific to all the updates that are required to mark an application as a `no show`. Can we move it in a function, that we can call on the application object and make the changes to mark is as `no show`. With that, we can reuse the same functionality later in other areas also.

app/Models/HR/ApplicationMeta.php
```diff
@@ -42,8 +42,8 @@ public function getCommunicationMailAttribute()
 
         $attr = [
             'mail-to' => $this->application->applicant->email,
-            'mail-sender' => $this->value->user,
-            'mail-date' => $this->created_at,
+            'mail-sender' => $this->value->user ?? null,
```
bohraji: Having null attributes can cause the email api to throw exceptions. I do not see an update for handling that. 
Thoughts?

bohraji: This has been handled in blade file below. Ignore comment above.

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

app/Console/Commands/HR/ApplicationNoShow.php
rathorevaibhav: That is a good idea. I'll proceed with the changes.

rathorevaibhav:
@bohraji yes the cronjob is accounted for. Have added it in the steps for deployment.

Issue #315, opened 2018-05-29
Job Preview


rathorevaibhav:
- [ ] Remove the column "See job" from the job list
- [ ] This see option should be displayed as "Preview Job" button and should come up in job edit screen.

mohitgusain:
Followings are  the issues under 315
1. First to remove the see job from the job list and replace it by the "Preview  Job " button
2. Same "Preview  Job " button should also come up into job edit screen

PR #344, opened 2018-06-01
Changes are completed according to job preview #315

mohitgusain:
#315

- First to remove the see job from the job list and replace it by the "Preview Job " button
- Same "Preview Job " button should also come up into job edit screen

### DIFF
--- a/resources/views/hr/job/index.blade.php
+++ b/resources/views/hr/job/index.blade.php
@@ -10,7 +10,7 @@
         <tr>
             <th>Job title</th>
             <th>Type</th>
-            <th>Job description</th>
+            
             <th>Total applicants</th>
         </tr>
         @foreach ($jobs as $job)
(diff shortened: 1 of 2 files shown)


resources/views/hr/job/index.blade.php
```diff
@@ -22,7 +22,7 @@
                 {{ ucfirst($job->type) }}
             </td>
             <td>
-                <a href="{{ $job->link }}" target="_blank">See</a>
+                <a href="{{ $job->link }}" target="_blank" class="btn btn-info" role="button">Preview job</a>
```
rathorevaibhav: We don't need the "Preview job" as a column in the table. It should be visible only in the Job edit screen. That means table column needs to be removed altogether.

resources/views/hr/job/index.blade.php
```diff
@@ -10,7 +10,7 @@
         <tr>
             <th>Job title</th>
             <th>Type</th>
-            <th>Job description</th>
+
```
rathorevaibhav: remove these extra lines.

resources/views/hr/job/index.blade.php
```diff
@@ -21,9 +21,7 @@
             <td>
                 {{ ucfirst($job->type) }}
             </td>
-            <td>
-                <a href="{{ $job->link }}" target="_blank">See</a>
-            </td>
+
```
rathorevaibhav: remove this line. No need to add any redundant changes.

PR #348, opened 2018-06-02
Search books

pankaj-ag:
- Added search functionality in books
- removed _created and _updated methods from Book modal for consistency.

### DIFF
--- a/app/Http/Controllers/KnowledgeCafe/Library/BookController.php
+++ b/app/Http/Controllers/KnowledgeCafe/Library/BookController.php
@@ -24,7 +24,8 @@ class BookController extends Controller
     public function index()
     {
         $this->authorize('list', Book::class);
-        $books = Book::with('categories')->orderBy('title')->paginate(config('constants.pagination_size'));
+        $searchString = (request()->has('search')) ? request()->input('search'): false;
+        $books = Book::getList($searchString);
         $categories = BookCategory::orderBy('name')->get();
         return view('knowledgecafe.library.books.index', compact('books', 'categories'));
     }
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/KnowledgeCafe/Library/BookController.php
```diff
@@ -78,12 +81,21 @@ public function edit(Book $book)
...
+        if (!isset($validatedData['categories'])) {
+            return response()->json([
+                'isUpdated' => false
+            ]); 
+        }
+
+        $categories = array_pluck($validatedData['categories'], 'id');
         return response()->json([
-            'success' => $book->_update($request->validated())
+            'isUpdated' => $this->categories()->sync($categories)
```
bohraji: This update is very specific to updating the categories of a book, which will one of the things can be updated. If in future, we need to update the property of a book, like name, or author this will need to be changed again.
Please update.

resources/views/knowledgecafe/library/books/index.blade.php
```diff
@@ -62,8 +85,8 @@
                 </div>
             </td> 
             @can('library_books.delete')
-            <td>
-                <div @click="deleteBook(index)" class="text-danger c-pointer ml-3"><i class="fa fa-times"></i>&nbsp;Delete</div>
+            <td class="action_column">
```
bohraji: The columns of a table take their width on their own. Why did we add a specific class and hard code a value?

bohraji:
Checked this PR :

- [ ] For duplication.
- [ ] For performance. 
- [ ] For coupling

resources/views/knowledgecafe/library/books/index.blade.php
pankaj-ag: removed

PR #352, opened 2018-06-02
#(books api)

pankaj-ag:
#354
- Show all the books on coloredcow.com.
- Wrote some API routes for access book data.

### DIFF
--- a/app/Http/Controllers/KnowledgeCafe/Library/BookController.php
+++ b/app/Http/Controllers/KnowledgeCafe/Library/BookController.php
@@ -173,7 +173,8 @@ class BookController extends Controller
 
         $books = (request()->has('cat')) ? 
                 Book::getByCategoryName(request()->input('cat')) : 
-                Book::getList();
+                Book::with(['categories'])->orderBy('title')
+                ->get();
 
         $data = [];
         foreach ($books as $book) {
--- a/app/Http/Controllers/KnowledgeCafe/Library/BookController.php
+++ b/app/Http/Controllers/KnowledgeCafe/Library/BookController.php
@@ -181,8 +181,10 @@ class BookController extends Controller
            ];
            $data['books'][] = $customBookData;
         }
-        
-        $data['categories'] = BookCategory::has('books')->pluck('name')->toArray();
+
+        $data['categories'] = BookCategory::has('books')
+                              ->pluck('name')->toArray();
+                              
         return response()->json($data);
     }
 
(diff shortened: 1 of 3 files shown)


rathorevaibhav:
@pankaj-ag please change the base branch for this PR.

Also, please mention the issue this PR is trying to resolve. That'll help to understand the requirement completely.

app/Http/Controllers/KnowledgeCafe/Library/BookController.php
```diff
@@ -168,5 +168,25 @@ public function formatBookData($book) {
...
+    public function getBookList() {
+
+        $books = (request()->has('cat')) ? 
+                Book::getByCategoryName(request()->input('cat')) : 
+                Book::with(['categories'])->orderBy('title')
+                ->get();
+
+        $data = [];
+        foreach ($books as $book) {
+            $customBookData = $book->toArray();
```
rathorevaibhav: can we be more descriptive here? customBookData doesn't define what it is supposed to store. If it helps, please add proper comments to describe in detail.

app/Models/KnowledgeCafe/Library/Book.php
```diff
@@ -30,5 +30,13 @@ public function _update($data) {
     public function categories() {
         return $this->belongsToMany(BookCategory::class, 'library_book_category', 'library_book_id', 'book_category_id');
     }
+
+    public static function getByCategoryName($categoryName) {
+        if(!$categoryName) {
+            return false;
+        }
+        $category = BookCategory::where('name', $categoryName)->first();
```
rathorevaibhav: I'm not sure but since you're in Book model, ideally you should query the Books and filter them based on category. This approach may fit well if it was written in `BookCategory.php`. Thoughts @pankaj-ag ?

app/Http/Controllers/KnowledgeCafe/Library/BookController.php
```diff
@@ -168,5 +168,25 @@ public function formatBookData($book) {
...
+    public function getBookList() {
+
+        $books = (request()->has('cat')) ? 
+                Book::getByCategoryName(request()->input('cat')) : 
+                Book::with(['categories'])->orderBy('title')
+                ->get();
+
+        $data = [];
+        $i = 0;
+        foreach ($books as $book) {
```
rathorevaibhav: Just a doubt, can we use to track the indexes?
```
foreach ($books as $index => $book) {
```

pankaj-ag: Thanks, @rathorevaibhav. I have made this change.

PR #356, opened 2018-06-04
Mark book as read

pankaj-ag:
#364

- Give read/unread option on book info page.
- UI changes as per requirnments
- Show a Random book when the user comes on the home screen.


app/Models/KnowledgeCafe/Library/Book.php
```diff
@@ -37,4 +38,21 @@ public static function getByCategoryName($categoryName) {
...
+         $result = ($read) ? $this->readers()->attach(auth()->user()) 
+                  : $this->readers()->detach(auth()->user());
+
+        return true;
+    }
+
+    public static function getRandomUnreadBook() {
+        return self::whereDoesntHave('readers', function ($query) {
+            $query->where('id', auth()->id());
+        })->get()->random();
```
bohraji: Can we use `inRandomOrder` and `first` eloquent functions to reduce the overhead of fetching all the books of user.

resources/assets/js/app.js
```diff
@@ -545,3 +544,53 @@ if (document.getElementById('books_category')) {
...
+        }
+    });
+}
+
+if(document.getElementById('home_page')) {
+    var el = document.getElementById("markBookAsRead");
+    el.addEventListener("click", markBookAsRead, false); 
+    let isModalShown = sessionStorage.getItem('book_modal_has_shown');
+    if(!isModalShown) {
+        sessionStorage.setItem("book_modal_has_shown", "true");
```
bohraji: 👍

bohraji:
Checked this PR :

- [x] For duplication.
- [x] For performance. 
- [x] For coupling

app/Models/KnowledgeCafe/Library/Book.php
pankaj-ag: Great suggestion. done

pankaj-ag:
@abhi1203 as we figured that we need to add a check on the server side also so that it will not make a random book query on every home URL request. I am tracking that on high priority in another issue #365.

Issue #285, opened 2018-05-24
Reminders to conduct application rounds


rathorevaibhav:
The following reminders/nudge should go to applicant or interviewer (BCC HR group):

- [x] At 8 am daily.
- [x] User should have the ability to send emails when an interview is not conducted.

The email template for the above two should be configurable.

PR #360, opened 2018-06-04
(#285) interview reminder enable

rathorevaibhav:
#285

We need to send reminders to applicants for all rounds except a few (like resume screening). So, we need a flag that'll track this.


rathorevaibhav:
@bohraji this is in continuation with #351. Need this for today's build. Please have a look at this also when you see the feedback implemented on #351

rathorevaibhav:
@bohraji this is in continuation with #351. Need this for today's build. Please have a look at this also when you see the feedback implemented on #351

Issue #371, opened 2018-06-06
Unit test for Project


rathorevaibhav:
Test the following cases:

- [x] ACL Project Module 
- [x] Create Projects
- [x] Modify Projects

rathorevaibhav:
Nice work on this @nishanth-kd @pankaj-ag. Next, we should target more on other test related issues. Closing this one.

PR #385, opened 2018-06-06
Unit test for Project (#371)

nishanth-kd:


nishanth-kd:
@rathorevaibhav I'm not sure why the TravisCI Build is failing. it works well on my machine maybe something needs to be configured on the Travis end or if I can debug on Travis and why there's a 500 on it. 

```
1) Tests\Feature\ProjectsTest::an_authorised_user_can_see_projects
Expected status code 200 but received 500.
Failed asserting that false is true.
/home/travis/build/ColoredCow/employee-portal/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestResponse.php:109
/home/travis/build/ColoredCow/employee-portal/tests/Feature/ProjectsTest.php:17
```

rathorevaibhav:
@nishanth-kd were you able to figure out why the TravisCI build is giving 500 error?

app/Http/Controllers/ProjectController.php
```diff
@@ -56,7 +56,7 @@ public function store(ProjectRequest $request)
             'status' => $validated['status'],
             'invoice_email' => $validated['invoice_email'],
         ]);
-        return redirect("/projects/$project->id/edit")->with('status', 'Project created successfully!');
+        return redirect($project->path() . '/edit')->with('status', 'Project created successfully!');
```
rathorevaibhav: I think a better approach would be to define named routes and use 
```
route('projects.edit', $project->id)
```

In case, the route has to be changed, it'll have to be modified at two places.

@nishanth-kd Thoughts?

.env.testing
```diff
@@ -1,5 +1,5 @@
 APP_ENV=testing
-APP_KEY=
+APP_KEY=base64:oMBcLYtjAwezLrQBgFNNYfDIDTLuoBkz9vn0et/fxFo=
```
rathorevaibhav: @nishanth-kd curious why this modification?

app/Http/Controllers/ProjectController.php
nishanth-kd: Yea, we can do a named routes. sounds good. Add them as well.

Issue #386, opened 2018-06-06
Update schedule for no-show-reminded applications


rathorevaibhav:
When clicking on the update schedule, if the application is no-show-reminded, it should change to in-progress in the following circumstances:

1. If only the schedule time is changed and the new value is in future.
2. If the interviewer is changed, there shouldn't be any change in the application status.

PR #388, opened 2018-06-06
(#386) no-show status change to in-progress

rathorevaibhav:
#386 

Changes include:

1. Marking a no-show application as in-progress if the scheduled time is updated for it considering that scheduled time is in future.

### DIFF
--- a/app/Models/HR/ApplicationRound.php
+++ b/app/Models/HR/ApplicationRound.php
@@ -28,12 +28,9 @@ class ApplicationRound extends Model
         switch ($attr['action']) {
             case 'schedule-update':
 
-                // if the application status is no-show or no-show-reminded, and the new schedule date is greater
-                // than the current time, we change the application status to in-progress
-                if (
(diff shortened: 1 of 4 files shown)


app/Models/HR/ApplicationRound.php
```diff
@@ -36,6 +27,15 @@ public function _update($attr)
 
         switch ($attr['action']) {
             case 'schedule-update':
+
+                // if the application status is no-show or no-show-reminded, and the new schedule date is greater
+                // than the current time, we change the application status to in-progress
+                if (
+                    in_array($application->status, [config('constants.hr.status.no-show.label'), config('constants.hr.status.no-show-reminded.label')])
```
bohraji: This module is difficult to read. Can we move it to `application`, and referrer it here in a readable function name.

Something like `$application->checkIfInProgress('$attr['scheduled_date']')`. All the specific of checking and marking an application in progress can be taken care by the `application` model.

Issue #377, opened 2018-06-06
Suggest next question from question pool based on the previous question


bohraji:
#282 
Phase 2

PR #396, opened 2018-06-08
377 question pool

bohraji:
PR for #377 
Mostly refactoring changes.

- Evaluation now are related application along with the application round
- Rename table from `application_round_evaluation` to `application_evaluation`
- Seeder to add `application_id` to the new table.
- Moving all evaluation related tables to `evaluation` folder inside HR models.
- New table to group evaluation parameter into segments and relate segments to application rounds

Commands to run at deployment
`php artisan migrate`
`php artisan db:seed --class=ApplicationEvaluationSeeder`


database/seeds/ApplicationEvaluationSeeder.php
```diff
@@ -0,0 +1,22 @@
...
+class ApplicationEvaluationSeeder extends Seeder
+{
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
+        $ApplicationEvaluations = ApplicationEvaluation::with('applicationRound')->get();
```
rathorevaibhav: Is this a one-time seeder? If not, we should add it to DatabaseSeeder.php

rathorevaibhav: Also, please specify the commands need to run at the time of deployment.

app/Models/HR/ApplicationRound.php
```diff
@@ -78,6 +79,7 @@ public function updateOrCreateEvaluation($evaluations = [])
                     [
                         'application_round_id' => $this->id,
                         'evaluation_id' => $evaluation['evaluation_id'],
+                        'application_id' => $this->hr_application_id,
```
rathorevaibhav: does an evaluation need to be associated with both application round and application? I am not sure what value it's adding since application round is already associated with an application.

rathorevaibhav: Also, it's possible that a database entry exists for which `$this->applicationRound->application` can be different from `$this->application`. Why another relation when there's already an existing one?

database/seeds/ApplicationEvaluationSeeder.php
bohraji: @rathorevaibhav this is a one time seeder. This will fill in the value for the new columns. Code updates after this deployment will capture the updates automatically for new entries.

app/Models/HR/ApplicationRound.php
bohraji: @rathorevaibhav the way we designed this feature earlier was to capture the assessment specific to the `application_round`. Over discussion, we have found that it makes more sense to have it associated with the whole application, so that the fields are not directly bounded by the application round, but rather shows the nature/evaluation of the whole application.

I have not removed the application round entry as of now, as there as some features that are still using it. Once the updates are in place, I'll remove the `application round id` column

bohraji:
@rathorevaibhav I have added my comments and also the commands to be run in deployment.

database/seeds/ApplicationEvaluationSeeder.php
rathorevaibhav: 👍

app/Models/HR/ApplicationRound.php
rathorevaibhav: @bohraji Thanks for clarifying. Looking good then. Just to make sure we don't lose the track, let's create a tracking of this (removing application round id)

bohraji:
Added comment in base issue.

Issue #389, opened 2018-06-07
Calendar integration


rathorevaibhav:
#406 please refer this PR for knowing the current progress in this issue

- [x] Ability to create a calendar event for the HR next round. The participants would be the next round interviewer and the applicant.
- [x]  Ability to add calendar event link should be attached in the mail.
- [ ]  If the schedule is updated then there should be an update on the calendar as well.
- [ ] Notification should be sent to both applicant and assigned interviewer about the schedule update.
- [x] When a calendar event is created, the Google invitation mail should get send when the calendar is created.

PR #406, opened 2018-06-10
(#389) Calendar Integration

rathorevaibhav:


app/Services/CalendarEventService.php
```diff
@@ -0,0 +1,157 @@
...
+        $client->useApplicationDefaultCredentials();
+        $client->setSubject(config('constants.gsuite.service-account-impersonate'));
+        $client->setScopes(Google_Service_Calendar::CALENDAR);
+        $this->service = new Google_Service_Calendar($client);
+    }
+
+    public function create($details, $calendarId = 'primary')
+    {
+        if (!isset($details['summary'], $details['attendees'], $details['start'], $details['end'])) {
+            throw new Exception('Invalid data for calendar creation');
```
bohraji: This exception should be caught in the implementing method above. The process should complete and a message should be sent stating that the `calendar creation` has failed.

bohraji: Also, since this is an external service integration, we need to have timeout function. If the creation is not completed in predetermined time, we should continue with an error message. We can track this in an other update.

app/Services/CalendarEventService.php
```diff
@@ -0,0 +1,157 @@
...
+            'attendees' => $this->attendees,
+            'start' => $this->startDateTime,
+            'end' => $this->endDateTime,
+        ]);
+        $event = $this->service->events->insert($calendarId, $event);
+        $this->id = $event->id;
+        $this->setHangoutLink($event->hangoutLink);
+    }
+
+    public function fetch($eventId, $calendarId = 'primary')
```
bohraji: This function has not been used.

app/Observers/HR/ApplicationRoundObserver.php
```diff
@@ -4,20 +4,51 @@
...
+     * Listen to the ApplicationRound created event.
+     *
+     * @param  \App\Models\HR\ApplicationRound  $applicationRound
+     * @return void
+     */
+    public function created(ApplicationRound $applicationRound)
+    {
+        $applicationRound->load('application', 'scheduledPerson');
+        if ($applicationRound->application->status != config('constants.hr.status.on-hold.label')) {
+            $applicationRound->scheduledPerson->notify(new ApplicationRoundScheduled($applicationRound));
```
bohraji: Please make sure that this is queued.

rathorevaibhav: Agreed. I think I'll target queue for this but will target separately for other modules.

app/Services/CalendarEventService.php
rathorevaibhav: This method will be used when we need to see the event details. I have a use case for that. Thought to add it separately as it may involve more changes and this PR is already huge.

app/Services/CalendarEventService.php
rathorevaibhav: @bohraji I think I'll need some more time in creating custom exceptions and returning HTTP response based on them. Targeting it separately. #418

rathorevaibhav:
@bohraji queue is implemented for the email. I thought of implementing the queue for calendar API also, but we'll need the response from the API (event id and hangout link). Will have to research more on the performance there.

Issue #269, opened 2018-05-22
Error reporting enhancements


rathorevaibhav:
1. Trigger emails to system admin regarding all types of errors occurred in the Laravel.

deepakbarmola:
Closing the issue as the linked PR is approved and merged.

PR #419, opened 2018-06-12
269 error report

abhishek-pokhriyal:
#269 

Changes:

1. App\Exceptions\Handler.php
    - Added 5 exceptions in `$dontReport` (these will not be reported)
    - Modified `report` method such that it sends an email to the admin along with the exception object, time when exception occurred, and user who was logged in at that time 
2. Created `ErrorReport` mailable class along with a markdown
3. The receiver of email is specified in `.env` file `ERROR_MAIL` variable

To start receiving emails:
1. make a copy of `.env.example`
2. rename the copy as `.env`
3. put your email address in `ERROR_MAIL` variable in `.env`
4. also populate `MAIL_FROM_ADDRESS` variable in `.env`


app/Mail/ErrorReport.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Build the message.
+     *
+     * @return $this
+     */
+    public function build()
+    {
+        return $this->from(config('mail.from.address'))
+            ->subject('[ERROR REPORT] ' . $this->exception->getMessage())
+            ->markdown('mail.errorreport');
```
rathorevaibhav: lets rename to `mail.error-report`. It's more readable with the same meaning.

config/mail.php
```diff
@@ -120,4 +120,6 @@
         ],
     ],
 
+    'errormail' => env('ERROR_EMAIL'),
```
rathorevaibhav: let's use `error_email`

resources/views/mail/errorreport.blade.php
```diff
@@ -0,0 +1,17 @@
+@component('mail::message')
+Hi,<br>
+An error while handling a request. Here are the details:
+
+- User: {{ $user->name }} ({{ $user->email }})
+- Time: {{ $timeOfException }}
+- Error message: {{ $exception->getMessage() }}
+- File: {{ $exception->getFile() }}
```
rathorevaibhav: 👍

app/Exceptions/Handler.php
```diff
@@ -31,12 +37,21 @@ class Handler extends ExceptionHandler
...
-        parent::report($exception);
+        date_default_timezone_set('Asia/Kolkata');
+        $timeOfException = date('l, j-F-Y, h:i:s A');
+        foreach ($this->dontReport as $dontReport) {
+            if (($exception instanceof $dontReport)) {
+                return;
+            }
+        }
+        $user = Auth::user();
+        \Mail::to(config('mail.errormail'))->send(new ErrorReport($exception, $user, $timeOfException));
```
rathorevaibhav: Also, you can define the `to` parameter in the `ErrorReport`. No need to explicitly mention it here. Just like this:
https://github.com/ColoredCow/employee-portal/blob/master/app/Http/Controllers/HR/ApplicationRoundController.php#L55

resources/views/mail/errorreport.blade.php
```diff
@@ -0,0 +1,17 @@
+@component('mail::message')
+Hi,<br>
+An error while handling a request. Here are the details:
+
+- User: {{ $user->name }} ({{ $user->email }})
```
rathorevaibhav: Just curious how these values are set? How does the blade file knows what is in the `$user` variable since I didn't see you passing parameters to the blade file in ErrorReport.php

delhiite: @rathorevaibhav I'll quote what Jeffrey Way said in his tutorial on "Sending Mail":

> ...a neat thing about these mailable classes is that any public property you define on it will
> immediately be available to the view.

In the `ErrorReport` mailable class there are three public properties: `$user`, `$timeOfException`, and `$exception`. The values for these three variables automatically become available to the markdown view `errorreport.blade.php`.

So we don't need to pass it explicitly using `with` method.

abhishek-pokhriyal:
@rathorevaibhav made all changes you asked for. please review.

app/Exceptions/Handler.php
```diff
@@ -36,9 +40,16 @@ class Handler extends ExceptionHandler
...
     {
-        parent::report($exception);
+        $timeOfException = Carbon::now()->format(config('constants.display_datetime_format'));
+        foreach ($this->dontReport as $dontReport) {
+            if (($exception instanceof $dontReport)) {
+                return;
+            }
+        }
+        $user = Auth::user();
+        \Mail::send(new ErrorReport($exception, $user, $timeOfException));
```
bohraji: Please add a Try block here. Imagine if there is an exception in the email service itself.

bohraji: Also, please import the `mail` class above rather than accessing it in the global namespace.

app/Exceptions/Handler.php
```diff
@@ -36,9 +40,16 @@ class Handler extends ExceptionHandler
...
     public function report(Exception $exception)
     {
-        parent::report($exception);
+        $timeOfException = Carbon::now()->format(config('constants.display_datetime_format'));
+        foreach ($this->dontReport as $dontReport) {
+            if (($exception instanceof $dontReport)) {
+                return;
+            }
+        }
+        $user = Auth::user();
```
bohraji: Not all errors will be generated by a user. Certain errors, like the one's triggered by a cron or a background process will have no user. We need to have a default value also.

app/Mail/ErrorReport.php
```diff
@@ -0,0 +1,39 @@
...
+    use Queueable, SerializesModels;
+    public $user;
+    public $timeOfException;
+    public $exception;
+    /**
+     * Create a new message instance.
+     *
+     * @return void
+     */
+    public function __construct(Exception $exception, User $user, string $timeOfException)
```
bohraji: Assign an default value for the user here. Also, since it is `Auth::user`, we can directly assign it to the var below rather than sharing it from the initialising class.

.env.example
```diff
@@ -59,3 +59,5 @@ SES_REGION=
 
 MAIL_FROM_ADDRESS=
 MAIL_FROM_NAME=
+
+ERROR_EMAIL=
```
bohraji: Lets have all error related info start with `MAIL`. We can have `MAIL_ERROR_ADDRESS`. Also update config.

app/Mail/ErrorReport.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+    /**
+     * Build the message.
+     *
+     * @return $this
+     */
+    public function build()
+    {
+        return $this->from(config('mail.from.address'))
+            ->to(config('mail.error_mail'))
```
bohraji: The config should be `mail.error_address`

config/mail.php
```diff
@@ -14,7 +14,7 @@
     | Supported: "smtp", "sendmail", "mailgun", "mandrill", "ses",
     |            "sparkpost", "log", "array"
     |
-    */
+     */
```
bohraji: Not sure if this formatting update was required.

resources/views/mail/error-report.blade.php
```diff
@@ -0,0 +1,17 @@
+@component('mail::message')
+Hi,<br>
+An error while handling a request. Here are the details:
+
+- User: {{ $user->name }} ({{ $user->email }})
```
bohraji: User info will be optional

database/migrations/2018_03_13_132221_create_users_table.php
```diff
@@ -34,4 +34,4 @@ public function down()
     {
         Schema::dropIfExists('users');
     }
-}
+}
```
bohraji: No Specific change to this file. Please correct.

config/mail.php
abhishek-pokhriyal: @bohraji _phpfmt_ does this update when I save the file

app/Exceptions/Handler.php
abhishek-pokhriyal: @bohraji what should happen when an exception occurs in email service? in other words, what should be put in `catch` block

config/mail.php
bohraji: 👍

app/Exceptions/Handler.php
bohraji: We just log the error in the `log file` for now.

bohraji:
Checked this PR :

- [x] For duplication.
- [x] For performance. 
- [x] For coupling

bohraji:
@abhishek-pokhriyal Please also request review from @wevekamola as discussed.

app/Exceptions/Handler.php
```diff
@@ -36,9 +41,33 @@ class Handler extends ExceptionHandler
...
+
+        if (Auth::check()) {
+            $user = Auth::user();
+            $userDetails = [
+                "name" => $user->name,
+                "email" => $user->email,
+            ];
+        } else {
+            $userDetails = [
+                "name" => 'Guest',
```
bohraji: Please update it to `System Generated`

app/Exceptions/Handler.php
```diff
@@ -36,9 +41,33 @@ class Handler extends ExceptionHandler
...
+                "email" => '',
+            ];
+        }
 
+        try {
+            Mail::send(new ErrorReport($exception, $userDetails, $timeOfException));
+        } catch (Exception $e) {
+            parent::report($e);
```
bohraji: 👍

app/Exceptions/Handler.php
```diff
@@ -36,9 +41,33 @@ class Handler extends ExceptionHandler
...
-        parent::report($exception);
-    }
+        $timeOfException = Carbon::now()->format(config('constants.display_datetime_format'));
+        foreach ($this->dontReport as $dontReport) {
+            if ($exception instanceof $dontReport) {
+                return;
+            }
+        }
+
+        if (Auth::check()) {
```
bohraji: This piece is very specific to what information is shown when rendering the exception in the mail. Can we move this whole block to the `Mail` class _(controller)_, and not pass anything specific to the user here.

app/Mail/ErrorReport.php
```diff
@@ -0,0 +1,39 @@
...
+    public $exception;
+    /**
+     * Create a new message instance.
+     *
+     * @return void
+     */
+    public function __construct(Exception $exception, $user, string $timeOfException)
+    {
+        $this->exception = $exception;
+        $this->user = $user;
```
bohraji: This can be-
`$this->user = Auth::check() ? Auth::user()->name : 'System Generated'`

Thoughts?

abhishek-pokhriyal: @bohraji :thumbsup:

app/Mail/ErrorReport.php
```diff
@@ -0,0 +1,39 @@
...
+use App\User;
+use Exception;
+use Illuminate\Bus\Queueable;
+use Illuminate\Mail\Mailable;
+use Illuminate\Queue\SerializesModels;
+
+class ErrorReport extends Mailable
+{
+    use Queueable, SerializesModels;
+    public $user;
```
wevekamola: @abhishek-pokhriyal isn't it should be `$userDetails`?

app/Mail/ErrorReport.php
wevekamola: @abhishek-pokhriyal As we are checking if the user is logged in or not in the handler.php,  and passing the information with variable `$userDetails`. does this check is needed here also?

wevekamola:
@abhishek-pokhriyal Few doubts, Please clarify.

wevekamola:
@abhishek-pokhriyal changes look good to me. @bohraji can you have a final quick look at the PR.

Issue #358, opened 2018-06-04
Custom env variables should be called from config instead


rathorevaibhav:
![image](https://user-images.githubusercontent.com/12053186/40901056-15a69dc2-67ec-11e8-8bf0-c769d0a60ddd.png)

PR #420, opened 2018-06-12
now values is call via config instead of .env

mohitgusain:
#358

Following  hr mails are  now call from config instead of .env ,these mails are inside the
1. app/Mail/HR/Applicant/ApplicantCreateAutoResponder.php
2. app/Mail/HR/Applicant/RoundReviewed.php
3. app/Mail/HR/Application/JobChanged.php
4. app/Mail/HR/Application/RoundNotConducted.php


mohitgusain:
please review it @rathorevaibhav

Issue #410, opened 2018-06-11
New actions for application


rathorevaibhav:
The following actions need to be added to the application.

- [x] Send for approval
- [x] Approve

1. Send for approval is basically accepting the applicant but having a final review from a supervisor/super-admin/hr-manager. The user can send an application for approval and assign another user. The assignee should get a notification as pending (on approval) and should see the three options:
    1. Approve
    2. Schedule another round
    3. Reject

2. Based on the role (superadmin/hr-manager), the user can directly approve.

rathorevaibhav:
This is complete. Closing.

PR #444, opened 2018-06-14
(#410) "Send for approval" action in application round

rathorevaibhav:
#410

Changes include:

1. All actions for an application round can be selected from the drop-down (except reject). Reject will be added in the upcoming tasks.
2. Added two new actions `Send for approval` and `approve`. On send for approval, the user is asked to select a supervisor from whom the approval is requested. Approve action doesn't do anything yet.
3. A new section called "Pending" to show all applications marked as "sent-for-approval". Authorization checks haven't been implemented yet. Will follow in upcoming PRs.
4. The person requested for approval can either mark it as approved (once the approve functionality is done) or reject or schedule another round.
5. Cleanups in the `application/edit.php` to show the dropdown.


app/Models/HR/ApplicationRound.php
```diff
@@ -67,6 +67,11 @@ public function _update($attr)
                 $application->reject();
                 $applicant->applications->where('id', $attr['refer_to'])->first()->markInProgress();
                 break;
+
+            case 'send-for-approval':
+                $fillable['round_status'] = 'confirmed';
```
bohraji: Status should be pending if the application is `send-for-approval`. @rathorevaibhav is my understanding correct here?

resources/views/hr/application/edit.blade.php
```diff
@@ -174,43 +174,36 @@
...
                                     <button type="button" class="btn btn-primary ml-auto" data-toggle="modal" data-target="#round_{{ $applicationRound->id }}">Send mail</button>
                                 @endif
+                                </div>
                             </div>
                             @endif
                         </div>
                         <input type="hidden" name="action" value="updated">
+                        <input type="hidden" name="next_round" value="">
                         @includeWhen($applicationRound->round_status != config('constants.hr.status.confirmed.label'), 'hr.round-review-confirm-modal', ['applicationRound' => $applicationRound])
+                        @includeWhen($loop->last, 'hr.application.send-for-approval-modal')
```
bohraji: If this is a default model and we are not passing any value in it, why do we need to include it in the loop. Can we put it outside the loop?

app/Models/HR/ApplicationRound.php
rathorevaibhav: @bohraji we're maintaining two statuses. One is for application and another is for an application round. For application round, the status should be confirmed as the interviewer is confirming the round and sending for supervisor's approval. The application will be marked as `sent-for-approval`.

resources/views/hr/application/edit.blade.php
rathorevaibhav: This modal contains a submit button which should only request to update the last/latest application round. Hence, it needs to be inside the loop (and the form tags also).

rathorevaibhav: @bohraji let me know if that clears your understanding.

app/Models/HR/ApplicationRound.php
bohraji: 👍

resources/views/hr/application/edit.blade.php
bohraji: 👍

pankaj-ag:
@rathorevaibhav I have tested all the functionalities on UAT. I also made a UI change on the dashboard. It looks good to me for production.

PR #456, opened 2018-06-15
Build – 15/06/2018

rathorevaibhav:


pankaj-ag:
@rathorevaibhav I have tested all the functionalities on UAT. I also made a UI change on the dashboard. It looks good to me for production.

PR #495, opened 2018-06-20
dropdown will appear when application is rejected

abhishek-pokhriyal:
When an application is rejected in any round and a mail is sent to applicant, the drop-down menu for other actions should not disappear. 

screenshot 90
.


rathorevaibhav:
@abhishek-pokhriyal is this ready for review? I see myself as reviewer but the "ready for review" is missing.

abhishek-pokhriyal:
@rathorevaibhav I think this issue is better handled in this PR 499 https://github.com/ColoredCow/employee-portal/pull/499/

resources/views/hr/application/edit.blade.php
```diff
@@ -174,7 +174,7 @@
                                     </div>
                                 @endif
                             </div>
-                            @if (!$applicationRound->mail_sent)
+                            @if ($applicationRound->mail_sent && $loop->last && $application->isRejected())
```
rathorevaibhav: The footer won't load when the mail is not sent for the application round. This should not happen as all the new rounds have default null value in mail_sent field.

PR #502, opened 2018-06-21
temporary fix for testing

rathorevaibhav:
We need to change this implementation and use better techniques to mock API responses.

Tracking in #503


pankaj-ag:
@rathorevaibhav can we merge this fix in develop bracnh right now? It will help to move forward.

Issue #345, opened 2018-06-01
UI broken in home page


rathorevaibhav:
The UI shows up fine in Chrome but does not render well in Safari. Didn't check on Firefox.

![image](https://user-images.githubusercontent.com/12053186/40845172-81c4d124-65d3-11e8-8dca-210a65331b48.png)

delhiite:
The view depends on the **font size** _set by the user_ in his browser preferences. So the view changes according to font size even on the same browser.

This is the view in Chrome when _font size = small_
![small-font](https://user-images.githubusercontent.com/11808845/41141827-10567dba-6b11-11e8-9baa-635b1565c9b8.png)

And the view changes to this (the broken view) when _font size = medium_
![medium-font](https://user-images.githubusercontent.com/11808845/41141828-10c611ac-6b11-11e8-80a4-8a530c61f8aa.png)

delhiite:
@rathorevaibhav Please have a look and share your thoughts.

PR #465, opened 2018-06-16
345 UI broken in home page

vikasrnv:
#345 
 
1. some changes in the `font-size` and `grid-matrix`
2. now it look stable in every font size

PR #505, opened 2018-06-21
345 ui fix home page

vikasrnv:
#345 

  * Fix  `font-size` in mobile view
  * home UI fix


vikasrnv:
This is same kind of PR as  #465 but 
There were some instability in mobile view which in this PR.

rathorevaibhav:
@vikasrnv there are some conflicts in this. can you resolve before we proceed on the code review?

resources/views/home.blade.php
```diff
@@ -8,33 +8,33 @@
...
                 <br><h2 class="text-center">Finance</h2><br>
             </a>
         </div>
         @endcan
 
-        @if(auth()->user()->hasAnyPermission(['weeklydoses.view', 'library_books.view']))
-        <div class="col-md-3 card mx-5 mt-3 mb-5">
-            <a class="card-body no-transition" href="/knowledgecafe">
-                <br><h2 class="text-center">KnowledgeCafe</h2><br>
+         @if(auth()->user()->hasAnyPermission(['crm_talent.view', 'crm_client.view']))
```
rathorevaibhav: can we load KnowledgeCafe before CRM?

resources/views/hr/index.blade.php
```diff
@@ -3,23 +3,23 @@
...
-    <div class="d-flex justify-content-start row flex-wrap">
+    <div class="d-flex justify-content-center row flex-wrap">
+        <div class="col-md-3 card m-3">
         @can('hr_recruitment_applications.view')
-        <div class="col-md-3 card mx-5 mt-3 mb-5">
             <a class="card-body no-transition" href="{{ route('applications.job.index') }}">
                 <br><h2 class="text-center">Recruitment</h2><br>
             </a>
         </div>
+        <div class="col-md-3 card m-3">
```
rathorevaibhav: is this supposed to be inside `@can`?

resources/views/knowledgecafe/index.blade.php
```diff
@@ -3,16 +3,16 @@
...
-        <div class="col-md-3 card mx-5 mt-3 mb-5">
+        <div class="col-md-3 card m-3">
             <a class="card-body no-transition" href="/knowledgecafe/library/books">
                 <br><h2 class="text-center">Library</h2><br>
             </a>
         </div>
         @endcan
         @can('weeklydoses.view')
-        <div class="col-md-3 card mx-5 mt-3 mb-5">
+        <div class="col-md-3 card m-3">
```
rathorevaibhav: I think we need `mb-5`

resources/views/home.blade.php
vikasrnv: yes we can do this way also but I switch them just because the size of `KnowledgeCafe` is differ from other three.
also it does not effect much so I choose this way

resources/views/knowledgecafe/index.blade.php
vikasrnv: I will update them

resources/views/crm/index.blade.php
```diff
@@ -4,16 +4,20 @@
 <div class="container">
     <br>
     <div class="d-flex justify-content-start row flex-wrap">
-        <div class="col-md-3 card mx-5 my-3">
+ <div class="col-md-4">
```
bohraji: Please update formatting

PR #518, opened 2018-06-23
(#517) employee test coverage

rathorevaibhav:
#517 

Changes include:

1. Unit test case for Employee.
2. Feature test case for Employee. According to current functionality.
3. Updated routes for employees.
4. Updated functions.php for multiple entities created from a factory.


rathorevaibhav:
@pankaj-ag assigning this to you. Can you take a look?

database/factories/EmployeeFactory.php
```diff
@@ -0,0 +1,13 @@
...
+
+use App\Models\HR\Employee;
+use Faker\Generator as Faker;
+
+$factory->define(Employee::class, function (Faker $faker) {
+    return [
+        'name' => $faker->name,
+        'designation' => $faker->jobTitle,
+        'joined_on' => $faker->date,
+        'user_id' => null,
```
pankaj-ag: Just curious Employee will not have the user id?

routes/web.php
```diff
@@ -73,10 +73,12 @@
...
-            Route::get('/', 'EmployeeController@index')->name('employees');
-            Route::get('/{employee}', 'EmployeeController@show')->name('employees.show');
-        });
+        Route::resource('employees', 'Employees\EmployeeController')
+            ->only(['index', 'show'])
+            ->names([
+                'index' => 'employees',
+                'show' => 'employees.show',
+            ]);
         // A better version of employee-reports would be employees/reports. The
```
pankaj-ag: Just a suggestion Can we remove these comments and create an issue for this?

tests/Feature/HR/EmployeeTest.php
```diff
@@ -0,0 +1,36 @@
...
+    /** @test */
+    public function a_guest_cant_see_employee_list()
+    {
+        $this->get(route('employees'))->assertRedirect(route('login'));
+    }
+
+    /** @test */
+    public function an_authorized_user_can_see_employee_list()
+    {
+        $this->anAuthorizedUser();
```
pankaj-ag: Not sure this is an appropriate name. I got confused when I read it the first time. I thought it will return a user with some role.

tests/Unit/HR/EmployeeTest.php
```diff
@@ -0,0 +1,26 @@
...
+    public function setUp()
+    {
+        parent::setUp();
+        $this->employee = create(Employee::class);
+    }
+
+    /** @test */
+    public function it_is_created()
+    {
+        $this->assertTrue(isset($this->employee->id));
```
pankaj-ag: Not sure what we are checking in this test. But for now, it's fine. It will break in the setup method itself if something goes wrong.

rathorevaibhav: We're checking if the employee is created and has an ID assigned to it. Didn't get your question completely.

tests/Feature/HR/EmployeeTest.php
rathorevaibhav: This creates an authorized user. By default, it creates a super admin. @pankaj-ag Maybe we rename it to `$this->signInAsAuthorizedUser()`? Thoughts?

database/factories/EmployeeFactory.php
rathorevaibhav: Yeah, ex-employees won't have a user id. Only the current employees will have a user id. Also, since we're only using this factory in the Unit Test, we'll assign user_id when we go for Feature Tests.

routes/web.php
rathorevaibhav: Agreed. Created #541. Will remove these comments.

tests/Unit/HR/EmployeeTest.php
pankaj-ag: Yes, I can see that and it's fine. I was thinking that if there is an error while creating a new employee it will not execute any test in this file.

database/factories/EmployeeFactory.php
pankaj-ag: Alright.

tests/Feature/HR/EmployeeTest.php
pankaj-ag: Yea that will be good.

rathorevaibhav: @pankaj-ag I think let's target this separately since it'll involve changes in existing test cases which are using `$this->anAuthorizedUser()`

rathorevaibhav: Created #566

Issue #522, opened 2018-06-25
Database Design and Implementation for Organisations


nishanth-kd:
- [ ] Database schema design.
- [ ] Configuring and using two databases.

PR #524, opened 2018-06-25
Multi Tenant Database Configuration (#522)

nishanth-kd:
For #522 

Make a GET request to `/organizations` endpoint.

Left a few comments on the PR for review.

### DIFF
--- a/app/Http/Controllers/OrganizationController.php
+++ b/app/Http/Controllers/OrganizationController.php
@@ -40,7 +40,6 @@ class OrganizationController extends Controller
      */
     public function store(Request $request)
     {
-        
     }
 
     /**
--- a/app/Models/CreatesTenant.php
+++ b/app/Models/CreatesTenant.php
@@ -6,42 +6,48 @@ use App\Models\Configuration;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Artisan;
 
-trait CreatesTenant {
-
-    protected static function bootCreatesTenant() {
-        static::created(function($model){
+trait CreatesTenant
+{
+    protected static function bootCreatesTenant()
+    {
+        static::created(function ($model) {
             $model->createConnection();
             $model->migrateSchema();
             $model->seedSchema();
         });
     }
 
-    public function createConnection() {
+    public function createConnection()
+    {
         $connection = config('database.connections.default');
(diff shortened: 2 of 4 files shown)


app/Http/Controllers/OrganizationController.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Show the form for creating a new resource.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function create()
+    {
+        # for demo : create a DB emp_org_coloredcow and emp_portal_master and make a GET Request to /organization
+        Organization::create([
```
nishanth-kd: Demo Endpoint

app/Models/Configuration.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class Configuration extends Model
+{
+    protected $connection = 'master';
```
nishanth-kd: should probably abstract to a trait rather than that a Parent Model class ?

Trait will allow us to use it in User.php and other models where we don't want to extract our Model from another base model.

app/Models/CreatesTenant.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return "org_{$this->slug}";
+    }
+
+    public function getDatabaseNameAttribute()
+    {
+        return "emp_org_{$this->slug}";
+    }
+
+    # Do we want to create the organization database programatically? @rudresh @vaibhav @pankaj
```
nishanth-kd: Creating database from within the application ?

Permissions and best practices ? @rudresh @rathorevaibhav

app/Models/CreatesTenant.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public function migrateSchema()
+    {
+        Artisan::call('migrate', array('--database' => $this->connection_name, '--path' => 'database/migrations/'));
+    }
+
+    public function seedSchema()
+    {
+        Artisan::call('db:seed');
```
nishanth-kd: depends on what all we want to seed with. For example: we might not want to enable knowledge cafe for certain tenants, in which case running seeders for knowledge cafe wouldn't make sense.

Issue #208, opened 2018-05-14
Authorization checks on routes


rathorevaibhav:
Authorization checks are currently applied on controllers. To add another level of security, apply authorization on routes.

Satendra-SR:
@rathorevaibhav 
please let me know what type of authorization are you talking about.
In web.php file, authorization is already there. the following middleware is already in use.

`Route::middleware('auth')->group(function () { }`

Satendra-SR:
> @rathorevaibhav  **and I have discussed on this issue and what I understood is:**
* The **_authorization checks_** are present there in **controllers** but we have to insure the **security**. So what we deiced that there should be some kind of **authorization** there in  the **_routes_** aswell.

Satendra-SR:
What  I understand yet is, We have various **_Middlewares_** that are already defined in `kernel.php`.

![screenshot from 2018-06-19 17-04-23](https://user-images.githubusercontent.com/39792899/41595270-c2d3189c-73e3-11e8-978b-8a2d9ed5434b.png)
@rathorevaibhav 
@nishanth-kd 
So, should I use one of these **Middleware** (most probably : **_role_**) or should I define new **_Middlewares_** for **_authorization_** purpose.

rathorevaibhav:
@Satendra-SR 

Here are the links you can visit to know more about authorization checks in routes:
1. Spatie Laravel permission package (we're using this)
    https://github.com/spatie/laravel-permission#using-a-middleware
2. Laravel authorization documentation
    https://laravel.com/docs/5.6/authorization#via-middleware

Satendra-SR:
@rathorevaibhav 
#208
based on the [conversation](https://github.com/ColoredCow/employee-portal/pull/526#discussion_r203722450), should we continue to do this, or we have any other way to implement this?

PR #526, opened 2018-06-25
#208 added authorization checks on routes

Satendra-SR:
#208
>**What I did yet:**

- define policy for `getProjects`.
- define new permission for `getProjects` in `permissionTableSeeder.php`.
- add permissions to routes.
- replaced projects to projects.index in name route.


Satendra-SR:
#208 
- Add some new permissions to `routes` related to **volunteers**.
- Make changes to `jobPolicy`, define policies related to **Volunteer**.
- Add permissions to `PermissionsTableSeeder` related to Volunteer.

@nishanth-kd 
please review it.

rathorevaibhav:
@Satendra-SR can you check why the TravisCI build is failing? Please make sure all the checks are passing before assigning for review.

Satendra-SR:
OK, I'll make sure this won't happen again. I will check it.

Satendra-SR:
@rathorevaibhav 
@nishanth-kd 
resolve the tavisCI fail. please review it.

routes/web.php
```diff
@@ -83,27 +86,34 @@
         // route below can be merged with employees grouped route above.
         Route::get('employee-reports', 'Employees\ReportsController@index')->name('employees.reports');
 
-        Route::resource('applicants', 'ApplicantController')->only(['index', 'edit']);
+        Route::resource('applicants', 'ApplicantController')->only(['index', 'edit'])->middleware('can:hr_applicants.view');
```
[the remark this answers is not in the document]
Satendra-SR: it makes sense.
only view policy was applied. but now, it checks for both view and edit.

routes/web.php
```diff
@@ -56,23 +56,29 @@
                     'store' => 'volunteer.opportunities.store',
                     'update' => 'volunteer.opportunities.update',
                     'edit' => 'volunteer.opportunities.edit',
-                ]);
+                ])->middleware('can:hr_volunteers_jobs.view, hr_volunteers_jobs.update, hr_volunteers_jobs.create, hr_volunteers_jobs.delete');
```
nishanth-kd: you need to have .view permission for .edit ? I think you should replace this with a policy. This is too clunky

routes/web.php
```diff
@@ -116,7 +130,7 @@
                     'store' => 'books.store',
                     'destroy' => 'books.delete',
                     'update' => 'books.update',
-                ]);
+                ])->middleware('can:library_books.view,library_books.create, library_books.delete, library_books.update');
```
nishanth-kd: Policy for all these Resource groups.

Satendra-SR: you are right KD. I think I have to do more research on this topic(how to implement spatie laravel permissions to the resource group), Please give me some time for this particular issue.

routes/web.php
nishanth-kd: I think we should not have the same checks duplicated in Routes as well as controllers. Because it violates DRY. 

If we need to be extra cautious, I would recommend adding a test for it.

nishanth-kd: Either we should remove from controller or from here. @Satendra-SR

PR #532, opened 2018-06-29
ACL cleanup for existing modules and users

rathorevaibhav:
Changes include:

1. Updated permissions for all modules.
2. There are some file changes due to hotfixes did recently in master.


resources/views/layouts/app.blade.php
```diff
@@ -63,6 +71,7 @@
                             </li>
                         </li>
                         @endif
+                        @if(auth()->user()->can('crm_talent.view') || auth()->user()->can('crm_client.view'))
```
pankaj-ag: Just curious, Can we combined both permissions. 
```
$user->hasAnyPermission(['crm_talent.view', 'crm_client.view']);
```

Or assign them to role itself.

rathorevaibhav:
@pankaj-ag done with the review feedback. Ready for review again.

Issue #520, opened 2018-06-23
Organization onboarding


rathorevaibhav:
We need a seamless onboarding for organizations.

The onboarding process should ask for all the GSuite credentials required for application setup.

PR #535, opened 2018-07-02
Welcome page

rathorevaibhav:
#520 

Changes include:

1. New welcome page at `/` route.


app/Http/Controllers/WelcomeController.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace App\Http\Controllers;
+
+class WelcomeController extends Controller
+{
+    public function index()
+    {
+        if (auth()->check()) {
+            return redirect('home');
```
pankaj-ag: Just Curious can't we use guest or auth middleware?

resources/views/welcome.blade.php
```diff
@@ -1 +1,130 @@
...
+
+        <title>{{ config('app.name', 'Laravel') }}</title>
+        <link rel="icon" href="{{ asset('favicon.png') }}" sizes="32x32" />
+
+        <!-- Fonts -->
+        <link href="https://fonts.googleapis.com/css?family=Raleway:100,600" rel="stylesheet" type="text/css">
+        <link href="{{ mix('css/app.css') }}" rel="stylesheet">
+
+        <!-- Styles -->
+        <style>
```
pankaj-ag: Just a question why we are not using css/app.css or welcome.css for this? I think that will help us to use SAAS.

app/Http/Controllers/WelcomeController.php
rathorevaibhav: We need to redirect if the user is already logged-in the welcome page shouldn't appear (for now). He should be directed to the homepage. Not very sure how I can define the default route for logged-in users using a middleware.

rathorevaibhav:
@pankaj-ag done with the requested changes. Ready for review again.

app/Http/Controllers/WelcomeController.php
pankaj-ag: I think you can. Found this in the documentation for guest middleware.
 
image

PR #537, opened 2018-07-03
Volunteer manager

rathorevaibhav:
Changes include:

1. New role `volunteer-manager`
2. Changes in JobController and JobPolicy based on routes
3. Not checking edit and update method for authorization in JobController. Was giving an error. Will fix in another PR.

### DIFF
--- a/app/Policies/HR/JobPolicy.php
+++ b/app/Policies/HR/JobPolicy.php
@@ -75,7 +75,8 @@ class JobPolicy
      * @param  \App\User  $user
      * @return mixed
      */
-    function list(User $user) {
+    public function list(User $user)
+    {
         if (request()->is('hr/volunteers*')) {
             return $user->hasPermissionTo('hr_volunteers_jobs.view');
         }
--- a/app/Policies/HR/JobPolicy.php
+++ b/app/Policies/HR/JobPolicy.php
@@ -2,8 +2,8 @@
 
 namespace App\Policies\HR;
 
-use App\User;
 use App\Models\HR\Job;
+use App\User;
 use Illuminate\Auth\Access\HandlesAuthorization;
 
 class JobPolicy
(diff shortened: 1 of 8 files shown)


app/Policies/HR/JobPolicy.php
```diff
@@ -19,7 +19,10 @@ class JobPolicy
      */
     public function view(User $user, Job $job)
     {
-        return $user->hasPermissionTo('hr_jobs.view');
+        if ($job->type == 'volunteer') {
```
pankaj-ag: Can we create a new function Isvolunteer().

app/Policies/HR/JobPolicy.php
```diff
@@ -42,7 +48,10 @@ public function create(User $user)
      */
     public function update(User $user, Job $job)
     {
-        return $user->hasPermissionTo('hr_jobs.update');
+        if ($job->type == 'volunteer') {
```
pankaj-ag: Same as above. Make it a function.

app/Policies/HR/JobPolicy.php
```diff
@@ -65,6 +77,9 @@ public function delete(User $user, Job $job)
      */
     public function list(User $user)
     {
-        return $user->hasPermissionTo('hr_jobs.view');
+        if (request()->is('hr/volunteers*')) {
```
pankaj-ag: This is fine but I think you can create a helper function also.

resources/views/hr/job/index.blade.php
```diff
@@ -16,7 +16,11 @@
         @foreach ($jobs as $job)
         <tr>
             <td>
-                <a href="/hr/jobs/{{ $job->id }}/edit">{{ $job->title }}</a>
+                @if ($type == 'volunteer')
```
pankaj-ag: Function?

rathorevaibhav:
@pankaj-ag thanks. Since this is a priority item, I'm merging it. Targeting feedback in the next one.

PR #548, opened 2018-07-05
185 create invoice billing stage

mohitgusain:
1. Now invoices can create from project billing stages.
2. A new migration created because during invoice create from billing stage it shows an error of `values cant'be null` for some data fields in the database like currency_ transaction_charge and for two more fields in the database.


rathorevaibhav:
@mohitgusain before assigning please make sure the all the status checks are passing.

rathorevaibhav:
@pankaj-ag These column definitions already exist in a previous migration and I think the objective of this migration was to make these fields nullable. Considering that, I don't think the `down` correctly undo the changes done by the `up` method. I think the `down` method needs to be changed a bit.
image

Though its a very small thing and won't impact anything at all but IMO we should be clear on the understanding of what `down` method is supposed to do. Thoughts?

pankaj-ag:
Yes, @rathorevaibhav I agree with you.  I am reassigning it to @mohitgusain to make this change.

pankaj-ag:
@mohitgusain can you go through the @rathorevaibhav's comment and implement the change. Let me know if you have any doubts about this.

mohitgusain:
@pankaj-ag Yes I want to know and understand how to implement and  what type of changes should I have to do

mohitgusain:
@pankaj-ag as you cleared my doubts about the down function in the migration file. So for make value of nullable false in down function of migration should code be like this? 

 `$table->string('currency_transaction_charge')->nullable(false)->change();`

am I right? Please inform me.

pankaj-ag:
Yes, @mohitgusain something like that. Please test the functionality before assigning it to @rathorevaibhav.

mohitgusain:
@pankaj-ag this is the [link](https://stackoverflow.com/questions/24419999/laravel-migration-change-and-make-column-nullable) from  where I get this imformation

pankaj-ag:
Have a look in this one. 
https://github.com/laravel/framework/issues/11380

rathorevaibhav:
Working as expected. Merging.

Issue #459, opened 2018-06-15
Invoice fields that are related to "paid" status should be required when status is paid


rathorevaibhav:
The following fields should be required if invoice status is "paid".

- [x] Paid on
- [x] Received amount
- [x] Payment type
- [x] If payment type is cheque, check status and cheque date should be required

Satendra-SR:
when I go through the issue, I saw that the paid information is visible if we select unpaid status.
When we create an invoice from **/finance/invoices/create** the paid section opens only when the status is paid. We can see below

**when the `status` is `unpaid`**
![unpaid2](https://user-images.githubusercontent.com/39792899/42432310-8cef1efe-8367-11e8-8985-e910045e77d4.png)

**when `status` is `paid`**
![unpaid3](https://user-images.githubusercontent.com/39792899/42432315-919d6816-8367-11e8-9918-e820644da82c.png)

Similarly, if we create an invoice from **/projects/2/edit**, the paid information is visible if the `unpaid` `status` is selected.
![unpaid1](https://user-images.githubusercontent.com/39792899/42432248-3c3ffe2e-8367-11e8-992a-702de0e9cf8e.png)

#459
@rathorevaibhav 
My query is, should we hide paid section until the paid section will not be selected?

rathorevaibhav:
@Satendra-SR To your query, yes. The fields in the payment details section should be hidden until the status is "unpaid". When the status changes to "paid", the fields should show up.

Furthermore, this issue mentions that the fields in the  payment details should be required if the status is "paid". The form should not gets submitted if form validation fails and should report validation errors. This needs to be implemented at `invoice/create` and `invoice/edit` screens and also in the `create invoice modal` in the project stage screen.

Let me know if this resolves your query.

Satendra-SR:
@rathorevaibhav
Thank you for the reply.  actually, I'm clear about my issue, but I saw something inappropriate, so I decided you should know that( if not already known).

rathorevaibhav:
Thanks. That show/hide thing was meant to be fixed in other PR but got missed. Glad you noticed.

The show/hide thing is already implemented in the `invoice create` screen. You can reuse the code from there to put it in the project stage create invoice model.

Satendra-SR:
@rathorevaibhav 
I just got to know that this is already an issue and assigned to @mohitgusain .
thank you.

rathorevaibhav:
Complete. Further enhancements will be taken care in #611

PR #550, opened 2018-07-09
459 paid status

Satendra-SR:
#459
### What I did yet:
- Updated front-end validation in `invoice>create`.
- Updated front-end validation in `invoice>edit`.

#### here is a screenshot for the above description.
invoicesedit


Satendra-SR:
#459
### updates:
- updated required fields in projects/{id}/edit while creating invoices.

#### here is a screenshot for the above description.
projectinvoice

Satendra-SR:
#459 
### What I did:
- updated backend validation as required

### working on
- front end validation is not working fine. Trying to resolve them.

Satendra-SR:
- all changes are applied in both invoice > (create/update) and project/id/edit to create invoices.

- [ ] Paid on
- [ ] Received amount
- [ ] Payment type
- [ ] If the payment type is cheque, check status and cheque date should be required

### what I did to apply it:
- all the mentioned field are now required.
- uses **v-if** instead of **v-show** to eliminate the `<div> ` that are not selected. for example if **unpaid** is selected all the field related to **paid** status is now eliminated. these fields are not going with input array.
that means the fields are required only if the status is paid.
-` isset()` is used in the controller instead of the **collection**. as it was used previously. it can be done with **Collections**. @pankaj-ag  said that he will create an **issue** the above

resources/assets/js/components/ProjectStageComponent.vue
```diff
@@ -176,15 +176,15 @@
                                 <br>
                                 <h4 class="my-4"><u>Payment Details</u></h4>
                                 <div class="form-row">
-                                    <div class="form-group" :class="[clientCountryGstApplicable ? 'col-md-4' : 'col-md-5']">
-                                        <label for="paid_on">Paid on</label>
-                                        <input type="date" class="form-control" name="paid_on" id="paid_on" placeholder="dd/mm/yyyy">
+                                    <div class="form-group" :class="[clientCountryGstApplicable ? 'col-md-4' : 'col-md-5']" v-if="status == 'paid'">
```
rathorevaibhav: instead of checking `status == 'paid'` at multiple places, can you create a variable `isPaid` and check that? `isPaid` will change to true/false based on status.

resources/assets/js/components/ProjectStageComponent.vue
```diff
@@ -176,15 +176,15 @@
...
+                                        <label for="paid_on" class="field-required">Paid on</label>
+                                        <input type="date" class="form-control" required="required" name="paid_on" id="paid_on" placeholder="dd/mm/yyyy">
                                     </div>
-                                    <div class="form-group" :class="[clientCountryGstApplicable ? 'col-md-4' : 'offset-md-1 col-md-5']">
-                                        <label for="paid_amount">Received amount</label>
+                                    <div class="form-group" :class="[clientCountryGstApplicable ? 'col-md-4' : 'offset-md-1 col-md-5']" v-if="status == 'paid'">
```
rathorevaibhav: use `isPaid`

resources/assets/js/components/ProjectStageComponent.vue
```diff
@@ -205,32 +205,32 @@
                                     </div>
                                 </div>
                                 <br>
-                                <div class="form-row">
+                                <div class="form-row" v-if="status == 'paid'">
```
rathorevaibhav: Use `isPaid` here and also at other places. I'm skipping this comment in other places.

resources/assets/js/components/ProjectStageComponent.vue
```diff
@@ -264,6 +264,7 @@
                 clientCountryGstApplicable: this.client.country == 'india' ? true : false,
                 selectedPaymentType: '',
                 selectedChequeStatus: '',
+                status:'',
```
rathorevaibhav: By default, the status should be unpaid for a new invoice. For an existing invoice, the status should be the status of that invoice. Please update.

Satendra-SR:
Done with the feedback.
- isPaid is used seperatly for checking if the status is paid or not.
- initially, the status is unpaid now.
@rathorevaibhav 
please review it.

nishanth-kd:
@Satendra-SR I believe this was previously reviewed by @rathorevaibhav, skipping this one.

nishanth-kd:
What is the review status of this one @Satendra-SR @rathorevaibhav ?

Satendra-SR:
Hi @nishanth-kd, @rathorevaibhav has suggested that you should review it from the beginning. so can you please review it?

app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -80,19 +80,20 @@ public function store(InvoiceRequest $request)
             'sent_amount' => $validated['sent_amount'],
             'currency_sent_amount' => $validated['currency_sent_amount'],
             'gst' => isset($validated['gst']) ? $validated['gst'] : null,
-            'paid_on' => $validated['paid_on'] ? DateHelper::formatDateToSave($validated['paid_on']) : null,
-            'paid_amount' => $validated['paid_amount'],
-            'payment_type' => $validated['payment_type'],
+            'paid_on' => isset($validated['paid_on']) ? DateHelper::formatDateToSave($validated['paid_on']) : null,
+            'paid_amount' => isset($validated['paid_amount']),
```
nishanth-kd: So what are the possible values of paid_amount ? false and true? where do you store the actual paid amount in ?

nishanth-kd: it feels like this needs some refactoring because if it's boolean then shouldn't it be `is_paid_amount`. What is the significance of this variable otherwise.?

nishanth-kd: If it's not boolean then `isset($validated['paid_amount'])` isn't right. If this field is not applicable, you should store a null whenever it is not, `false` wouldn't make sense.

Satendra-SR: oh! Sorry @nishanth-kd 
I got it wrong. 
It was just checking if the value is present or not. I forget to validate it further.

rathorevaibhav:
@Satendra-SR please have a look at the following test feedback:

I was able to submit the form without entering the paid amount field. Though the field has the red asterisk, the front-end validation failed.
image

Also, when I tried to change an invoice status from "paid" to "unpaid", it gave the following error.
image

Satendra-SR:
I think that was the silly mistake, how can I forget about the impact of my work on other functionalities.
now it works fine. sorry for the inconvenience to you @rathorevaibhav.

rathorevaibhav:
@Satendra-SR I am able to update invoice without entering the paid amount. It's not asking in the front-end validation but fails in back-end when paid amount is missing.

Just this one feedback. The front-end validation should prompt about the empty paid field. Otherwise, everything else is looking good.

Let's target it separately.

Issue #463, opened 2018-06-15
Status modal doesn't disappear by clicking on x button.  


pankaj-ag:
When we enable the book suggestion notification the status modal doesn't disappear without page reload. 

![image](https://user-images.githubusercontent.com/10670690/41469200-017925a4-70ca-11e8-87cf-4d5d9254ba90.png)

rathorevaibhav:
@pankaj-ag very weird it's not working. Can you put more details and pages where this is not working?

pankaj-ag:
@rathorevaibhav currently this is the only page I know having this issue.

rathorevaibhav:
This is complete. Closing.

PR #554, opened 2018-07-10
463 status modal doesn't disappear on clicking x button

abhishek-pokhriyal:
#463 

Reason for  problem: 
@rathorevaibhav pointed out that jQuery code for closing the modal was defined before (above) the VueJs code that prepares the book `index.blade.php` page. So jQuery code for closing the modal was being executed before the modal was even rendered.

Our solution:
jQuery for closing the modal is executed when the document is ready that is the jQuery code is now put inside `$document.ready()`

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -223,6 +223,10 @@ $(document).ready(() => {
         }
     }
     $('[data-toggle="tooltip"]').tooltip();
+    $('.status-close').on('click', function(){
+        let wrapper = $(this).closest('.alert');
+        wrapper.fadeOut(500);
+    });
 });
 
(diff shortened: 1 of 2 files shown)


abhishek-pokhriyal:
@wevekamola and @bohraji please review.

abhishek-pokhriyal:
@wevekamola and @bohraji please review.

abhishek-pokhriyal:
@wevekamola and @bohraji please review.

wevekamola:
@abhishek-pokhriyal The current implementation looks like a temporary fix as the same functnality is working on the home page. I think the issue needs a bit more debugging.

rathorevaibhav:
Works as expected. Merging.

Issue #482, opened 2018-06-18
Update books page 


pankaj-ag:
![image](https://user-images.githubusercontent.com/10670690/41523733-adcd0606-72f8-11e8-8f5b-bf474e1463b3.png)

Fix the alignment in all the devices.

abhishek-pokhriyal:
After discussing with @pankaj-ag what I understood about this issue.
The layout of the page that displays all the books breaks when the device is width is between `768px` to `991px`, that is, tablet screen. 

The layout renders fine for all the other screen sizes.

rathorevaibhav:
This is complete. Closing.

PR #558, opened 2018-07-11
book card width increased for tablet screen (> 767px)

abhishek-pokhriyal:
#482 

The book page for tablet screen appears broken. Here's the screenshot:
screenshot 94

Problems:
1. The text (book title and author) is coming out of the book card.
1. There is some space missing at the end of each row of book cards
1. The size of book card is little small.

Did the following the changes:
1. used `word-break` to automatically break long words in the next line
1. increased the card-width to `49%` so that only two cards fit in the row and no space is left on the right

Here's how the page look after these changes:
screenshot 93

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -380,7 +380,8 @@ input:focus+.slider {
 @media (min-width: 768px) and (max-width: 1024px) {
     #books_listing {
         .book_card {
-            width: 32%;
+            width: 49%;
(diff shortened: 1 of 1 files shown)


rathorevaibhav:
Seems good. However, the UI breaks if screen size is less than 320px. I think we're not following the mobile first approach. Need to target separately #590 

image

rathorevaibhav:
Merging for now.

Issue #544, opened 2018-07-05
GSuite requires full name (first and last) for an applicant to be onboarded


rathorevaibhav:
When creating approving an applicant and creating his/her Google account, the process fails without an error if the name doesn't contain a first and last name.
This shouldn't happen. In case there's just the first name, the supervisor should be able to create the Google account for the applicant.

PR #562, opened 2018-07-11
g-suite require full name

vikasrnv:
#544 

- g-suite onboard problem solved
- now every applicant needs to fill `last_name` also .
 
- Client applications are requested to update these two field in employee_application form.     
     - remove
         -   `name`    
    - add 
        -  `first_name` 
        -  `last_name`

### DIFF
--- a/app/Http/Controllers/HR/ApplicantController.php
+++ b/app/Http/Controllers/HR/ApplicantController.php
@@ -45,6 +45,7 @@ class ApplicantController extends Controller
     public function store(ApplicantRequest $request)
     {
         $validated = $request->validated();
+        $validated['name'] = $validated['first_name'] . ' ' . $validated['last_name']
         return Applicant::_create($validated);
     }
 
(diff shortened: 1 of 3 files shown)


vikasrnv:
my test `Travis CI` failed as the test case are written in `employee-portal/tests/Feature/HR/ApplicationTest.php` based on previous application form data. so they are conflicting with the changes which I have made on `employee-portal/app/Models/HR/Applicant` file. 

I think there is a need for updating the test cases.
so what should be the ideal changes that I need to make?
@bohraji @rathorevaibhav

vikasrnv:
now I have put the code in `app/Http/Controllers/HR/ApplicantController.php` insteat of  `employee-portal/app/Models/HR/Applicant` due to which test is pass.

vikasrnv:
@bohraji 
So we have also considered this two problem before starting the issue.
 - @rathorevaibhav is already aware of the effect of this in other website and I also mention in it in `pr 
   description`.
- We already having functionality `split_name` which split the name into `first_name` and `last_name` so
  there will be no error because of this when we onbord user.
- The problem was when an applicant submits their application by entering only `fisrt_name` then 
  `split_name` send last_name as `null` due to which required field last_name of `G-suite` give the error.
  so as a solution we make application field also required.

rathorevaibhav:
We need to fix the forms in our WP site to accommodate these changes. Putting this on hold until that is fixed.

Will update once WP changes are done.

PR #570, opened 2018-07-13
460 billing percentage

Satendra-SR:
### What I did:

- create **3** separate `<div>` for **billing %**
   - First one will be **visible** when we create a new invoice
   - The second one will be **visible** when we edit the invoice.
   - The third one is for new invoice while we create an invoice using add projects in edit invoice page.
- use v-if to toggle the above three div

### screenshot for the above description

invoiiiiice


app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -80,14 +80,14 @@ public function store(InvoiceRequest $request)
             'sent_amount' => $validated['sent_amount'],
             'currency_sent_amount' => $validated['currency_sent_amount'],
             'gst' => isset($validated['gst']) ? $validated['gst'] : null,
-            'paid_on' => $validated['paid_on'] ? DateHelper::formatDateToSave($validated['paid_on']) : null,
-            'paid_amount' => $validated['paid_amount'],
-            'payment_type' => $validated['payment_type'],
+            'paid_on' => isset($validated['paid_on']) ? DateHelper::formatDateToSave($validated['paid_on']) : null,
+            'paid_amount' => isset($validated['paid_amount']),
```
nishanth-kd: we're setting the value to boolean @Satendra-SR ?

app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -80,14 +80,14 @@ public function store(InvoiceRequest $request)
...
+            'payment_type' => isset($validated['payment_type']),
```
nishanth-kd: boolean ?

resources/assets/js/components/InvoiceProjectComponent.vue
```diff
@@ -5,6 +5,7 @@
             :index="index"
             :item="item"
             :client="client"
+            :editMode="editMode"
```
nishanth-kd: do we need to specify attributes here if we're basically not assigning any value to them.

nishanth-kd: editMode should be a computed property, please read up on computed properties.

resources/views/finance/invoice/create.blade.php
```diff
@@ -33,7 +33,8 @@
                 <br>
                 <invoice-project-component
                 :billings="{{ json_encode([]) }}"
-                :client="activeClient">
+                :client="activeClient"
+                :invoice="[]">
```
nishanth-kd: if invoice is empty you can just return an empty array in the `data() { invoice: [] }` section

resources/views/finance/invoice/edit.blade.php
```diff
@@ -37,7 +37,8 @@
                 <br>
                 <invoice-project-component
                 :billings="{{ json_encode($invoice_billings) }}"
-                :client="{{ json_encode($invoice_client) }}">
+                :client="{{ json_encode($invoice_client) }}"
+                :invoice="{{ json_encode($invoice_client) }}">
```
nishanth-kd: this makes sense.

app/Http/Controllers/Finance/InvoiceController.php
Satendra-SR: actually, I was working on this issue and on #459 and I think there was some confusion.

resources/assets/js/components/InvoiceProjectComponent.vue
Satendra-SR: the `editMode` compare the invoice coming from the `edit invoice` and `create invoice` view. it is not being edited at the time of execution. Is there any need to put it in computed property? please tell me if I'm wrong.

resources/views/finance/invoice/create.blade.php
Satendra-SR: actually, It was for checking if we try to create or to edit an invoice. if the `invoice` is an empty array, it will set `editMode` `false`, and if it is not empty then the `editMode` will be `true`.

app/Http/Controllers/Finance/InvoiceController.php
Satendra-SR: I will correct it.

resources/views/finance/invoice/create.blade.php
Satendra-SR: I think it should be like `:invoice="{{ json_encode([]) }}`

nishanth-kd: umm, no, if it's empty always at create why do you need to pass it ?

resources/assets/js/components/InvoiceProjectComponent.vue
nishanth-kd: Yes, have you read on computed properties ?

I think you need to some clarity on when to use these attributes and when to use computed properties and when to define data variables in the component.

nishanth-kd: please read up and do your research.

Satendra-SR: sure, I will do it.

resources/views/finance/invoice/create.blade.php
Satendra-SR: have a look
`:billings="{{ json_encode([]) }}` it is also used in above code.
#### according to my understanding
`:billings="{{ json_encode([]) }}` and `:invoice="{{ json_encode([]) }}` is used for comparision perpose.
- `:billings` and `:invoice` are send from both `edit` and `create` view.
- it will check in `InvoiceProjectComponent.vue` if it is **empty** or not and changes **functionality** accordingly.
@nishanth-kd 
please let me know if I 'm able to make you understand.

Satendra-SR:
@nishanth-kd 
I put this one on hold and I have mentioned it in the related issue.
[here is the link](https://github.com/ColoredCow/employee-portal/issues/460#issuecomment-405946569).
I show you have done code refactoring in One of your Project. I will need your help in this particular issue.

rathorevaibhav:
1. Got a case where something is not right.

    - I was able to add "Test stage" in "Test project" and select the billing 2% in edit invoice screen. The "Sample stage" billing was already attached to the invoice. (see screenshot below)
image


    - Now when I remove all the invoices (without updating) and add project again, select "Test stage" in test project, now no billing comes up.
image

2. When updating an invoice with two billings, the second one doesn't show the billing in the select field.
image

@Satendra-SR please have a look at this case. We can discuss more details on this.

Issue #574, opened 2018-07-14
New role: organization-admin


rathorevaibhav:
This'll be the person who is onboarding an organization.

- Will have access to the ACL page
- Will have access to settings page
- Will have access to organization details page

PR #582, opened 2018-07-15
organization onboarding

rathorevaibhav:
Covers

1. New role organization admin #574


rathorevaibhav:
Will work on this later. Not needed right now.

PR #591, opened 2018-07-18
458 amount should not show zero

Satendra-SR:
#458 
#### I just do two things:
- changed project stage amount(cost) to null. previously, it was 0.
- changed billing % to null. previously, it was 0.

### DIFF
--- a/resources/assets/js/components/ProjectStageBillingComponent.vue
+++ b/resources/assets/js/components/ProjectStageBillingComponent.vue
@@ -37,13 +37,13 @@
         props: ['billing', 'index', 'stageCostWithGst', 'gstAmount', 'stageCostWithoutGst', 'currency', 'clientCountryGstApplicable'],
         computed: {
             billingCostWithoutGst: function() {
-                return parseFloat((this.billing.percentage/100)*this.stageCostWithoutGst).toFixed(2);
+                return this.stageCostWithoutGst  !== 'NaN' ? parseFloat((this.billing.percentage/100)*this.stageCostWithoutGst).toFixed(2) : parseFloat(0).toFixed(2);
             },
             billingGstAmount: function() {
                 return parseFloat((this.billing.percentage/100)*this.gstAmount).toFixed(2);
             },
             billingCostWithGst: function() {
-                return parseFloat((this.billing.percentage/100)*this.stageCostWithGst).toFixed(2);
+                return this.stageCostWithGst !=='NaN' ? parseFloat((this.billing.percentage/100)*this.stageCostWithGst).toFixed(2) : parseFloat(0).toFixed(2);
             }
         },
         methods: {
--- a/resources/assets/js/components/ProjectStageBillingComponent.vue
+++ b/resources/assets/js/components/ProjectStageBillingComponent.vue
@@ -37,13 +37,13 @@
         props: ['billing', 'index', 'stageCostWithGst', 'gstAmount', 'stageCostWithoutGst', 'currency', 'clientCountryGstApplicable'],
         computed: {
             billingCostWithoutGst: function() {
-                return this.stageCostWithoutGst  == 0 ?parseFloat(0).toFixed(2) : parseFloat((this.billing.percentage/100)*this.stageCostWithoutGst).toFixed(2);
+                return this.stageCostWithoutGst  == 0 ? parseFloat(0).toFixed(2) : parseFloat((this.billing.percentage/100)*this.stageCostWithoutGst).toFixed(2);
             },
             billingGstAmount: function() {
                 return parseFloat((this.billing.percentage/100)*this.gstAmount).toFixed(2);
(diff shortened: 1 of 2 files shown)


resources/assets/js/components/ProjectStageComponent.vue
```diff
@@ -302,7 +302,7 @@
         methods: {
             addBilling() {
                 this.stageBillings.push({
-                    'percentage': 0,
+                    'percentage': '',
```
nishanth-kd: does it add null or an empty string into the percentage field in the database? 
Also what is the datatype of the field in the DB.

Satendra-SR: here is the migration for this particular field
`$table->float('percentage', 5, 2);` it doesn't accept null value.
I tried to add null value in billing percentage but no entry was there in the database. 
but I thing frontend validation should be there. I will work on that also.

Satendra-SR: I talk with @rathorevaibhav and he suggested to me that if we need these types of validation, I should create a new issue and should be assigned to him so that he can decide if this functionality should be there in employee portal or not.

rathorevaibhav: Thought to pitch in and clarify the confusion.
On @nishanth-kd's comment: A percentage is mandatory for a billing. In the ProjectStageController, we don't save empty billings. The data type is a float.
https://github.com/ColoredCow/employee-portal/blob/master/app/Http/Controllers/ProjectStageController.php#L54

On @Satendra-SR's comment: Yes it'll better to have front-end validation as the back-end validation is in place already. Can you create a new issue and I'll plan when and how it should be executed?

Satendra-SR: #594 is the new issue related to it.

nishanth-kd: Yup just wanted to make sure we're not storing `""` where there should be null

rathorevaibhav:
@Satendra-SR please have a look at the following test feedbacks:

1. When there's no stage cost and billing percentage, the cost column says `$NaN` (not a number). It's should be because of system is parsing empty values and trying to calculate the cost from percentage (which is not a number). In this case, it should show `$0`.
image

I think apart from this, it is working as expected.

Satendra-SR:
@rathorevaibhav 
done with your feedback.
here is the screenshot:

nozero

please test it.

Satendra-SR:
done with feedbacks.

resources/assets/js/components/ProjectStageBillingComponent.vue
```diff
@@ -37,13 +37,13 @@
         props: ['billing', 'index', 'stageCostWithGst', 'gstAmount', 'stageCostWithoutGst', 'currency', 'clientCountryGstApplicable'],
         computed: {
             billingCostWithoutGst: function() {
-                return parseFloat((this.billing.percentage/100)*this.stageCostWithoutGst).toFixed(2);
+                return this.stageCostWithoutGst  !== 'NaN' ? parseFloat((this.billing.percentage/100)*this.stageCostWithoutGst).toFixed(2) : parseFloat(0).toFixed(2);
```
bohraji: @Satendra-SR Here, we are checking if the vue is having trouble parsing _(which is resulting in **NAN**)_ and then correcting it. Can we update the functionality so that we correct the issue where this NaN value is getting generated.

Satendra-SR:
done with feedback

resources/assets/js/components/ProjectStageBillingComponent.vue
```diff
@@ -37,13 +37,13 @@
         props: ['billing', 'index', 'stageCostWithGst', 'gstAmount', 'stageCostWithoutGst', 'currency', 'clientCountryGstApplicable'],
         computed: {
             billingCostWithoutGst: function() {
-                return this.stageCostWithoutGst  !== 'NaN' ? parseFloat((this.billing.percentage/100)*this.stageCostWithoutGst).toFixed(2) : parseFloat(0).toFixed(2);
+                return this.stageCostWithoutGst  == 0 ?parseFloat(0).toFixed(2) : parseFloat((this.billing.percentage/100)*this.stageCostWithoutGst).toFixed(2);
```
bohraji: Please correct the formatting on this.

Satendra-SR:
done with formating

rathorevaibhav:
@Satendra-SR when editing an existing project stage or adding billing to it, the functionality is working great.

When creating a new stage and then adding an empty billing, it's showing `$NaN`. Please check.
image

Satendra-SR:
I forgot to check the **null** value. now it is working fine.

rathorevaibhav:
have created #621 for one bug. Should target it separately.

Otherwise looking good. Merging.

Issue #478, opened 2018-06-18
Receivable or Balance due should show sum of all previous pending amount


rathorevaibhav:
Currently, when viewing the finance reports, the receivable/balance-due only shows the sum of numbers for the filtered timespan.

It should show the sum of all previous dues till the filtered end date.

For example, consider an invoice sent in May 2016 which is unpaid for two months. Now when we see the reports for July, the unpaid invoice amount should also show up in receivable.

For this, we can divide the receivable into two parts:
1. Receivable for filtered timespan (this is the current receivable calculation).
2. Total receivable till the filtered end date (will be the sum of all receivables from the beginning of time till the filtered end date).

PR #602, opened 2018-07-23
#478 total receivable

Satendra-SR:
#478 

### What I did:

- create a method `filterByEndDates` in `Invoice` Modal.
- enhanced my previous mistake;
` $unpaidInvoices = Invoice::where('status', 'unpaid')->orWhere(function ($query) use ($endDate) {
           $query->where('sent_on', '<=', $endDate);
         })->get();`

and replaced with

 `$invoices = Invoice::filterByEndDate($endDate);`.
- previously, it's functionality was not fulfill the requirement need. means it was not performed as expected. this time. I solved it.

### DIFF
--- a/app/Http/Controllers/Finance/ReportsController.php
+++ b/app/Http/Controllers/Finance/ReportsController.php
@@ -75,14 +75,16 @@ class ReportsController extends Controller
         $attr['paidInvoices'] = $arrangedInvoices['paid'];
 
         $attr['report'] = self::getCumulativeAmounts($arrangedInvoices['sent'], $arrangedInvoices['paid']);
-        $unpaidInvoices = Invoice::where('status', 'unpaid')->orWhere(function ($query) use ($endDate) {
+        $invoices = Invoice::where('status', 'unpaid')->orWhere(function ($query) use ($endDate) {
             $query->where('sent_on', '<=', $endDate);
         })->get();
         foreach (config('constants.currency') as $currency => $currencyMeta) {
             $attr['totalReceivables'][$currency] = 0;
         }
-        foreach ($unpaidInvoices as $unpaidInvoice) {
-            $attr['totalReceivables'][$unpaidInvoice->currency_sent_amount] += $unpaidInvoice->sent_amount;
+        foreach ($invoices as $invoice) {
+            if ($invoice->status != 'paid') {
+                $attr['totalReceivables'][$invoice->currency_sent_amount] += $invoice->sent_amount;
+            }
         }
 
         return $attr;
--- a/app/Http/Controllers/Finance/ReportsController.php
+++ b/app/Http/Controllers/Finance/ReportsController.php
@@ -75,9 +75,7 @@ class ReportsController extends Controller
         $attr['paidInvoices'] = $arrangedInvoices['paid'];
 
         $attr['report'] = self::getCumulativeAmounts($arrangedInvoices['sent'], $arrangedInvoices['paid']);
-        $invoices = Invoice::where('status', 'unpaid')->orWhere(function ($query) use ($endDate) {
-            $query->where('sent_on', '<=', $endDate);
-        })->get();
+        $invoices = Invoice::filterByEndDates($endDate);
         foreach (config('constants.currency') as $currency => $currencyMeta) {
             $attr['totalReceivables'][$currency] = 0;
         }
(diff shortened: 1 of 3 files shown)


Satendra-SR:
@bohraji 
one thing I want to ask you that if you notice I have used two `for` loop. and it can be optimized.
but the first loop
` foreach (config('constants.currency') as $currency => $currencyMeta) {
              $attr['totalReceivables'][$currency] = 0;
          }`
is running just twice. and if I put it inside the second loop, it will always check the condition.
so what should be the optimum way?

app/Http/Controllers/Finance/ReportsController.php
```diff
@@ -75,14 +75,14 @@ public static function getReportAttributes()
...
-            $query->where('sent_on', '<=', $endDate);
-        })->get();
+        $invoices = Invoice::filterByEndDates($endDate);
         foreach (config('constants.currency') as $currency => $currencyMeta) {
             $attr['totalReceivables'][$currency] = 0;
         }
-        foreach ($unpaidInvoices as $unpaidInvoice) {
-            $attr['totalReceivables'][$unpaidInvoice->currency_sent_amount] += $unpaidInvoice->sent_amount;
+        foreach ($invoices as $invoice) {
+            if ($invoice->status != 'paid') {
```
bohraji: @Satendra-SR In earlier implementation, we were only fetching the unpaid invoices. But in this updated implementation, we are fetching all the invoices before the date and then using only the `not - paid` ones. Please update to filtering only unpaid invoices when fetching it from database.

app/Models/Finance/Invoice.php
```diff
@@ -77,4 +77,13 @@ public function getClientAttribute()
     {
         return optional($this->projectStageBillings()->first()->projectStage->project)->client;
     }
+
+    public static function filterByEndDates($end, $paginated = false)
+    {
+        $invoices = self::with('projectStageBillings.projectStage.project.client')
+            ->whereDate('sent_on', '<=', $end)
```
bohraji: This filter is same as the one above where we are doing filtering using both `start date` and `end date`. Can we have a common query scope for both of them?

app/Http/Controllers/Finance/ReportsController.php
Satendra-SR: sure, I was trying to filter it in the controller. but I think that was a bad practice.
thanks

app/Models/Finance/Invoice.php
Satendra-SR: sure.

app/Models/Finance/Invoice.php
```diff
@@ -77,4 +77,13 @@ public function getClientAttribute()
     {
         return optional($this->projectStageBillings()->first()->projectStage->project)->client;
     }
+
+    public static function filterByEndDates($end, $paginated = false)
```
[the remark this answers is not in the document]
Satendra-SR: right. it makes sense. I think I forgot that one.

app/Models/Finance/Invoice.php
```diff
@@ -78,12 +78,15 @@ public function getClientAttribute()
...
     }
 
-    public static function filterByEndDates($end, $paginated = false)
+    public static function filterByEndDate($date, $paginated = false)
     {
-        $invoices = self::with('projectStageBillings.projectStage.project.client')
-            ->whereDate('sent_on', '<=', $end)
-            ->orderBy('sent_on', 'desc');
+        $unpaidInvoices = self::where(function ($query) use ($date) {
+            $query->where('sent_on', '<=', $date)
```
bohraji: Let's make this a query scope and reuse it here.

app/Models/Finance/Invoice.php
```diff
@@ -78,12 +78,15 @@ public function getClientAttribute()
...
+                ->where('status', 'unpaid');
```
bohraji: This will also be a query scope

app/Models/Finance/Invoice.php
```diff
@@ -78,12 +78,15 @@ public function getClientAttribute()
         return optional($this->projectStageBillings()->first()->projectStage->project)->client;
     }
 
-    public static function filterByEndDates($end, $paginated = false)
+    public static function filterByEndDate($date, $paginated = false)
```
bohraji: This function not only filters by the end date, but also filters by the unpaid invoices. We can have the filter for paid, unpaid as an optional parameter for the function like we currently have `paginated` var.

app/Models/Finance/Invoice.php
Satendra-SR: yes, you are right. I should do it as it can be used further in the future.

Issue #607, opened 2018-07-24
Store stages using ajax


Satendra-SR:
- At present, the stages are stored normally without ajax.
- We should store these stages using ajax

PR #614, opened 2018-07-25
607 store stages using ajax

Satendra-SR:
#607 
### what I did:

- earlier, the stage data was stored using the simple form.
- now, these data are stored using ajax.
- replace hardcoded routes with routes name.

### DIFF
--- a/resources/assets/js/components/ProjectStageComponent.vue
+++ b/resources/assets/js/components/ProjectStageComponent.vue
@@ -297,16 +297,11 @@
                 return parseFloat(this.inputStageCost).toFixed(2);
             },
         },
-        mounted: function(){
-
-                let route = this.stage.id ? '/project/stages/' + this.stage.id : '/project/stages';
-                console.log(route);
-
-        },
         methods: {
-            storeStages(){
+            async storeStages(){
                  let route = this.stage.id ? '/project/stages/' + this.stage.id : '/project/stages';
-                 axios.put(route, {name:this.stage.name,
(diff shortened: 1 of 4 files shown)


resources/assets/js/components/ProjectStageComponent.vue
```diff
@@ -300,6 +298,29 @@
...
+                 end_date:this.stage.end_date,
+                 type:this.stageType,
+                 project_id:this.projectId,
+                  });
+
+                 let methodName = 'post';
+                 let url = this.stageRoute;
+                 if(this.stage.id){
+                    methodName = 'put';
+                    url = this.stageRoute +'/' + this.stage.id;
```
bohraji: Let's store the routes for `updating` and `creating` stages separately and use them directly. Currently they seem to follow a pattern so its easy to make the `update` route dependent on the `create` route. But if they both change, then it will need to be updated in JS code, which defeats the purpose of using route name.

Satendra-SR:
I have created a new [Issue](https://github.com/ColoredCow/employee-portal/issues/619) related to the above feedback.

rathorevaibhav:
Nice work @Satendra-SR 👍 

Minor feedback on user experience though. When updating any stage, the stage collapses and the user has to reopen that stage by clicking on the "pencil" icon. Can that be avoided? We can target it separately. #622

Satendra-SR:
Thanks, @rathorevaibhav.
Surely I will work on those feedbacks.

Issue #566, opened 2018-07-12
Meaningful names for authorized users in the test cases


rathorevaibhav:
`$this->anAuthorizedUser()` can be replaced by `$this->signInAsAuthorizedUser()`.

Also, `$this->anAuthorizedUser()` creates a super-admin by default. We should use a user specific to that module. Checking super admin by default doesn't make sense IMO.

PR #616, opened 2018-07-25
(#566) test case cleanups

rathorevaibhav:
#566 

Changes include:

1. Remove `anAuthorizedUser` as it was confusing.
2. Updated `signIn` method. By default, it'll log in as an employee. 
3. Some cleanups in test files.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -9,10 +9,10 @@ abstract class TestCase extends BaseTestCase
 {
     use CreatesApplication;
 
-    protected function signIn($role = 'employee')
+    protected function signIn($role = null)
     {
         $user = create(User::class);
-        $user->assignRole($role);
+        is_null($role) ?: $user->assignRole($role);
         $this->be($user);
         return $this;
     }
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -2,15 +2,17 @@
 
 namespace Tests;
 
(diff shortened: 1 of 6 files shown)


tests/TestCase.php
```diff
@@ -2,15 +2,17 @@
...
 abstract class TestCase extends BaseTestCase
 {
     use CreatesApplication;
 
-    protected function signIn($user = null)
+    protected function signIn($role = 'employee')
     {
-        $user = $user ?: create('App\User');
+        $user = create(User::class);
+        $user->assignRole($role);
```
pankaj-ag: @rathorevaibhav I think default $role should be null. Otherwise, I won't be able to signin without any role.

rathorevaibhav: Not sure. Why do we need to sign-in without any role? There's nothing a person can do in functionality without any role.

Moreover, IMO, anyone who's onboarding will be an employee of an organization. I'm considering employee will be the role with lowest privileges.

pankaj-ag: > There's nothing a person can do in functionality without any role.

How will you test this?

rathorevaibhav:
@pankaj-ag done with the feedback.

Assigning no-role will make sense. For example, if we want to check something that is not accessible if a person is not an employee, we won't be able to test it if we assign the employee role to every user.

PR #617, opened 2018-07-25
ability to add calendar event

vikasrinvi:
#389 

- [X] ability to add the calendar event in the mail
   (if the calendar is created for the attendees then calendar event is also sent along in the mail)
- [X]  If the schedule is updated then there should be an update on the calendar as well.
      now there needs some changes in the `update schedule` which we track in next `issue`. so, for now, 
     we use the default `end date ` i.e `+30 minute` of `startdate`


vikasrnv:
# preview of mail sent to the attendee

the link is attached at the end of the mail body 

screenshot 38

resources/views/mail/hr/calendar-link.blade.php
```diff
@@ -0,0 +1,10 @@
+<div>
+{!! $applications->mail_body !!}
+    <div>
+        @if($applications->calendar_event != Null)
+
+        <P>Please <a href= "{{ $applications->calendar_event }}">click here</a> to check the calendar event </P>
```
bohraji: @vikasrnv Please update the formatting of `<p>` tags.

app/Services/CalendarEventService.php
```diff
@@ -63,6 +63,25 @@ public function fetch($eventId, $calendarId = 'primary')
         $this->setEndDateTime($event->end->dateTime, $event->end->timeZone);
         $this->id = $eventId;
     }
+    public function update($eventId, $applicationRound, $calendarId = 'primary')
+    {
+        $event = $this->service->events->get($calendarId, $eventId);
+
+        $this->setStartDateTime($applicationRound->scheduled_date, $event->start->timeZone);
+
+        $endDateTime = date('Y-m-d H:i:s', strtotime('+30 minutes', strtotime(date($applicationRound->scheduled_date))));
```
bohraji: @vikasrnv when creating a calendar invite, we are taking the event end date as an input from the user. However, in this update function we are by default creating events of `30minutes`. Please confirm.

vikasrnv: @bohraji now there needs some change in the `update schedule` which we track in next issue. so, for now, we use the default `end date ` i.e `+30 minute` of `startdate` 
I will also update this in PR description.

app/Mail/HR/Applicant/RoundReviewed.php
```diff
@@ -42,8 +42,6 @@ public function build()
             ->bcc(config('constants.hr.default.email'))
             ->from(config('constants.hr.default.email'), config('constants.hr.default.name'))
             ->subject($this->applicationRound->mail_subject)
-            ->view('mail.plain')->with([
-            'body' => $this->applicationRound->mail_body,
-        ]);
+            ->view('mail.hr.calendar-link', ['applications' => $this->applicationRound]);
```
rathorevaibhav: I think the better name for the email template would be `mail.hr.application-round-feedback`. Also, please fix the `applications` key. It should be `applicationRound`.

app/Observers/HR/ApplicationRoundObserver.php
```diff
@@ -51,4 +51,18 @@ public function createCalendarEvent(ApplicationRound $applicationRound)
             'calendar_event' => $event->id,
         ]);
     }
+
+    public function updated(ApplicationRound $applicationRound)
+    {
+        if (request()->action == "schedule-update") {
```
rathorevaibhav: can we define all these actions in the config? This'll help to maintain the consistency. Maybe in a separate PR?

rathorevaibhav: Please avoid using double-quotes in PHP wherever possible. I'm sure you know the difference between double and single quotes in PHP.

app/Services/CalendarEventService.php
```diff
@@ -63,6 +63,25 @@ public function fetch($eventId, $calendarId = 'primary')
         $this->setEndDateTime($event->end->dateTime, $event->end->timeZone);
         $this->id = $eventId;
     }
+    public function update($eventId, $applicationRound, $calendarId = 'primary')
+    {
+        $event = $this->service->events->get($calendarId, $eventId);
+
+        $this->setStartDateTime($applicationRound->scheduled_date, $event->start->timeZone);
```
rathorevaibhav: this is a service class. It shouldn't contain anything specific to your application logic. You shouldn't pass `$applicationRound`.
Suppose if there's another feature to update calendar event for employee meeting, you won't be able to use this method as it'll expect an application round. Instead, it should accept a `dates` array containing a start and end date.

app/Services/CalendarEventService.php
rathorevaibhav: The date time format is defined in `config/constants.php`. Please use that.

Also, use Carbon class as we're using it at all places.

app/Services/CalendarEventService.php
```diff
@@ -63,6 +63,25 @@ public function fetch($eventId, $calendarId = 'primary')
...
+
+        $endDateTime = date('Y-m-d H:i:s', strtotime('+30 minutes', strtotime(date($applicationRound->scheduled_date))));
+        $this->endDateTime = $this->startDateTime;
+        $this->endDateTime['dateTime'] = Carbon::parse($endDateTime)->format(config('constants.calendar_datetime_format'));
```
rathorevaibhav: you must use `setEndDateTime` to set the values.

app/Services/CalendarEventService.php
```diff
@@ -63,6 +63,25 @@ public function fetch($eventId, $calendarId = 'primary')
...
         $this->id = $eventId;
     }
+    public function update($eventId, $applicationRound, $calendarId = 'primary')
+    {
+        $event = $this->service->events->get($calendarId, $eventId);
+
+        $this->setStartDateTime($applicationRound->scheduled_date, $event->start->timeZone);
+
+        $endDateTime = date('Y-m-d H:i:s', strtotime('+30 minutes', strtotime(date($applicationRound->scheduled_date))));
+        $this->endDateTime = $this->startDateTime;
```
rathorevaibhav: what does this line represent?

app/Services/CalendarEventService.php
```diff
@@ -63,6 +63,25 @@ public function fetch($eventId, $calendarId = 'primary')
...
+        $this->endDateTime['dateTime'] = Carbon::parse($endDateTime)->format(config('constants.calendar_datetime_format'));
+
+        $eventDate = new Google_Service_Calendar_Event([
+            'start' => $this->startDateTime,
```
rathorevaibhav: you must use `$this->getStartDateTime()` to retrieve values instead of directly accessing the values. If you're having doubts over why getters and setters are used, you can read about them here: http://thisinterestsme.com/php-getters-and-setters/

app/Observers/HR/ApplicationRoundObserver.php
vikasrnv: @rathorevaibhav I think its better to track this in next PR

rathorevaibhav: Can you create an issue for this and mention it here?

app/Mail/HR/Applicant/RoundReviewed.php
rathorevaibhav: Any update on the mail template?

resources/views/mail/hr/application-round-feedback.blade.php
```diff
@@ -0,0 +1,10 @@
+<div>
+{!! $applicationRound->mail_body !!}
+    <div>
+        @if($applicationRound->calendar_event != Null)
+
+        <p>Please <a href= "{{ $applicationRound->calendar_event }}">click here</a> to check the calendar event </p>
```
rathorevaibhav: can we add the ICS file instead of mentioning the link? the mail body will contain the complete mail with the signature at the end which generally represents the end of the mail. Now if the calendar link comes below it, there are high chances the recipient may miss it.

Issue #273, opened 2018-05-23
Show the number of copies of a book


pankaj-ag:
##### Requirements
1. Add a new column `number_of_copies` in the database table `library_books`
1. This count will appear on the book card of that specific book
1. The `super-admin` and `book_manager` can click on this number to update the count
1. When a new book is added, an input field will ask the user to enter the number of copies available in the library

rathorevaibhav:
@pankaj-ag I believe you meant `show the number of copies of a book`. Is that correct?

abhishek-pokhriyal:
@pankaj-ag I've updated description. Please let me know if I missed something.

abhishek-pokhriyal:
This issue is partially completed and the remaining work is being tracked in #629. Closing.

PR #628, opened 2018-07-30
273 book can have multiple copies

abhishek-pokhriyal:
#273 

### Work done in this PR
1. Wrote a test case for the feature that _the user can specify the number of copies of a book_.
1. When the user adds a new book, she is asked to input the number of copies of that book available in the library.
1. A new column `number_of_copies` is added to the `library_books` table which stores this count.

### Remaining work
Currently, the user can specify the count only at the time of adding a new book. The count cannot be updated once set.


abhishek-pokhriyal:
@wevekamola @bohraji please review this.

Issue #629, opened 2018-07-30
Option to update book copies count


abhishek-pokhriyal:
Currently, the user can provide the number of copies available at the time of adding a book. Once the book is saved the user cannot modify the count of the copies. So there should be some way from the UI from where the user can update the book count.

This issue completes issue #273 which is partially solved in the PR https://github.com/ColoredCow/employee-portal/pull/628.

PR #632, opened 2018-08-02
(#629) update copies count

abhishek-pokhriyal:
#629 

### Number of copies of a book can be updated

Currently, when adding a new book, there is an option to set the number of copies available. Once set this option cannot be updated from the UI. In this PR, an option is given just for that.

The option `Copies Available` is given below the `Update Catgories` in the dropdown.

### Note
There is some new code in `assets/js/app.js` so don't forget to run `npm run dev`


tests/Feature/KnowledgeCafe/Library/BookTest.php
```diff
@@ -109,4 +109,15 @@ public function a_book_can_have_multiple_copies()
         $book = Book::whereTitle('Test Book')->first();
         $this->assertTrue($book->number_of_copies == $book_copies);
     }
+
+    /** @test */
+    public function an_authorized_user_can_update_number_of_copies()
+    {
+        $updated_book_copies = rand(1, 10);
+        $this->signInAsSuperAdmin();
+        $book = create(Book::class, ['title' => 'Test Book']);
```
pankaj-ag: Can we assign the title name to a variable and then use it. 

> $title = "test_title"
>$book = create(Book::class, ['title' => $title]);

Issue #637, opened 2018-08-03
HR templates don't have dynamic template variables


rathorevaibhav:
The HR emails that we send after conducting each round, the templates don't have dynamic variables and the user has to manually enter applicant name and interview details in every email.

Add the template variables.

PR #646, opened 2018-08-06
template variable

mohitgusain:
#637
- [x] Template variables support in the HR round feedback emails.


rathorevaibhav:
Looks good on testing. Merging.

Issue #651, opened 2018-08-08
Redesigning invoice schema


rathorevaibhav:
We need to take a fresh look at the current schema for the invoice and remove all the redundant fields.

We're targeting the following points:
1. Proper relationship
2. Entity segregation
3. Default values, data-types, range etc.
4. Nomenclature

PR #652, opened 2018-08-08
(#651) invoice schema

rathorevaibhav:


database/migrations/2018_08_08_144909_create_invoices_table.php
```diff
@@ -0,0 +1,80 @@
...
+            // due amount.
+            // gst.
+            // tds.
+            // bank charges on fund transfer.
+            // bank service tax on fund transfer.
+            // conversion rate. does it makes sense here?
+
+            $table->increments('id');
+            $table->unsignedInteger('invoice_id');
+            $table->string('type');
```
rudresh: What is the meaning of type here?

rathorevaibhav: I was thinking it will contain values like `sent_amount`, `paid_amount`, `due_amount`, `gst`, `tds`, `bank_charges` etc. For some invoices, some of these might not be present and hence will have only those values which are applicable.

database/migrations/2018_08_08_144909_create_invoices_table.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function up()
+    {
+        Schema::create('invoices', function (Blueprint $table) {
+            $table->increments('id');
+            $table->unsignedInteger('project_invoice_id');
+            $table->string('status')->default('unpaid');
+            $table->date('sent_on');
+            $table->date('paid_on')->nullable();
+            $table->string('payment_type')->default('wire-transfer');
```
pankaj-ag: Just a suggestion, can we create a new table for payment_type and reference the id here. It will help us to save other information as well.

database/migrations/2018_08_08_144909_create_invoices_table.php
```diff
@@ -0,0 +1,80 @@
...
+                ->references('id')->on('invoices')
+                ->onDelete('cascade');
+        });
+
+        Schema::create('invoice_cheques', function (Blueprint $table) {
+            $table->increments('id');
+            $table->unsignedInteger('invoice_id');
+
+            $table->string('status');
+            $table->date('received_date')->nullable();
```
pankaj-ag: what about received_at, cleared_at and bounced_at

database/migrations/2018_08_08_144909_create_invoices_table.php
```diff
@@ -0,0 +1,80 @@
...
+            // gst.
+            // tds.
+            // bank charges on fund transfer.
+            // bank service tax on fund transfer.
+            // conversion rate. does it makes sense here?
+
+            $table->increments('id');
+            $table->unsignedInteger('invoice_id');
+            $table->string('type');
+            $table->string('currency', 3)->default(config('constants.countries.india.currency'));
```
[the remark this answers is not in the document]
rathorevaibhav: The service tax that bank charges on foreign exchange is always in INR.
GST and TDS is INR but that is only applicable to invoices for Indian clients.

I think apart from this, all other amounts will have the same currency.

Should we add the currency to the invoice and remove from here?

database/migrations/2018_08_08_144909_create_invoices_table.php
rathorevaibhav: We support three right now.
1. Wire transfer - no additional meta
2. Cash - no additional meta
3. Cheque - status, bounced, received, cleared dates

database/migrations/2018_08_08_144909_create_invoices_table.php
rathorevaibhav: Oh, I thought of it. But we only require date fields, not timestamps. Will it make sense if we do `received_at` for a date?

rathorevaibhav: Maybe `received_on`, `bounced_on` and `cleared_on`?

database/migrations/2018_08_08_144909_create_invoices_table.php
```diff
@@ -0,0 +1,80 @@
...
+            $table->string('payment_type')->default('wire-transfer');
+            $table->date('due_date');
+            $table->text('comments')->nullable();
+            $table->string('file_path')->nullable(); // string or text?
+            $table->timestamps();
+        });
+
+        Schema::create('invoice_amounts', function (Blueprint $table) {
+            // the following fields we're capturing right now will go to this table with their currencies
+            // sent amount.
```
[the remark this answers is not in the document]
rathorevaibhav: sent amount will always be required in the invoice.
paid will be optional at the start, but eventually if the invoice is being marked as paid, it should have a paid amount.
due amount is optional. It's the remaining balance that the client did not pay (or will pay later).

Yes, I think having a Payment Object Model makes more sense.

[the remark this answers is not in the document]
rathorevaibhav: If we are to track salaries and organization expenses later, I think Payment object will make sense to reuse. Thoughts?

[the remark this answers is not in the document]
rathorevaibhav: Then, it should have polymorphic columns also to see which entity invoice/employee-salary etc that payment is related to. Thoughts?

database/migrations/2018_08_08_144909_create_invoices_table.php
```diff
@@ -0,0 +1,80 @@
...
+            $table->string('currency', 3)->default(config('constants.countries.india.currency'));
+            $table->decimal('amount', 10, 2);
+            $table->timestamps();
+
+            $table->foreign('invoice_id')
+                ->references('id')->on('invoices')
+                ->onDelete('cascade');
+        });
+
+        Schema::create('invoice_cheques', function (Blueprint $table) {
```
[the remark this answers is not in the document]
rathorevaibhav: Agreed.

Issue #609, opened 2018-07-24
Extra white space in after apply  any functionality into   module  


mohitgusain:
**Initially Screen looks like**
![screenshot 28](https://user-images.githubusercontent.com/30264862/43142015-058972b0-8f75-11e8-8db9-a195183ba70d.png)

**After applied any functionality**
![screenshot 29](https://user-images.githubusercontent.com/30264862/43142102-40d95f56-8f75-11e8-8293-c854473903bd.png)

as we see in the second screen there is extra white space after **Edit Client** as a comparison to the above screen.
Also check it into another module once.

abhishek-pokhriyal:
@rathorevaibhav can we close this?

PR #655, opened 2018-08-10
(#609) removed extra line in alert status

abhishek-pokhriyal:
#609

Removed the extra line that comes after we click on closing the status alert.

### DIFF
--- a/resources/views/status.blade.php
+++ b/resources/views/status.blade.php
@@ -13,6 +13,5 @@
 		<p>{!! session('status') !!}</p>
 		<span class="status-close"><b>&times;</b></span>
 	</div>
-	<br>
 @endif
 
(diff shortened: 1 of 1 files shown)


resources/views/status.blade.php
```diff
@@ -13,6 +13,5 @@
 		<p>{!! session('status') !!}</p>
 		<span class="status-close"><b>&times;</b></span>
 	</div>
-	<br>
```
rathorevaibhav: why was this there in the first place? I believe we'll need this gap when the status is shown. This br tag should ideally be inside the status div tag. @abhishek-pokhriyal thoughts?

abhishek-pokhriyal: @rathorevaibhav make sense. I thought the line break after the alert message was not needed.

rathorevaibhav:
Since we're using this status alert box at many places, a thorough testing needs to be done to make sure there's no side effect to any module.

PR #662, opened 2018-08-18
411 offer letter

mohitgusain:
- [x] Offer letter field added to the Send for approval model

- [x] Status changed from pending to Send for approval

### DIFF
--- a/config/constants.php
+++ b/config/constants.php
@@ -109,7 +109,7 @@ return [
             ],
             'sent-for-approval' => [
                 'label' => 'sent-for-approval',
-                'title' => 'Pending',
+                'title' => 'Sent-For-Approval',
                 'class' => 'badge badge-info',
             ],
             'confirmed' => [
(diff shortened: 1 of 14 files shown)


config/constants.php
```diff
@@ -109,7 +109,7 @@
             ],
             'sent-for-approval' => [
                 'label' => 'sent-for-approval',
-                'title' => 'Pending',
+                'title' => 'Sent-For-Approval',
```
rathorevaibhav: Title should be `Sent for approval`

resources/views/hr/application/index.blade.php
```diff
@@ -79,6 +71,15 @@
                 @endif
                 </a>
             </li>
+            <li class="nav-item">
```
rathorevaibhav: what does this change mean?

config/constants.php
mohitgusain: Ok.

resources/views/hr/application/index.blade.php
mohitgusain: Nothing else, but I thought just status should be in also in serial after one by one. Should I change it or make it as earlier?

Issue #678, opened 2018-09-08
Use ajax to generate offer letter


mohitgusain:
use ajax request instead of page reload while generating offer letter pdf.

PR #680, opened 2018-09-09
#678 Ui fix and generate offer letter using ajax

mohitgusain:
#678

### DIFF
--- a/config/pdf.php
+++ b/config/pdf.php
@@ -1,12 +1,12 @@
 <?php
 
 return [
-	'mode'                  => 'utf-8',
-	'format'                => 'A4',
-	'author'                => '',
-	'subject'               => '',
-	'keywords'              => '',
-	'creator'               => 'Laravel Pdf',
-	'display_mode'          => 'fullpage',
-	'tempDir'               => base_path('storage/app/temp/')
+   'mode'                  => 'utf-8',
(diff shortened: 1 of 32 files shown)


config/pdf.php
```diff
@@ -1,12 +1,12 @@
...
 return [
-	'mode'                  => 'utf-8',
-	'format'                => 'A4',
-	'author'                => '',
-	'subject'               => '',
-	'keywords'              => '',
-	'creator'               => 'Laravel Pdf',
-	'display_mode'          => 'fullpage',
-	'tempDir'               => base_path('storage/app/temp/')
+   'mode'                  => 'utf-8',
```
abhishek-pokhriyal: Is this PSR-2? Read about PSR-2 [here](https://www.php-fig.org/psr/psr-2/)

resources/views/hr/application/approve-applicant-modal.blade.php
```diff
@@ -1,4 +1,4 @@
-<div class="modal fade" id="approve_application" tabindex="-1" role="dialog" aria-labelledby="approve_application" aria-hidden="true" v-if="selectedAction == 'approve'">
```
abhishek-pokhriyal: why is this deleted?

PR #684, opened 2018-09-15
Dynami offer letter

Satendra-SR:
- removed "generate offer letter" option


resources/views/hr/application/approve-applicant-modal.blade.php
```diff
@@ -1,4 +1,4 @@
-<div class="modal fade" id="approve_application" tabindex="-1" role="dialog" aria-labelledby="approve_application" aria-hidden="true" v-if="selectedAction == 'approve'">
```
abhishek-pokhriyal: why is this removed?

Issue #696, opened 2018-10-13
Only the following options should show up if an applicant is approved:


vikasrinvi:
1. Onboard
2. Mark as rejected

PR #699, opened 2018-10-13
modal one time action

vikasrinvi:
#696 

Only the `onboard` and `rejected` option will shows after the applicant is approved

### DIFF
--- a/resources/views/hr/application/edit.blade.php
+++ b/resources/views/hr/application/edit.blade.php
@@ -198,7 +198,7 @@
                                     <select name="action_type" id="action_type" class="form-control w-50" v-model="selectedAction" data-application-job-rounds="{{ json_encode($application->job->rounds) }}">
                                         <option v-for="round in applicationJobRounds" value="round" :data-next-round-id="round.id">Move to @{{ round.name }}</option>
                                         <option value="send-for-approval">Send for approval</option>
-                                        @if ($applicationRound->round_status != config('constants.hr.status.approved.label') || $applicationRound->round_status != config('constants.hr.status.approved.label'))
+                                        @if ($applicationRound->round_status != config('constants.hr.status.onboarded.label') || $applicationRound->round_status != config('constants.hr.status.rejected.label'))
                                             <option value="approve">Approve</option>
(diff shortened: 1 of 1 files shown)


resources/views/hr/application/edit.blade.php
```diff
@@ -196,9 +196,11 @@
                                 <div class="d-flex align-items-center">
                                 @if ($applicationRound->showActions)
                                     <select name="action_type" id="action_type" class="form-control w-50" v-model="selectedAction" data-application-job-rounds="{{ json_encode($application->job->rounds) }}">
-                                        <option v-for="round in applicationJobRounds" value="round" :data-next-round-id="round.id">Move to @{{ round.name }}</option>
-                                        <option value="send-for-approval">Send for approval</option>
-                                        <option value="approve">Approve</option>
+                                    @if ($applicationRound->round_status != config('constants.hr.status.approved.label'))
```
abhishek-pokhriyal: Why are you checking the **Application Round** status (`$applicationRound->round_status`) and not **Application** status?

resources/views/hr/application/edit.blade.php
```diff
@@ -196,9 +196,11 @@
                                 <div class="d-flex align-items-center">
                                 @if ($applicationRound->showActions)
                                     <select name="action_type" id="action_type" class="form-control w-50" v-model="selectedAction" data-application-job-rounds="{{ json_encode($application->job->rounds) }}">
-                                        <option v-for="round in applicationJobRounds" value="round" :data-next-round-id="round.id">Move to @{{ round.name }}</option>
-                                        <option value="send-for-approval">Send for approval</option>
-                                        <option value="approve">Approve</option>
+                                    @if (!$application->isApproved())
```
abhishek-pokhriyal: why have you used `!$application->isApproved()` and not `$application->isRejected()`. I think the latter one is more readable. Thoughts?

vikasrinvi: `$application->isRejected()` will check if the application label is rejcted or not but here we need to check that if the application is already approved then there should be only two option i.e. `onboard` or `reject`.   making sense??

Issue #692, opened 2018-10-06
ApplicationRound - Model and Controller logic should be separated


mohitgusain:
- [x] Model Application round logic should be in the controller.

- [ ] Onboard create user logic should be in Event Listener /Model Observer.

abhishek-pokhriyal:
@Satendra-SR @vikasrinvi @mohitgusain This looks way too much complicated. Looking at the code, I've no clue how to proceed. A lot of code is written in `ApplicationRound` model (which, ideally, should not be there) changing which may affect (break) the functionality in other places. I'll need to discuss this @rathorevaibhav before taking any decision. Putting this on-hold for now.

vikasrinvi:
@abhishek-pokhriyal so when you targeting this discussion?

PR #709, opened 2018-10-27
#692 application round model cleanup

abhishek-pokhriyal:
#692

- [x] Model Application round logic moves to the controller.

### DIFF
--- a/app/Http/Controllers/HR/ApplicationRoundController.php
+++ b/app/Http/Controllers/HR/ApplicationRoundController.php
@@ -3,12 +3,17 @@
 namespace App\Http\Controllers\HR;
 
 use App\Helpers\ContentHelper;
+use App\Helpers\FileHelper;
 use App\Http\Controllers\Controller;
 use App\Http\Requests\HR\ApplicantRoundMailRequest;
 use App\Http\Requests\HR\ApplicationRoundRequest;
 use App\Mail\HR\Applicant\RoundReviewed;
+use App\Mail\HR\SendForApproval;
+use App\Mail\HR\SendOfferLetter;
 use App\Models\HR\ApplicantRound;
+use App\Models\HR\ApplicationMeta;
 use App\Models\HR\ApplicationRound;
+use App\User;
 use Carbon\Carbon;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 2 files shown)


app/Models/HR/ApplicationRound.php
```diff
@@ -165,7 +38,7 @@ public function updateOrCreateEvaluation($evaluations = [])
         return true;
     }
 
-    protected function _updateOrCreateReviews($reviews = [])
+    public function _updateOrCreateReviews($reviews = [])
```
vikasrinvi: @abhishek-pokhriyal  if this change is required

app/Http/Controllers/HR/ApplicationRoundController.php
```diff
@@ -24,9 +29,117 @@ class ApplicationRoundController extends Controller
...
+                // credentials bug in the Google API services is resolved.
+                $email = $attr['onboard_email'] . '@' . env('GOOGLE_CLIENT_HD');
+                $applicant->onboard($email, $attr['onboard_password'], [
+                    'designation' => $attr['designation'],
+                ]);
+
+                User::create([
+                    'email' => $email,
+                    'name' => $applicant->name,
+                    'password' => Hash::make($attr['onboard_password']),
```
vikasrinvi: `Illuminate\Support\Facades\Hash` class is missing

app/Http/Controllers/HR/ApplicationRoundController.php
```diff
@@ -24,9 +29,117 @@ class ApplicationRoundController extends Controller
...
+                    ]),
+                ]);
+
+                $subject = $validated['subject'];
+                $body = $validated['body'];
+
+                if (!$application->offer_letter) {
+                    $application->offer_letter = FileHelper::generateOfferLetter($application);
+                }
+                Mail::send(new SendOfferLetter($application, $subject, $body));
```
vikasrinvi: `Illuminate\Support\Facades\Mail` is missing

abhishek-pokhriyal: It's present: https://github.com/ColoredCow/employee-portal/blob/b6d0314c51080c8f1bc2e6dd4bc09f88d6f1c53a/app/Http/Controllers/HR/ApplicationRoundController.php#L20

app/Models/HR/ApplicationRound.php
abhishek-pokhriyal: This change (`protected` to `public`) is required because now this method needs to be called from outside this class (`ApplicationRound `). Earlier it was called from the inside the class so it was not a problem.

vikasrinvi:
@rathorevaibhav  We will track the second half of this problem in this issue #710 ( i.e.Onboard create user logic should be in Event Listener /Model Observer.)

Issue #717, opened 2019-07-04
User option to collect the book from library. 


pankaj-ag:

pankaj-ag:
Closing as complete.

PR #720, opened 2019-07-04
(#717)Update with master

pankaj-ag:
#717


pankaj-ag:
@rathorevaibhav I have changed the base for this branch from master.

PR #724, opened 2019-07-28
Hide book comment section

abhishek-pokhriyal:
### Two small fixes:

1. Hide comment section since it under not complete yet.
1.  Apply bootstrap's `img-fluid` class for making book image responsive.


pankaj-ag:
@abhishek-pokhriyal this looks good to me.

PR #725, opened 2019-08-02
Book comments for user

pankaj-ag:
- Created a new Vue component for comments. 
- Make a new controller for managing the comments.
- Added CURD operations for comments.
- Change the label from Book A month to A book a month.

### DIFF
--- a/app/Http/Controllers/CommentController.php
+++ b/app/Http/Controllers/CommentController.php
@@ -29,8 +29,5 @@ class CommentController extends Controller
      * @param  \App\Comment  $comment
      * @return \Illuminate\Http\Response
      */
-    public function destroy(Comment $comment)
-    {
-        //
-    }
+
 }
--- a/app/Http/Controllers/CommentController.php
+++ b/app/Http/Controllers/CommentController.php
@@ -7,20 +7,11 @@ use App\Models\Comment;
 
 class CommentController extends Controller
 {
-
     /**
      * Update the specified resource in storage.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @param  \App\Comment  $comment
-     * @return \Illuminate\Http\Response
-     */
-
-    /**
-     * Update the specified resource in storage.
-     *
-     * @param  \Illuminate\Http\Request  $request
-     * @param  \App\Comment  $comment
+     * @param  \App\Models\Comment  $comment
      * @return \Illuminate\Http\Response
      */
     public function update(Request $request, Comment $comment)
@@ -34,7 +25,7 @@ class CommentController extends Controller
     /**
      * Remove the specified resource from storage.
      *
-     * @param  \App\Comment  $comment
+     * @param  \App\Models\Comment  $comment
      * @return \Illuminate\Http\Response
      */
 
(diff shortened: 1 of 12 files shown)


app/Http/Controllers/CommentController.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Http\Request;
+use App\Models\Comment;
+
+class CommentController extends Controller
+{
+    /**
+     * Update the specified resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \App\Comment  $comment
```
rathorevaibhav: namespace correction

app/Http/Controllers/CommentController.php
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public function update(Request $request, Comment $comment)
+    {
+        $comment->update(['comment' => $request->comment]);
+        return response([
+            'status' => 'Comment updated successfully'
+        ]);
+    }
+
+    /**
```
rathorevaibhav: remove redundant docblock

app/Http/Controllers/KnowledgeCafe/Library/BookCommentController.php
```diff
@@ -15,8 +15,13 @@ public function store(Request $request, Book $book)
             'user_id' => auth()->id(),
             'comment' => $request->comment,
         ]);
-
         $book->comments()->save($comment);
-        return response()->json();
+        return response()->json($book->comments->last());
```
rathorevaibhav: will this make another query? If yes, do you think we can use this instead?
```suggestion
        return response()->json($comment);
```

resources/assets/js/components/Book/BooksCommentsComponent.vue
```diff
@@ -14,22 +22,32 @@
 
 <script>
     export default {
-        props: ['book', 'new_comment_route', 'comments'],
+        props: ['book', 'new_comment_route', 'book_comments', 'user'],
```
rathorevaibhav: Can we do this? and directly use `comments` variable at all places? instead of having two variables `all_comments` and `book_comments`
```suggestion
        props: ['book', 'new_comment_route', 'comments', 'user'],
```

resources/assets/js/components/Book/BooksCommentsComponent.vue
```diff
@@ -14,22 +22,32 @@
...
-                    comment:response.comment
-                });
-           }
+                this.new_comment = '';
+                this.all_comments.push(response.data);
+           },
+
+           async deleteComment(data) {
+
+               console.log(`${this.new_comment_route}/${data.comment.id}`);
```
rathorevaibhav: remove

resources/assets/js/components/Book/BooksCommentsComponent.vue
rathorevaibhav: also I'm not sure. Are we using camelCasing or snake_casing?

app/Http/Controllers/KnowledgeCafe/Library/BookCommentController.php
pankaj-ag: No $book->comments is a collection.

resources/assets/js/components/Book/BooksCommentsComponent.vue
pankaj-ag: I am not able to send the props as camel case in vue js. 

Do you have any idea ?

PR #728, opened 2019-09-16
Kindle books

pankaj-ag:
- Mark book for Kindle

### DIFF
--- a/app/Models/KnowledgeCafe/Library/Book.php
+++ b/app/Models/KnowledgeCafe/Library/Book.php
@@ -41,7 +41,7 @@ class Book extends Model
         return $query->where('on_kindle', true);
     }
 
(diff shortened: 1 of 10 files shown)


app/Models/KnowledgeCafe/Library/Book.php
```diff
@@ -23,21 +22,30 @@ public function categories()
...
             ->orderBy('readers_count', 'desc')
             ->get();
     }
 
+    public function scopeKindle($query)
+    {
+        return $query->where('on_kindle', true);
+    }
+
+    public function scopeExcludeKindle($query)
```
rathorevaibhav: I think a better name for this would be 
```suggestion
    public function scopeNotOnKindle($query)
```

PR #730, opened 2019-11-17
Invoice updates

pankaj-ag:
We need to experiment with wkhtmltopdf in the employee portal. 

I have set up a dummy template which is using hardcoded values.


resources/views/project/invoice-templates/Untitled-2
```diff
@@ -0,0 +1,116 @@
+<!DOCTYPE html>
```
rathorevaibhav: please rename the template.

Issue #805, opened 2021-07-21
University auto create and mapping with applicant


rathorevaibhav:
**Describe the task**
When a new application is created, the system should update the universities table if the name of the college doesn't exist in the table.

![image](https://user-images.githubusercontent.com/12053186/125084498-3cf9c080-e0e7-11eb-8792-73982ac14c66.png)

![image](https://user-images.githubusercontent.com/12053186/125083690-52bab600-e0e6-11eb-82c2-08b9032a0ba0.png)

**Expected behavior**
1. Once an application is entered, it should check if the college is already present in the universities list or not.
2. If yes, update the `hr_university_id` in the `hr_applicants` table.
3. If no, check in aliases of the universities.
4. If an alias matches, update the `hr_university_id` in the `hr_applicants` table.
5. If an alias doesn't match, then:
    1. Create a new entry in the `hr_universities` table.
    2. Update the `hr_university_id` in the `hr_applicants` table.

PR #846, opened 2021-07-23
Added University auto create and mapping with applicant functionality

am-Anmol:
Target #805 
## Summary
1.First checking if `college` is entered or not.
2.Then checking `university` is present or not in the database, if
    i.)Present then updating `hr_university_id`.
    ii.)Not Present then creating a new entry in the `university` table and updating `hr_university_id`.


Modules/HR/Entities/Applicant.php
```diff
@@ -36,6 +36,16 @@ public static function _create($attr)
             'linkedin' => isset($attr['linkedin']) ? $attr['linkedin'] : null,
         ]);
 
+        if (isset($attr['college']))
```
rathorevaibhav, line 39: let's move this code to the model observer class.
https://laravel.com/docs/8.x/eloquent#observers

Modules/HR/Entities/Applicant.php
```diff
@@ -36,6 +36,16 @@ public static function _create($attr)
...
+        {   $university = University::select('id')->where('name', $attr['college'])->first();
```
rathorevaibhav, line 40: please properly format

rathorevaibhav: The issue also says to check in university aliases. I don't see code for that. Am I missing anything?

Modules/HR/Entities/Applicant.php
```diff
@@ -36,6 +36,16 @@ public static function _create($attr)
...
+            if($university==null)
+            { 
+                University::create(['name'=>$attr['college']]);
+                $university = University::select('id')->where('name', $attr['college'])->first();
+            }  
+            Applicant::where('email', $attr['email'])->update(['hr_university_id' => $university->id]);
```
rathorevaibhav, line 46:
```suggestion
            $applicant->update(['hr_university_id' => $university->id]);
```

rathorevaibhav (changes requested):

Modules/HR/Entities/Applicant.php
am-Anmol: ->@rathorevaibhav Can I create a custom static method that checks university and fires events in the same file `Modules/HR/Entities/Applicant.php`? like `public static function checkUniversity(parameters){ }`
->Or should I create a model observer class only?

Modules/HR/Entities/Applicant.php
am-Anmol: I think there is no need to check in the `university aliases` table because we can add the `University Aliases` name only if the university exists in `hr_universities`.

Modules/HR/Entities/Applicant.php
rathorevaibhav: I'd prefer observer as it'll help to declutter the code in the model file and only relating to handling applicant create/update/store functionality.

Modules/HR/Entities/Applicant.php
rathorevaibhav: So consider this scenario:
> Universities table has entry for name="Graphic Era Dehradun"
> Someone may enter "Graphic Era (Deemed to be University)" in their college name
> Now based on your code, it will create a new university and now system will have two universities but in reality they are one.
> If you had Graphic Era (Deemed to be University) as an alias set for existing university, then it will map it to the right one.

am-Anmol:
## Summary
1. Moved code to Model Observer class named `ApplicantObserver`.
2. Approach followed for checking university and mapping explained through flowchart below :

image

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,43 @@
...
+class ApplicantObserver
+{
+    /**
+     * Listen to the Applicant create event.
+     *
+     * @param  \Modules\HR\Entities\Applicant  $Applicant
+     * @return void
+     */
+    public function created(Applicant $applicant)
+    {   $data = request()->all();
```
rathorevaibhav, line 18: formatting.

rathorevaibhav: Did this file get automatically formatted with phpcs? Can you check manually?

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,43 @@
...
+        if(isset($data['college']))
+        {
+            $university = University::select('id')->where('name', $data['college'])->first();
+            if($university==null)
+            {
+                $universityAlias = UniversityAlias::select('hr_university_id')->where('name', $data['college'])->first();
+                if($universityAlias==null)
+                {
+                    University::create(['name'=>$data['college']]);
+                    $university = University::select('id')->where('name', $data['college'])->first();
```
rathorevaibhav, line 28: Don't need second query
```suggestion
                    $university = University::create(['name'=>$data['college']]);
```

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,43 @@
...
+     * @return void
+     */
+    public function created(Applicant $applicant)
+    {   $data = request()->all();
+        if(isset($data['college']))
+        {
+            $university = University::select('id')->where('name', $data['college'])->first();
+            if($university==null)
+            {
+                $universityAlias = UniversityAlias::select('hr_university_id')->where('name', $data['college'])->first();
```
rathorevaibhav, line 24: you don't need to make another query to get aliases. While querying university, you can get the alias relationship also. Something like:
```php
// with aliases or whatever the relationship name is defined in University model
$university = University::with(['aliases'])->where('name', $data['college'])->first();

// then access it via
$university->aliases()
```

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,43 @@
...
+                if($universityAlias==null)
+                {
+                    University::create(['name'=>$data['college']]);
+                    $university = University::select('id')->where('name', $data['college'])->first();
+                    $applicant->update(['hr_university_id' => $university->id]);
+                }
+                else
+                {
+                    $university = UniversityAlias::select('hr_university_id')->where('name', $data['college'])->first();
```
rathorevaibhav, line 33: if you already have the $aliases as PHP variable, you can search within the array. No need to make another DB query.

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,43 @@
...
+                    $applicant->update(['hr_university_id' => $university->hr_university_id]);
```
rathorevaibhav, line 34: this `else` will run when `$universityAlias==null` is false, which basically means there are aliases present for the university.

However, what happens if, within the aliases array, you're unable to find the university name match? In that case, it should create a new university, right? But it doesn't seem to be happening now.

rathorevaibhav (changes requested):

Modules/HR/Observers/Recruitment/ApplicantObserver.php
am-Anmol: If within the aliases array if unable to find the university name match then this `if` condition works and creates a new university
```
   if($universityAlias==null)
    {
        $university = University::create(['name'=>$data['college']]);
        $applicant->update(['hr_university_id' => $university->id]);
    }
```

Modules/HR/Observers/Recruitment/ApplicantObserver.php
am-Anmol: Actually this code is `$university = University::with(['aliases'])->where('name', $data['college'])->first();` is not returning desired result as expected rather than returning `NULL` ,so that's why using another query for it.

Modules/HR/Entities/Applicant.php
```diff
@@ -35,7 +35,7 @@ public static function _create($attr)
             'course' => isset($attr['course']) ? $attr['course'] : null,
             'linkedin' => isset($attr['linkedin']) ? $attr['linkedin'] : null,
         ]);
-
+
```
kuldeep3, line 38: Remove extra indentation

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,43 @@
...
+                else
+                {
+                    $applicant->update(['hr_university_id' => $universityAlias->hr_university_id]);
+                }
+            }
+            else
+            {
+                $applicant->update(['hr_university_id' => $university->id]);
+            }
+        }
```
kuldeep3, line 41:
```suggestion
        if (isset($data['college'])) {
            $university = University::select('id')->where('name', $data['college'])->first();
            if ($university == null) {
                $universityAlias = UniversityAlias::select('hr_university_id')->where('name', $data['college'])->first();
                if ($universityAlias == null) {
                    $university = University::create(['name' => $data['college']]);
                    $applicant->update(['hr_university_id' => $university->id]);
                } else {
                    $applicant->update(['hr_university_id' => $universityAlias->hr_university_id]);
                }
            } else {
                $applicant->update(['hr_university_id' => $university->id]);
            }
        }
```

kuldeep3 (commented):
Please fix the indentation issue and then reassign it for review.

Modules/HR/Entities/Applicant.php
am-Anmol: Ok

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function created(Applicant $applicant)
+    {
+        $data = request()->all();
+
+        if (isset($data['college'])) {
+            $university = University::select('id')->where('name', $data['college'])->first();
+            if ($university==null) {
+                $universityAlias = UniversityAlias::select('hr_university_id')->where('name', $data['college'])->first();
+                if ($universityAlias==null) {
```
P4NK4J, line 25:
```suggestion
                if ($universityAlias == null) {
```

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,36 @@
...
+     * @param  \Modules\HR\Entities\Applicant  $Applicant
+     * @return void
+     */
+    public function created(Applicant $applicant)
+    {
+        $data = request()->all();
+
+        if (isset($data['college'])) {
+            $university = University::select('id')->where('name', $data['college'])->first();
+            if ($university==null) {
```
P4NK4J, line 23:
```suggestion
            if ($university == null) {
```

P4NK4J (commented):
Minor comments, else looking good.

Modules/HR/Observers/Recruitment/ApplicantObserver.php
am-Anmol: Ok, I'll do that.

P4NK4J (approved):
Looks good.

P4NK4J:
@rathorevaibhav , Since this PR involves your suggestions, hence it would be better if you could also have a look and approve the PR.

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,36 @@
...
+     * @param  \Modules\HR\Entities\Applicant  $Applicant
+     * @return void
+     */
+    public function created(Applicant $applicant)
+    {
+        $data = request()->all();
+
+        if (isset($data['college'])) {
+            $university = University::select('id')->where('name', $data['college'])->first();
+            if ($university == null) {
```
GauravGusain98, line 23: You need to make your code short. You can write it like this:

```suggestion
            if (!$university) {
```

GauravGusain98, line 23: You need to make your code short. You can write it like this:

```suggestion
            if (!$university) {
```

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,36 @@
...
+                $universityAlias = UniversityAlias::select('hr_university_id')->where('name', $data['college'])->first();
+                if ($universityAlias == null) {
```
GauravGusain98, line 25: same here

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,36 @@
...
+                    $applicant->update(['hr_university_id' => $university->id]);
+                } else {
+                    $applicant->update(['hr_university_id' => $universityAlias->hr_university_id]);
+                }
+            } else {
+                $applicant->update(['hr_university_id' => $university->id]);
+            }
+        }
+    }
+}
```
GauravGusain98, line 36: Try avoiding, so many nested if-else conditions. You can use the `return` inside the `if` condition if you think the code will not run after it if the condition is true.

Give it a try.

GauravGusain98 (changes requested):
Please have look at the comments and make the suggested changes.

Modules/HR/Observers/Recruitment/ApplicantObserver.php
am-Anmol: Ok, I made these changes.

Modules/HR/Observers/Recruitment/ApplicantObserver.php
am-Anmol: Ok, I made these changes.

Modules/HR/Observers/Recruitment/ApplicantObserver.php
am-Anmol: I have tried but not working as expected so going with the previous one.

Modules/HR/Observers/Recruitment/ApplicantObserver.php
```diff
@@ -0,0 +1,36 @@
...
+                if (!$universityAlias) {
+                    $university = University::create(['name'=>$data['college']]);
+                    $applicant->update(['hr_university_id' => $university->id]);
+                } else {
+                    $applicant->update(['hr_university_id' => $universityAlias->hr_university_id]);
+                }
+            } else {
+                $applicant->update(['hr_university_id' => $university->id]);
+            }
+        }
```
GauravGusain98, line 34:
```suggestion
        if (isset($data['college'])) {
            $university = University::select('id')->where('name', $data['college'])->first();
            if (!$university) {
                $universityAlias = UniversityAlias::select('hr_university_id')->where('name', $data['college'])->first();
                if (!$universityAlias) {
                    $university = University::create(['name'=>$data['college']]);
                    $applicant->update(['hr_university_id' => $university->id]);
                } else {
                    $applicant->update(['hr_university_id' => $universityAlias->hr_university_id]);
                }
                return;
            } 
            $applicant->update(['hr_university_id' => $university->id]);
        }
```


This is what I meant by reducing the conditions, as using too many conditions is not a good way of writing code. Can you test it using this approach?

GauravGusain98 (changes requested):
Looks good to me. Just some minor fixes. Please make these changes

Modules/HR/Observers/Recruitment/ApplicantObserver.php
am-Anmol: Yes, this approach is working perfectly.

Modules/HR/Observers/Recruitment/ApplicantObserver.php
am-Anmol: yes, this file gets automatically format with phpcs.

P4NK4J (approved):
Looks good to me

PR #847, opened 2021-07-23
Continuous Deployment

rathorevaibhav:
## Summary
1. Production and staging builds
2. Removed automerge.yml - we're not using it anymore
3. Cleanups in existing documentation


.github/workflows/production-deployment.yml
```diff
@@ -0,0 +1,57 @@
...
+      with:
+        host: ${{ secrets.SSH_HOST }}
+        username: ${{ secrets.SSH_USERNAME }}
+        key: ${{ secrets.SSH_PRIVATE_KEY }}
+        envs: SSH_BUILD_DIRECTORY,SSH_POST_BUILD_SCRIPT
+        script: |
+          cd $SSH_BUILD_DIRECTORY
+
+          echo "Enabling maintenance mode"
+          php artisan down
```
pankaj-ag, line 31: How much downtime we are expecting? Or are we planning to make the late-night build?

.github/workflows/staging-deployment.yml
```diff
@@ -0,0 +1,58 @@
...
+      with:
+        host: ${{ secrets.SSH_HOST }}
+        username: ${{ secrets.SSH_USERNAME }}
+        key: ${{ secrets.SSH_PRIVATE_KEY }}
+        envs: SSH_BUILD_DIRECTORY,SSH_POST_BUILD_SCRIPT
+        script: |
+          cd $SSH_BUILD_DIRECTORY
+
+          echo "Enabling maintenance mode"
+          php artisan down
```
pankaj-ag, line 32: This looks fine.

pankaj-ag (changes requested):

.github/workflows/production-deployment.yml
rathorevaibhav: Around 5min. This gets trigger on releases. We'll be doing that probably late nights.

Also, these workflows will only trigger after manual approval from the GitHub environment setting.

pankaj-ag: Okay, I think it's fine.
We can also use this feature to protect some routes. 
https://stackoverflow.com/questions/30572829/laravel-maintenance-mode-just-for-some-urls

pankaj-ag (approved):

Issue #845, opened 2021-07-22
Opportunity delete when we update any field 


am-Anmol:
**Describe the bug**
In Opportunities Tab, while updating any opportunity details (like Job title, status, type, Job Description) it deletes this opportunity but ideally, it should update details.

**Steps to replicate**
1. Login and click on HR > Recruitment.
2. Click on the `Opportunities` tab.
3. Click on any Opportunity.
4. Then update any field `Job title`, `status`, `type`, `Job Description` and click on the `Update` button.
5. You will see the opportunity got deleted.

**Expected behavior**
It should update opportunity details.

**Screenshots**
![image](https://user-images.githubusercontent.com/48876126/126684579-56e57c64-d7b8-4f18-832a-a75227527865.png)

rathorevaibhav:
@am-Anmol this is not working on UAT. I changed the domain for a job opportunity and it got deleted after I update it.

rathorevaibhav:
This is working great on UAT. Thanks and great work @am-Anmol 👏

PR #857, opened 2021-07-28
Opportunity delete when update any opportunity 

am-Anmol:
Target #845 
## Summary
1. This bug is coming because there have been nesting of two forms one is for `update` and another one is for `delete`, so once we are clicking on the update button it's considering submitting for the delete button.
2. So changed nested to multiple forms.
3. And make that div to flexbox with some left margin to it.

### DIFF
--- a/resources/views/hr/job/edit.blade.php
+++ b/resources/views/hr/job/edit.blade.php
@@ -15,7 +15,7 @@
     <br><br>
     @include('status', ['errors' => $errors->all()])
     <h2 class="mb-3">Edit Opportunity</h2>
-    <form action="{{ $formAction }}" method="POST">
+    <form action="{{ $formAction }}" method="POST" id="update-form">
         @csrf
         @method('PATCH')
         <div class="card mb-3">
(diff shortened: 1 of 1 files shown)


resources/views/hr/job/edit.blade.php
```diff
@@ -118,15 +118,15 @@ class="form-control"
                 </div>
             </div>
         </div>
-        <div>
-            <button type="submit" class="btn btn-primary">Update</button>
-            <a href="{{ $job->link }}" target="_blank" class="btn btn-info" role="button">Preview job</a>
+        <div class="d-flex">
+            <button type="submit" class="btn btn-primary ml-2">Update</button>
+            </form>
```
rathorevaibhav, line 123: This isn't the right place to close this form. All child tags within this must be closed. However, you see the `<div>` tag isn't closing.

```html
<div class="d-flex"> <!-- this must close before form as it is a child element -->
    <button type="submit" class="btn btn-primary ml-2">Update</button>
    </form>
```

rathorevaibhav (changes requested):

am-Anmol:
## Summary
1. Added `form` attribute in submit button and assigned respective `id` to forms tag.
For more details on how to nest form in HTML5, please refer to this https://www.vivait.com.au/nested-forms-using-html5.

rathorevaibhav (approved):

PR #868, opened 2021-08-06
changed PHP versions to be installed.

srivastava-jyoti:
I have changed php versions to be installed in read.md file.

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -22,7 +22,7 @@ There are some topics and tools you need to know to prior to work in portal so p
 - [Git, GitHub and Development](https://product.hubspot.com/blog/git-and-github-tutorial-for-beginners)
 
 ### 4. PHP :running:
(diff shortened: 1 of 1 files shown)


pankaj-ag (approved):
It looks like a simple change. Can you redirect me to the issue for this PR? I am just curious, why 7.3 or higher was not sufficient.

srivastava-jyoti:
There was a compatibility problem of php version 8 with composer. If someone installs version 8 then, while installing composer on git console one need to type,
composer install --ignore-platform-reqs
here, is the reference :-https://github.com/composer/composer/issues/9529

P4NK4J:
@rathorevaibhav , I think we would require updating composer files for higher PHP version compatibilities. Let me know if we can do that in a separate PR and if we are good to merge this pull request.

rathorevaibhav:
@P4NK4J yeah let's target in a separate branch as an overall tech stack update. Merging this.

PR #871, opened 2021-08-07
 Added the taxes in formula for calculating TDS

abhinav80066:
## Summary
The Amount used to calculate the TDS before was without Taxes so I have added Taxes in the formula to calculate the correct TDS.

### DIFF
--- a/Modules/Invoice/Resources/views/subviews/edit/invoice-details.blade.php
+++ b/Modules/Invoice/Resources/views/subviews/edit/invoice-details.blade.php
@@ -180,7 +180,7 @@
         },
 
         updateTds() {
-            this.tds = this.amount - this.amount_paid
+            this.tds = this.amount - this.amount_paid + (this.amount * 0.18)
             this.tds_percentage = (this.tds/this.amount) * 100
         },
 


Modules/Invoice/Resources/views/subviews/edit/invoice-details.blade.php
```diff
@@ -180,7 +180,7 @@
         },
 
         updateTds() {
-            this.tds = this.amount - this.amount_paid
+            this.tds = this.amount - this.amount_paid + (this.amount * 0.18)
```
kuldeep3, line 183: I think `0.18` is like `18%` GST. I am wondering if instead of hardcoding it here directly if we could fetch it from `env` but on the second thought, I know this will not change so frequently, so I am ok with adding this value directly here. What are your thoughts on this?

kuldeep3 (approved):
LGTM 👍🏽

Modules/Invoice/Resources/views/subviews/edit/invoice-details.blade.php
abhinav80066: Yes, I have the same thought on it, I saw gst hardcoded on other files and the fact that it will not change so frequently, so I think it would be good to use the value directly.

PR #875, opened 2021-08-08
Migration failing leading to unsuccessful deployments

abhinav80066:
## Summary
The Table name was not correct, so I have changed the table name to the correct one(invoices).


pankaj-ag:
We need to investigate why it did not fail on UAT server.

pankaj-ag (approved):

rathorevaibhav:
@pankaj-ag it failed on UAT. Testing never happened there.

Issue #848, opened 2021-07-23
Artisan route cache command throws an error


rathorevaibhav:
**Describe the task**
When running the command `php artisan route:cache`, it results in an error
![image](https://user-images.githubusercontent.com/12053186/126778397-39d62b83-e97f-4ffd-96a8-49e922bf3ece.png)

**Expected behavior**
The command should work fine without any errors.

**References**
https://laravel.com/docs/8.x/deployment#optimizing-route-loading

rathorevaibhav:
This works great. @kickb71 thanks!

kickb71:
No problem sir!

PR #876, opened 2021-08-08
PHP artisan error

kickb71:
If I run this command ("php artisan route:cache") it shows error message in the terminal

I have corrected and rectified all the files which is causing the error's not to run this command.

Issue no:#848 .

### DIFF
--- a/app/Http/Controllers/Settings/PermissionController.php
+++ b/app/Http/Controllers/Settings/PermissionController.php
@@ -47,7 +47,7 @@ class PermissionController extends Controller
 
     public function updateRolePermissions(UpdateRolePermissionsRequest $request)
     {
-        $this->authorize('settings.update-role', Permission::class);
+        $this->authorize('settings.update', Permission::class);
         $validatedData = $request->validated();
         if (!isset($validatedData['permissions'])) {
             return response()->json([
--- a/app/Http/Controllers/Settings/PermissionController.php
+++ b/app/Http/Controllers/Settings/PermissionController.php
@@ -47,7 +47,7 @@ class PermissionController extends Controller
 
     public function updateRolePermissions(UpdateRolePermissionsRequest $request)
     {
-        $this->authorize('settings.role', Permission::class);
+        $this->authorize('settings.update.role', Permission::class);
         $validatedData = $request->validated();
         if (!isset($validatedData['permissions'])) {
             return response()->json([
--- a/app/Http/Controllers/Settings/PermissionController.php
+++ b/app/Http/Controllers/Settings/PermissionController.php
@@ -47,7 +47,7 @@ class PermissionController extends Controller
 
     public function updateRolePermissions(UpdateRolePermissionsRequest $request)
     {
-        $this->authorize('settings.update', Permission::class);
+        $this->authorize('settings.role', Permission::class);
         $validatedData = $request->validated();
         if (!isset($validatedData['permissions'])) {
             return response()->json([
--- a/app/Http/Controllers/Settings/PermissionController.php
+++ b/app/Http/Controllers/Settings/PermissionController.php
@@ -47,7 +47,7 @@ class PermissionController extends Controller
 
     public function updateRolePermissions(UpdateRolePermissionsRequest $request)
     {
-        $this->authorize('settings.update.role', Permission::class);
+        $this->authorize('settings.update-role', Permission::class);
         $validatedData = $request->validated();
         if (!isset($validatedData['permissions'])) {
             return response()->json([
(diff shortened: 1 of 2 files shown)


routes/web.php
```diff
@@ -119,7 +118,7 @@
 
             Route::get('{module}', 'PermissionController@index')->name('permissions.module.index');
             Route::put('users/{id}', 'PermissionController@updateUserRoles')->name('permissions.module.update');
-            Route::put('roles/{id}', 'PermissionController@updateRolePermissions')->name('permissions.module.update');
+            Route::put('roles/{id}', 'PermissionController@updateRolePermissions')->name('permissions.module.role');
```
P4NK4J, line 121: @kickb71 , I think the route name should specify what it's intending to. Can we rename this route to `role-update`, The name `role` seems quite generic. Let me know your thoughts.

P4NK4J (commented):
Minor questions, let me know what you think.

routes/web.php
kickb71: That is what causing the error sir 
I have changed in the permisson controller to

GauravGusain98 (changes requested):
Please check the comments and make the necessary changes.

routes/web.php
kickb71: I will change the name to "permission.module.update.role" is that fine sir?

GauravGusain98: You can use `permission.module.update-role`

GauravGusain98: Also, does this route being used anywhere in the codebase? I can't see any changes related to the route name in the files from where this route is being called.

app/Http/Controllers/Settings/PermissionController.php
```diff
@@ -47,7 +47,7 @@ public function updateUserRoles(UpdateUserRolesRequest $request)
 
     public function updateRolePermissions(UpdateRolePermissionsRequest $request)
     {
-        $this->authorize('settings.update', Permission::class);
+        $this->authorize('settings.update.role', Permission::class);
         $validatedData = $request->validated();
```
GauravGusain98, line 51: Revert these changes. They are not required. The old code is correct. 

Also, the name inside the authorize function is not the name of the route. It is the name of the function that is defined in the policy file. But for now, just revert these changes.

routes/web.php
```diff
@@ -119,7 +118,7 @@
 
             Route::get('{module}', 'PermissionController@index')->name('permissions.module.index');
             Route::put('users/{id}', 'PermissionController@updateUserRoles')->name('permissions.module.update');
-            Route::put('roles/{id}', 'PermissionController@updateRolePermissions')->name('permissions.module.update');
+            Route::put('roles/{id}', 'PermissionController@updateRolePermissions')->name('permissions.module.update.role');
```
GauravGusain98, line 121: I see that these 2 routes are not used anywhere in the codebase. Not sure why they were added. Do you know where these routes are used? 

Also, you can use the name `permissions.module.update-role` for the route

GauravGusain98 (changes requested):
Some minor changes required. Please make the changes and confirm if the routes are in use anywhere or not.

app/Http/Controllers/Settings/PermissionController.php
kickb71: If we revert these changes we get error when we run the 'php artisan route:cache' command.

routes/web.php
kickb71: yes I can change the name to "permissions.module.update-role"

routes/web.php
kickb71: yes it is used in the "settings->permission->role.blade.php"

GauravGusain98: Got it. I see that the URL is being used here. So for now, you can just change the name or route to `permission.module.update-role`.

app/Http/Controllers/Settings/PermissionController.php
GauravGusain98: No, it will not throw an error. Just try it.

routes/web.php
GauravGusain98: yeah

app/Http/Controllers/Settings/PermissionController.php
kickb71: I have tried it
error

GauravGusain98: I told you to just revert the changes of the PermissionController file. The route names were correct.

GauravGusain98 (approved):
Looks good to me.

rathorevaibhav:
@GauravGusain98 do we need another review on this or it's good to merge? If good to merge, then please merge it after approval. For other PRs, also.

GauravGusain98:
> @GauravGusain98 do we need another review on this or it's good to merge? If good to merge, then please merge it after approval. For other PRs, also.

Sure, I'll do this from now onwards.

Issue #794, opened 2021-07-01
New job opportunity should get created on ColoredCow website also


rathorevaibhav:
**Describe the task**
Currently, we have separate mechanisms to create an entry on the career page on the website and work opportunities in ColoredCow Portal. When a new opportunity is created on the portal, it doesn't get created on the website.

**Expected behavior**
- [x] When a new `JOB` type opportunity is created on Portal, a new post type for career should also be created in website.
- [x] The opportunity should also be updated or deleted at website, if same action happens at Portal.

**Note:** This task will require both Portal and website setup.

**References**
Check out the Corcel package to easily create entries in the WordPress database from Laravel
https://github.com/corcel/corcel

am-Anmol:
@GauravGusain98 
Can you please share steps to replicate for `Draft jobs are still showing in website` because I'm unable to do so?

GauravGusain98:
This is a job in uat for software developer fresher, which is in draft state
![image](https://user-images.githubusercontent.com/54376313/131288286-fa6eedc0-0e15-4d16-a416-8af91681dcc9.png)


But I still see the job in the website 
![image](https://user-images.githubusercontent.com/54376313/131288365-3e5478b6-0404-47ff-a6a4-026ca29c81d2.png)

GauravGusain98:
Also, I tried to make a job published but was not able to see that on the website
![image](https://user-images.githubusercontent.com/54376313/131288596-5a5615f5-4f97-4431-ada7-36949d80babd.png)

![image](https://user-images.githubusercontent.com/54376313/131288626-55fcb78e-8438-4bde-8822-6a043936c58a.png)

rathorevaibhav:
@srivastava-jyoti please test this one also.

srivastava-jyoti:
@rathorevaibhav I have tested this. It is working fine on UAT. Drafts and closed ones are not showing on the UAT of the main website. Only Published ones are visible.

rathorevaibhav:
This is working great! Great work @am-Anmol 👏

PR #889, opened 2021-08-13
New Job Opportunity also reflect on ColoredCow Website

am-Anmol:
Targets #794 

### Description
**Job opportunities will `create/update/delete` on the `ColoredCow website` also if updated from `Employee Portal`.**
- Changes made in `Modules/HR/Observers/Recruitment/JobObserver.php` :
     - **`Created` function (working):**
          - Check if `status` is `published` or not, if it is `not` then do nothing.
          - If it is `published` then creating a new entry of data in `wp_post` table with `post_title,post_content,post_type` and a new entry in `wp_postmeta` table with `hr_id` and value of `hr_id` from employee portal `hr_jobs` table.
     - **`Updated` function (working):**
          - Check if `status` is `published` or not, if it is `not` then do nothing.
          - If it is `published` then creating a new entry of data in `wp_post` table with `post_title,post_content,post_type` and a new entry in `wp_postmeta` table with `hr_id` and value of `hr_id` from employee portal `hr_jobs` table.
          - Then finding `post id` by matching `hr_id` in `wp_postmeta` table (Website) with `hr_id` in `hr_jobs` table (employee portal).
          - After that finding an old post with `post id` and deleting the old post.
     - **`Deleted` function (working):**
          - Matching `post_title` and `post_type` with records and deleting the post.
      
- **Prerequisites Required:**
     - Update `config/database.php` with the correct database configurations.
     - image

**References:**
Check out the Corcel package to easily create entries in the WordPress database from Laravel
https://github.com/corcel/corcel
### Checklist:

### DIFF
--- a/Modules/HR/Observers/Recruitment/JobObserver.php
+++ b/Modules/HR/Observers/Recruitment/JobObserver.php
@@ -19,16 +19,15 @@ class JobObserver
         $job->rounds()->attach(Round::all()->pluck('id')->toArray());
         $data = request()->all();
 
-        
-            $Corcel = new Corcel();
-            $Corcel->post_title = $data['title'];
-            $Corcel->post_content = $data['description'];
-            $Corcel->post_type = config('hr.post-type.career');
-            if ($data['status'] != 'published') {
-                $Corcel->post_status = 'draft';
-            }
-            $Corcel->save();
-            $Corcel->saveMeta('hr_id', $job['id']);
+        $Corcel = new Corcel();
+        $Corcel->post_title = $data['title'];
+        $Corcel->post_content = $data['description'];
+        $Corcel->post_type = config('hr.post-type.career');
+        if ($data['status'] != 'published') {
+            $Corcel->post_status = 'draft';
+        }
+        $Corcel->save();
+        $Corcel->saveMeta('hr_id', $job['id']);
         
     }
 
@@ -42,17 +41,17 @@ class JobObserver
     {
         $data = request()->all();
 
-            $post = Corcel::hasMeta('hr_id', $job['id'])->first();
-            $Corcel=Corcel::find($post->ID);
-            $Corcel->post_title = $data['title'];
-            $Corcel->post_content = $data['description'];
-            $Corcel->post_type = config('hr.post-type.career');
-            if ($data['status'] == 'published') {
-                $Corcel->post_status = 'publish';
-            } else {
-                $Corcel->post_status = 'draft';
-            }
-            $Corcel->update();
+        $post = Corcel::hasMeta('hr_id', $job['id'])->first();
+        $Corcel=Corcel::find($post->ID);
+        $Corcel->post_title = $data['title'];
+        $Corcel->post_content = $data['description'];
+        $Corcel->post_type = config('hr.post-type.career');
+        if ($data['status'] == 'published') {
+            $Corcel->post_status = 'publish';
+        } else {
+            $Corcel->post_status = 'draft';
+        }
+        $Corcel->update();
     }
 
     /**
--- a/Modules/HR/Observers/Recruitment/JobObserver.php
+++ b/Modules/HR/Observers/Recruitment/JobObserver.php
@@ -45,11 +45,7 @@ class JobObserver
         $Corcel->post_title = $data['title'];
         $Corcel->post_content = $data['description'];
         $Corcel->post_type = config('hr.post-type.career');
-        if ($data['status'] == 'published') {
-            $Corcel->post_status = 'publish';
-        } else {
-            $Corcel->post_status = 'draft';
-        }
+        $Corcel->post_status = $data['status'] == 'published' ? 'publish' : 'draft';
         $Corcel->update();
     }
 
--- a/Modules/HR/Observers/Recruitment/JobObserver.php
+++ b/Modules/HR/Observers/Recruitment/JobObserver.php
@@ -28,7 +28,6 @@ class JobObserver
         }
         $Corcel->save();
         $Corcel->saveMeta('hr_id', $job['id']);
-        
     }
 
     /**
--- a/Modules/HR/Observers/Recruitment/JobObserver.php
+++ b/Modules/HR/Observers/Recruitment/JobObserver.php
@@ -4,7 +4,6 @@ namespace Modules\HR\Observers\Recruitment;
 
 use Modules\HR\Entities\Job;
 use Modules\HR\Entities\Round;
-use Modules\HR\Http\Controllers\Recruitment\JobController;
 use Corcel\Model\Post as Corcel;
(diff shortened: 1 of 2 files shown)


Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -16,5 +17,49 @@ class JobObserver
     public function created(Job $job)
     {
         $job->rounds()->attach(Round::all()->pluck('id')->toArray());
+        $data = request()->all();
+
+        if ($data['status'] == 'published') {
+            $Corcel = new Corcel();
+            $Corcel->post_title = $data['title'];
+            $Corcel->post_content = $data['description'];
+            $Corcel->post_type = 'career';
```
GauravGusain98, line 26: Please use the value from config files instead of hardcoding. This way if we want to change the value of post_type, we have to only change that in one place instead of changing at every place.

GauravGusain98, line 26: Please use the value from config files instead of hardcoding. This way if we want to change the value of post_type, we have to only change that in one place instead of changing at every place.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -16,5 +17,49 @@ class JobObserver
...
+            $Corcel->save();
+            $Corcel->saveMeta('hr_id', $job['id']);
+            $post = Corcel::hasMeta('hr_id', $job['id'])->first();
+            $corcel = Corcel::find($post->ID);
+            $corcel->delete();
```
GauravGusain98, line 51: This is not looking correct to me as we need to update the record here. But in this case, we are creating a new entry and deleting the old one. Please use the update approach for this.

GauravGusain98 (changes requested):
Please make the suggested changes.

Modules/HR/Observers/Recruitment/JobObserver.php
am-Anmol: OK, I'll update this approach.

Modules/HR/Observers/Recruitment/JobObserver.php
am-Anmol: Ok, I make these changes.

am-Anmol:
**Update function approach:**
- Check if `status` is `published` or not, if it is `not` then do nothing.
- If it is `published` then find `post ID` in `wp_post` table using `job_id`.
- Then finding `post` with `post ID` and updating respective columns.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -16,5 +17,42 @@ class JobObserver
...
+     * Listen to the Job update event.
+     *
+     * @param  \Modules\HR\Entities\Job  $job
+     * @return void
+     */
+    public function updated(Job $job)
+    {
+        $data = request()->all();
+
+        if ($data['status'] == 'published') {
```
GauravGusain98, line 42: Just a doubt, can there be a case when the status is not published and we can update the job? If yes, then we need to update the job details for both the case.

GauravGusain98 (commented):
Please check the comment

GauravGusain98:
Please check the above comment

Modules/HR/Observers/Recruitment/JobObserver.php
am-Anmol: Yes, this can be a possible scenario then we have to update both of them.
If we update the job from the portal and mark status to published it also gets updated on a website post.

GauravGusain98: Let me rephrase my question, 
I was saying that suppose a job is not published but you updated the job details in the portal. In this case as well, the job details should get updated in the website post. But as per the code, it looks like that the job will only get updated when the status is published.

Also what about the case when we update the job from `published` to `draft`. I don't think in that case it will mark the job as a draft on the website. 

Let me know if there is any doubt in understanding.

GauravGusain98 (changes requested):
Please check the comments and if required do the requested changes.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -16,5 +17,42 @@ class JobObserver
...
+            $Corcel = new Corcel();
+            $Corcel->post_title = $data['title'];
+            $Corcel->post_content = $data['description'];
+            $Corcel->post_type = config('hr.post-type.career');
+            $Corcel->save();
+            $Corcel->saveMeta('hr_id', $job['id']);
+        }
+    }
```
GauravGusain98, line 30: Let me rephrase my question, 
I was saying that suppose a job is not published but you updated the job details in the portal. In this case as well, the job details should get updated in the website post. But as per the code, it looks like that the job will only get updated when the status is published.

Also what about the case when we update the job from `published` to `draft`. I don't think in that case it will mark the job as a draft on the website. 

Let me know if there is any doubt in understanding.

GauravGusain98 (changes requested):
Please do the following changes.

am-Anmol:
**Changes in created and updated function:**
- Created:
    - Checking if the status is `not published` then saving the post as a draft.
- Updated:
   - Checking if the status is `published` then updating the post as a `publish`, otherwise updating as `draft` only.

Modules/HR/Observers/Recruitment/JobObserver.php
am-Anmol: Yeah, I have understood your scenario and updated the code as per all scenarios possible while creating and updating any post through an employee portal.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -16,5 +17,50 @@ class JobObserver
...
+        $post = Corcel::hasMeta('hr_id', $job['id'])->first();
+        $Corcel = Corcel::find($post->ID);
+        $Corcel->post_title = $data['title'];
+        $Corcel->post_content = $data['description'];
+        $Corcel->post_type = config('hr.post-type.career');
+        if ($data['status'] == 'published') {
+            $Corcel->post_status = 'publish';
+        } else {
+            $Corcel->post_status = 'draft';
+        }
```
GauravGusain98, line 52:
```suggestion
      $Corcel->post_status = $data['status'] == 'published' ? 'published' : 'draft';
```

GauravGusain98 (changes requested):
Please check the suggestion

Modules/HR/Observers/Recruitment/JobObserver.php
am-Anmol: Ok I make changes

GauravGusain98 (approved):
Looks good to me

Issue #893, opened 2021-08-16
Dashboard nothing is visible and there is JS error


rathorevaibhav:
**Describe the task**
The dashboard page sections are not loading at all. There is an error in browser console related to JS. I'm assuming this might be because of the linting fixes done.

**Dashboard is empty**
![image](https://user-images.githubusercontent.com/12053186/129527807-b18fa5de-423b-463d-9807-667fb3de7f9d.png)

**JS Error**
![image](https://user-images.githubusercontent.com/12053186/129527760-bc253bb9-b833-45cf-9ee2-843bf0cd0a67.png)

**Expected behavior**
- [ ] The dashboard components should show up
- [ ] There should be no com

PR #895, opened 2021-08-16
Nothing is visible on Dashboard (due to js-errors)

am-Anmol:
Targets #893 

### Description

This is supposed to be `default exported` but there using `new Vue` which means importing this component i.e., wrong, Instead, this import should be done in main.js.
These causing js errors and the dashboard isn't showing.
After these changes, the dashboard is visible and passing Js compilation successfully.

Dashboard:
image

Js:
image

### Checklist:


GauravGusain98 (approved):
Looks good to me.

Issue #836, opened 2021-03-05
User-specific maximum appointments per day


abhishek-pokhriyal:
**Is your feature request related to a problem? Please describe.**
The user should be able to specify the maximum number of appointments she would like to have.

**Describe the solution you'd like**
Right now the maximum number of appointments a user can have in a day is 3. This is value is defined in the `config` and is constant among all the users.

**Solution**
1. Create a new table `user_meta`
2. A meta key called `max_appointments_per_day` will store the max number of appointments
3. Update the code here to first read from `user_meta` and then fall back to config value https://github.com/ColoredCow/portal/blob/3f5b1ce2dab23d7f93354a75f1cd447e7f2d8875/Modules/AppointmentSlots/Services/AppointmentSlotsService.php#L216

**Describe alternatives you've considered**
Calling the person and rescheduling!

**Additional context**
@rathorevaibhav shared that 3 appointments in a day sometimes becomes too much. So, it'd be nice if he has the ability to control these.

rathorevaibhav:
This works great. @srivastava-jyoti great work!

PR #904, opened 2021-08-21
Created User_Meta Table 

srivastava-jyoti:
Targets: [#836 ](https://github.com/ColoredCow/portal/issues/836)
The user should be able to specify the maximum number of appointments he/she would like to have.

Checklist:
- [x] Created User_Meta table.
- [x] Update the values in AppointmentSlots.php
- [x] Update Config.php


GauravGusain98:
@srivastava-jyoti 
From what I am understanding from the issue description is that we need to make the functionality to give flexibility to candidate to choose maximum number of slots per day. This will contain following points:
- we need to make an UI also where we ask candidate to enter number of slots. 
- The user_meta table should have user_id column as well which is a foreign key to users table. 
- For each candidate there will be an entry in user_meta table
- We can use a default value which can be used if there is no entry of the user in user_meta table
- We need to make functionality to Create and Update max slots as well. 

Please confirm the requirements once more.

srivastava-jyoti:
Yes, these are the requirements. I will modify it.

srivastava-jyoti:
@GauravGusain98 @abhishek-pokhriyal @ColoredCow/team-portal-reviewers 
I have some queries related to the issue.
1. Who is booking these slots?? If colored cow members are booking then, why do we need this feature of maximum appointments per day? Maybe we could book according to our own convenience and if extra is booked the alternate solution which is mentioned in the issue itself could be used.
2. If we are not booking it those who want to give the interview are doing it, then why do we need to update maximum appointments per day according to each user?? Let's say Vaibhav decided to take 2 interviews that day and slots are booked and lately somebody decided to keep it as 1. What about the 2nd one??

Please clarify my doubts.

database/migrations/2021_08_21_063543_create_user_meta_table.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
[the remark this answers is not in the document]
srivastava-jyoti: @GauravGusain98 Coding Standards and styleci/pr are not getting passed for database seeder.php. Can you please look for once where the problem is??

GauravGusain98:
@srivastava-jyoti Please fix all the failed checks

DatabaseSeeder.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
GauravGusain98, line 1: This file should not be in the root directory. Please add the changes in the correct file.

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
```diff
@@ -213,7 +213,7 @@ private function getUserFreeSlots($userId)
         });
 
         $datesToRemove = $reservedSlotsCount->filter(function ($value, $key) {
-            return $value >= config('hr.daily-appointment-slots.max-reserved-allowed', 3);
+            return $value >= config('hr.daily-appointment-slots.max-reserved-allowed', 'max-slots');
```
GauravGusain98, line 216: How will the $value which I think is an integer type be compared with the 'max-slots' which is string. 

You have to fetch the value of the max slot for the user from the database and then compare it with the value. 
For this, you need to define a relationship of user and user_meta inside `user.php` module. You can take reference from [Here](https://laravel.com/docs/8.x/eloquent-relationships#one-to-one).

After the relationship is made you can fetch the max slots like this way:
`$user->meta-> max_appointments_per_day`

Modules/HR/Config/config.php
```diff
@@ -124,7 +124,7 @@
     ],
     'daily-appointment-slots' => [
         'total' => 6,
-        'max-reserved-allowed' => 3,
+        'max-reserved-allowed' => 'max-slots',
```
GauravGusain98, line 127: The `max-reserved-allowed` with the value of `max-slots` is not making any sense. Please change the key name. 

Also, I think this is not even required now as we can directly fetch the max-slots from the database using the relationship. 

Make sure you make these changes everywhere `config('hr.daily-appointment-slots.max-reserved-allowed')` is used

GauravGusain98 (changes requested):
Please check the feedbacks

Modules/HR/Entities/MaximumSlot.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Modules\HR\Entities;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\DB;
+
+class MaximumSlot extends Model
```
GauravGusain98, line 8: This name should UserMeta

Modules/HR/Config/config.php
```diff
@@ -124,7 +124,6 @@
     ],
     'daily-appointment-slots' => [
         'total' => 6,
-        'max-reserved-allowed' => 3,
```
GauravGusain98, line 127: Let it be as it is in the config. i.e 
```
'max-reserved-allowed' => 3,
```

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
```diff
@@ -213,7 +214,9 @@ private function getUserFreeSlots($userId)
         });
 
         $datesToRemove = $reservedSlotsCount->filter(function ($value, $key) {
-            return $value >= config('hr.daily-appointment-slots.max-reserved-allowed', 3);
+            $userData = MaximumSlot::getUserData();
```
GauravGusain98, line 217: You don't have to use it like this. You need to create a relationship between the `users` table and the `user_meta` table in the UserMeta Entity. So if I want to access something from `user_meta` table then I could use this relationship to access user_meta of the current user, like this:
```
auth()->user()->meta-> max_appointments_per_day
```

The relationship concept is widely used throughout the application. You can take reference from any Entity present in the application for the relationships.

GauravGusain98: Also, if a user don't have the entry in the user_meta table, then by default you should use the value from the config.

GauravGusain98 (changes requested):
@srivastava-jyoti Please check the comments. Also let me confirm the use case fo this issue:

- So every user in the database will have user meta entry. Currently we are not using any UI based method for entering the value to the user_meta table. We can create a new issue for that, but for now we are just focusing on the backend part.
- In Laravel, if we want create a relation between two table (i.e concept of Normalization in DBMS). So we use [relationships concept of Laravel.](https://laravel.com/docs/8.x/eloquent-relationships). 
- Once the relationship is built, we can easily access user_meta table values from user table. 
- In our case we want to access `appointement_slots` value for a user. Assume, for a user with id 1 we have an entry in the user_meta table and for user with id 2 we don't have such entry. In this case we will fetch the value of max_appoinment_slot for first user from the user_meta table using its user_id. And for the second suer we will check if its user_meta has any value related to max_appointment slots. If not then we will be using the default value from config.

Let me know if there is any doubt.

Modules/HR/Entities/MaximumSlot.php
srivastava-jyoti: @GauravGusain98 I have understood that each appointment slot should be distinct for each user. And if the values are not present for that user we will be using config value.

GauravGusain98:
@srivastava-jyoti I gave a couple of feedbacks earlier, but I don't see any progress in them. Could you please check it once?

app/User.php
```diff
@@ -83,4 +84,9 @@ public function scopeFindByEmail($query, $email)
     {
         return $query->where('email', $email);
     }
+
+    public function maxslot()
+    {
+        return $this->hasMany(UserMeta::class, 'user_id', 'id');
```
GauravGusain98, line 90: This relationship is incorrect. It should be hasOne relationship. As one user will only have one entry in the user_meta table

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
```diff
@@ -213,7 +214,9 @@ private function getUserFreeSlots($userId)
         });
 
         $datesToRemove = $reservedSlotsCount->filter(function ($value, $key) {
-            return $value >= config('hr.daily-appointment-slots.max-reserved-allowed', 3);
+            $userData = UserMeta::getUserData();
+
+            return $value >= $userData;
```
GauravGusain98, line 219: If user don't have entry in user_meta table then you should use default value form config

GauravGusain98 (changes requested):
Please work on these comments as well.

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
srivastava-jyoti: Okay, I will correct this.

app/User.php
srivastava-jyoti: This means I need to update records in user_meta table such that there will be only one entry for each user.

srivastava-jyoti:
@ Gaurav I have established hasOne relationship between tables but when I am trying to fetch the value from database using this query:`$max_slot= DB::table('maxslots')
            ->join('users', 'users.id', 'maxslots.user_id')
            ->select('max_interviews_per_day')
            ->where('maxslots.user_id','=', Auth::user('id'))
            ->latest('maxslots.created_at', 'desc')
            ->first();` 

I am able to book only one  slot for that particular day. Please guide me what I am doing wrong.

GauravGusain98:
> @GauravGusain98 I have established hasOne relationship between tables but when I am trying to fetch the value from database using this query:
> `$max_slot= DB::table('maxslots') ->join('users', 'users.id', 'maxslots.user_id') ->select('max_interviews_per_day') ->where('maxslots.user_id','=', Auth::user('id')) ->latest('maxslots.created_at', 'desc') ->first();`
> 
> I am able to book only one slot for that particular day. Please guide me what I am doing wrong.

@srivastava-jyoti I think the doubts are clear from the discussion we had.

srivastava-jyoti:
This is not required. I have integrated the entire part with #941

Issue #921, opened 2021-08-30
Not getting payment information fields for International invoices.


mohitsharma-22:
**Describe the bug**
We select paid from the drop-down to save the payment details received for invoices.
I am not able to get the "Payment Information:" details for international invoices. Seems to be a Bug. Please check and fix it.

**To Reproduce**
1.  Go to the Employee portal.
2. Click the invoices link from Finance dropdown.
3. Click on any international invoice.
4. Select paid from the Status drop down.
5. In the Right panel Payment Information: section should appear to fill up the details on the payment received against the invoice.
6. Payment Information: section is not showing up for international invoices. It is working fine for Indian invoices.

**Screenshots**
<img width="700" alt="Screenshot 2021-08-30 at 3 09 03 PM" src="https://user-images.githubusercontent.com/16541548/131321236-a071f5aa-2159-4941-8c18-58b74dca950f.png">

rathorevaibhav:
Might be a side-effect of changes done in this PR #852

rathorevaibhav:
This is deployed on UAT. @kickb71 please test.

kickb71:
> This is deployed on UAT. @kickb71 please test on UAT. I'll then merge to production.

Sure sir let me check @rathorevaibhav

kickb71:
@rathorevaibhav It works perfectly on UAT.

rathorevaibhav:
@mohitsharma-22 this is deployed on Production. Please check.

mohitsharma-22:
@rathorevaibhav  I am able to see the payment information section now and am able to add the details. 

Only one issue, the value of *Conversion Rate Diff:* is not getting saved now. It gets removed while saving the invoice. Can you please check?

<img width="500" alt="Screenshot 2021-08-31 at 10 24 02 AM" src="https://user-images.githubusercontent.com/16541548/131444165-760ad62e-29d1-4fcd-9af7-7f4ce70cf6c0.png">

kickb71:
@mohitsharma-22 The issue is solved & made PR

mohitsharma-22:
@rathorevaibhav It is working fine.

PR #922, opened 2021-08-30
Conversion difference is solved

kickb71:
This is an bug issue where the conversion to USD is not working which is rectified 

the Issue is https://github.com/ColoredCow/portal/issues/921


P4NK4J (approved):
Looks good.

Issue #316, opened 2018-05-29
New fields for job


rathorevaibhav:
**Description**
A job application may get closed after a particular date. There can be many reasons like no more vacancies or we're no longer hiring for that particular job description.
Right now, there is no way to hide that job field.

**Expected behavior**
When creating or editing a new job opportunity field, it should have two new date fields: `start_date` and `end_date`.

Add two new fields in the screen below
![image](https://user-images.githubusercontent.com/12053186/128664290-64797d37-8b28-4bbe-bf77-c43dd1e89628.png)

srivastava-jyoti:
@rathorevaibhav Okay, I will take help from @am-Anmol for setting up the database configurations.
@kuldeep3 I have removed > in start_date and made the pull request and working on the first feature which you asked for.

srivastava-jyoti:
@kuldeep3 This is how the page looks now.
![Screenshot (54)](https://user-images.githubusercontent.com/79009032/131243504-698efdf7-382f-4f3c-ba3b-0bdde3af33f8.png)

rathorevaibhav:
@kickb71 please test this also.

kickb71:
@rathorevaibhav I have tested it, I can see the date field, If I go into the job, I can see the start date and end date as well. 
But I can't edit the date it the start date is less than today's date.

rathorevaibhav:
This is working fine. The dates are saving. I'll open another issue to show/hide from website

PR #924, opened 2021-08-31
Added javascript code for hiding previous dates

srivastava-jyoti:
Issue:[#316](https://github.com/ColoredCow/portal/issues/316)

Tasks:

- [x] To remove > from start date in edit file
- [x] To hide previous dates

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -1002,20 +1002,12 @@ function datePickerChart(){
 	 });
 }
 
-var date = new Date();
-var tdate = date.getDate();
-if (tdate < 10) {
-	tdate = "0" + tdate;
+if ( document.getElementById("start_date") && document.getElementById("end_date") ){
+	var today = new Date().toISOString().split('T')[0];
+    	document.getElementsByName("start_date")[0].setAttribute("min", today);
+	document.getElementsByName("end_date")[0].setAttribute("min", today);
 }
-var month = date.getMonth() + 1;
-if (month < 10) {
-	month = "0" + month;
-}
-var year = date.getUTCFullYear();
-var minDate = year + "-" + month + "-" + tdate;
-document.getElementById("start_date").setAttribute("min", minDate);
-document.getElementById("end_date").setAttribute("min", minDate);
- 
+
 function lineChart(){
 	 var cData = value;
 	 var ctx = $("#line-chart");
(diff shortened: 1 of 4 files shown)


srivastava-jyoti:
@GauravGusain98 I have made a new PR for this issue. As in the previous one somehow other files got merged and I tried reverting but I was unable to revert.

For the issue which You pointed out that the changes are not visible on your local so for this we need to run `npm run watch`
in command prompt. I was facing same issue got checked on stackoverflow. Here's the link:[link](https://stackoverflow.com/questions/53425186/laravel-reactjs-changes-are-not-reflecting-after-change-in-my-jsreactjs-file)

Here's the screenshot from my local
Screenshot (57)

resources/js/app.js
```diff
@@ -1001,6 +1001,20 @@ function datePickerChart(){
...
+	tdate = "0" + tdate;
+}
+var month = date.getMonth() + 1;
+if (month < 10) {
+	month = "0" + month;
+}
+var year = date.getUTCFullYear();
+var minDate = year + "-" + month + "-" + tdate;
+document.getElementById("start_date").setAttribute("min", minDate);
+document.getElementById("end_date").setAttribute("min", minDate);
```
GauravGusain98, line 1017: Can you add a check here for the page where you need to load this script. You just need to add an id on the page and check the length of the id like this:

```suggestion
if ($('#page_id').length) {
    var date = new Date();
    var tdate = date.getDate();
    if (tdate < 10) {
                tdate = "0" + date;
    }
    var month = date.getMonth() + 1;
    if (month < 10) {
        	month = "0" + month;
    }
    var year = date.getUTCFullYear();
    var minDate = year + "-" + month + "-" + date;
    document.getElementById("start_date").setAttribute("min", minDate);
    document.getElementById("end_date").setAttribute("min", minDate);
}
```

GauravGusain98 (changes requested):
Please do the suggested changes

resources/js/app.js
srivastava-jyoti: @GauravGusain98 I have made the required changes. And for the id part, I have used a unique id for start and end dates.These are not generic for the entire hr module. The previous dates are not visible for job description pages only.

GauravGusain98 (approved):
Looks good to me

PR #928, opened 2021-09-01
Added opportunity show up in website category list

am-Anmol:
Targets #794 

### Description

- Upgraded `create` and `update` functionality which uses `wp_Terms` and `wp_TermRelationship` to show up in category using domain entered in portal.
- Corrected config post_type into lower-case because it is case-sensitive there on the website.

### Checklist:


P4NK4J (approved):
Looks good to me

Issue #802, opened 2021-07-12
Capturing reasons when rejecting a job application


rathorevaibhav:
**Describe the task**
When rejecting an application, we need to capture reasons as well.
Check designs from the links below:
1. Prototype: https://www.figma.com/proto/ZjX8ogSM5yrZGhvO5vEWJn/Employee-Portal?page-id=2%3A3&node-id=194%3A5770&viewport=369%2C-133%2C0.20342423021793365&scaling=min-zoom
2. Design file link: https://www.figma.com/file/ZjX8ogSM5yrZGhvO5vEWJn/Employee-Portal?node-id=2%3A3

**Expected behavior**
- [x] Complete the feature of Rejection
- [x] Create  Database table
- [ ] Use the Database table to store the values

**References**
Check out ColoredCow/old-portal#164 for more details

rathorevaibhav:
DB schema can look like this:

`hr_application_rejection_reasons` table
- hr_application_round_id (foreign key)
- reason_title (string)
- reason_comment (text)

An application round that got rejected can have more than one entry in the above table. For each row in the table:
1. round id signifies the id from `hr_application_rounds` table
2. title signifies the reason that was checked
3. comment signifies the comment for that checked reason. This field can be nullable.

rathorevaibhav:
@kickb71 receiving this error on UAT. Please check.
![image](https://user-images.githubusercontent.com/12053186/132954638-96a38911-99a0-4054-a79e-e89054839b72.png)

rathorevaibhav:
This works great.

PR #942, opened 2021-09-06
Reject

kickb71:
This Part Pull request solves the Feature part, Rejection Reasons part and Creation of DB table.

The issue is #802 
 
**Expected behavior**
- [x] Complete the feature of Rejection
- [x] Create  Database table
- [x] Use the Database table to store the values

Note: All the single quotes in the JS file changes to double quotes and the spaces because of the formatter


database/migrations/2021_09_06_081550_hr_application_rejection_reasons.php
```diff
@@ -0,0 +1,35 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('hr_application_rejection_reasons', function (Blueprint $table) {
+            $table->increments('id');
+            $table->integer('hr_application_round_id')->unsigned();
+            $table->foreign('hr_application_round_id')->references('id')->on('hr_application_round');
+            $table->string('reason_title', 255);
+            $table->text('reason_comment');
```
GauravGusain98, line 21: I think this should be nullable. Just check if we can leave the comment field empty or not.

resources/js/app.js
```diff
@@ -9,24 +9,24 @@ require("./bootstrap");
 import "jquery-ui/ui/widgets/datepicker.js";
```
GauravGusain98, line 9: It is very difficult to find the changes you have made with these many formatting changes. So only format the code that you have added.

resources/views/hr/application/rejection-modal.blade.php
```diff
@@ -41,6 +41,28 @@
                     <span aria-hidden="true">&times;</span>
                 </button>
             </div>
+                <div class="form-check">
+                    <form action="#" class="textbox">
+                        <h5><b> Select Reasons: </b></h5>
+                        <input type="checkbox" id="rejectReason1"> No responce <br>
```
GauravGusain98, line 47: Whenever you are using a text inside a blade tag, always sanitize it using `{{ __('') }}`

```suggestion
                        <input type="checkbox" id="rejectReason1"> {{ __('No response') }} <br>
```

resources/views/hr/application/rejection-modal.blade.php
```diff
@@ -41,6 +41,28 @@
...
+                        <label for="Rejection">
+                            <input type="checkbox" id="rejectReason2" /> Skills mismatch
+                        </label>
+                        <br />
+                        <div id="txtReasonRejected1" style="display: none">
+                            <input type="text" id="Reasons1" />
+                        </div>
+                        <input type="checkbox" id="rejectReason3"> Culture mismatch <br>
+                        <input type="checkbox" id="rejectReason4"> Salary  expectation mismatch <br>
```
GauravGusain98, line 56:
```suggestion
                        <input type="checkbox" id="rejectReason4"> {{ __('Salary expectation mismatch') }} <br>
```

resources/views/hr/application/rejection-modal.blade.php
```diff
@@ -41,6 +41,28 @@
...
+                        <label for="Rejection">
+                            <input type="checkbox" id="rejectReason5" /> Not enough knowledge/inclination for ColoredCow
+                        </label>
+                        <br />
+                        <div id="txtReasonRejected2" style="display: none">
```
GauravGusain98, line 61:
```suggestion
                        <div id="txtReasonRejected2" class="d-none">
```

resources/views/layouts/app.blade.php
```diff
@@ -21,6 +21,9 @@
     <script type="text/javascript" src="{{ URL::asset('js/app.js') }}"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.8.0/Chart.js"></script>
 
+    <!-- JQuery -->
+    <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.8.3/jquery.min.js"></script>
```
GauravGusain98, line 25: Why you have added this?

GauravGusain98 (changes requested):
Please check the comments.

database/migrations/2021_09_06_081550_hr_application_rejection_reasons.php
kickb71: It will automatically add text when you click check box that's why its not nullable

resources/js/app.js
kickb71: I did only use the formater app it made changes every where

resources/views/hr/application/rejection-modal.blade.php
kickb71: Sure let me change it  for all

resources/views/layouts/app.blade.php
kickb71: > Why you have added this?

I have added this Jquery for when a check box is clicked a textbox should appear that's why I need jquery

resources/js/app.js
GauravGusain98: So when you format a document. There is option to format only changes. So just use that option. You can find it in settings.

resources/views/layouts/app.blade.php
GauravGusain98: But we are using jquery earlier in app.js and it was working. Did you tried testing the code without importing this line. I would suggest testing it once without this line.

resources/js/app.js
kickb71: I will change it.  It won't happen again!

Modules/HR/Config/config.php
```diff
@@ -127,4 +127,11 @@
         'max-reserved-allowed' => 3,
     ],
     'offer-letters-dir' => 'offer-letters',
+    'reasons-for-rejections' => [
+        'No response',
+        'Skills mismatch',
+        'Culture mismatch',
+        'Salary expectation mismatch',
+        'Not enough knowledge/inclination for ColoredCow'
```
P4NK4J, line 135: @kickb71 I think we should keep key=>value pair in the config, and store the slug in the database, since the labels may change in the future. Let me know your thoughts on it.

Modules/HR/Http/Requests/Recruitment/ApplicationRoundRequest.php
```diff
@@ -51,6 +51,7 @@ public function rules()
                 $rules['mail_to_applicant.confirm.body'] = 'nullable|string|required_with:send_mail_to_applicant';
             }
         } elseif (request()->input('action') == 'reject') {
+            $rules['reject_reason'] = 'nullable|array';
```
P4NK4J, line 54: @kickb71 Any reason why nullable rule is being used here ?

Modules/HR/Routes/web.php
```diff
@@ -65,7 +65,8 @@
             Route::post('/applicant', 'ApplicantController@store')->name('hr.applicant.store');
             Route::post('/excel-import', 'ApplicantController@importExcel')->name('hr.applications.excel-import');
 
-            Route::resource('applications/rounds', 'ApplicationRoundController')->only(['store', 'update']);
+            Route::resource('applications/rounds', 'ApplicationRoundController')->only(['store', 'update', 'storeReason']);
+            // Route::post('/applicationround/{applicationRound}/store', 'ApplicationRoundController@storeReason')->name('reasons.post');
```
P4NK4J, line 69: @kickb71 Any specific reason, why the commented code is left in this file ?

database/migrations/2021_09_06_081550_hr_application_rejection_reasons.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class HrApplicationRejectionReasons extends Migration
```
P4NK4J, line 7: Why `Hr` naming convention is used here while at all other places `HR` is mentioned ? Let's use similar conventions everywhere.

resources/views/hr/application/rejection-modal.blade.php
```diff
@@ -41,6 +41,18 @@
                     <span aria-hidden="true">&times;</span>
                 </button>
             </div>
+            <div class="form-check mt-3">
+                <p class="font-weight-bold">Select Reasons</p>
+                @foreach (config('hr.reasons-for-rejections') as $index => $reasonForRejection)
+                    <div class="rejection-reason-block">
+                        <label for="reasonTitle{{ $index }}">
+                            <input type="checkbox" class="reject-reason mr-1" id="reasonTitle{{ $index }}" name='reject_reason[{{ $index }}][title]' value='{{ $reasonForRejection }}'>{{ $reasonForRejection }}<br>
```
P4NK4J, line 49: I think we should use slug and value pair here, since the vocabulary for reasons may change in the future, we could have constant slug saved in the database then. Let me know your thoughts,

P4NK4J (changes requested):
@kickb71 Left some comments, please take a look.

Modules/HR/Config/config.php
kickb71: If we change the key value means it will be hard. In for loop in the blade file and javascript file I have to change in both places. If we use a new value means we should just add here the new value but if we use the key-value means we have to change all the files for future use. 
@P4NK4J  I think it should be the same. If it should be changed means I happy to change it

Modules/HR/Routes/web.php
kickb71: I forgot to remove it. Let me remove it

Modules/HR/Http/Requests/Recruitment/ApplicationRoundRequest.php
kickb71: Yes. If you reject a person for only 2 reasons but all the others must not save that is why I gave nullable.

database/migrations/2021_09_06_081550_hr_application_rejection_reasons.php
kickb71: I can't understand what are trying to say?

resources/views/hr/application/rejection-modal.blade.php
kickb71: I think it will hard to pass in this foreach

Modules/HR/Config/config.php
P4NK4J: @kickb71 , I am saying the `key` will be saved in the database, which would never get changed. If there is any change to the reason let's say `No response` is changed to `No response given`, that would be easy to make only in the config, without changing it from the database, hence recommending to use the following
```suggestion
        'no-response' => 'No response',
        ...
```

Modules/HR/Http/Requests/Recruitment/ApplicationRoundRequest.php
P4NK4J: If the others are not selected from the form field, they won't be saved then.

Modules/HR/Routes/web.php
P4NK4J: I can see it hasn't been removed. Let me know if there is something which I am missing.

database/migrations/2021_09_06_081550_hr_application_rejection_reasons.php
P4NK4J: Let me know if it's clear after today's discussion

resources/views/hr/application/rejection-modal.blade.php
P4NK4J: Instead of $index, you use a slug in the value, and reason name in the label, while for the integer value you can initialize a different variable. Let me know if that makes sense.

P4NK4J (changes requested):
@kickb71  Left some comments, please check.

Modules/HR/Config/config.php
kickb71: The same thing happens without the key. I don't know the difference

Modules/HR/Http/Requests/Recruitment/ApplicationRoundRequest.php
kickb71: There are two things the title and comment box we can leave the comment box empty that's why its nullable

P4NK4J:
@kickb71 , Removing the status ready for review, I think there are some changes need to be made in this PR, as per our discussion.

database/migrations/2021_09_06_081550_hr_application_rejection_reasons.php
P4NK4J: @kickb71 Please resolve feedback before you assign for review

P4NK4J (commented):
Left one comment, please check @kickb71

resources/views/hr/application/rejection-modal.blade.php
P4NK4J: @kickb71 Also, you are not saving slugs in the value here, please make changes.

database/migrations/2021_09_06_081550_hr_application_rejection_reasons.php
kickb71: Changed!

resources/views/layouts/app.blade.php
```diff
@@ -14,13 +14,17 @@
 
     <!-- Styles -->
     <link href="{{ mix('css/app.css') }}" rel="stylesheet">
-    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet">
+    {{-- <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet"> --}}
```
P4NK4J, line 17: Why this CDN is commented @kickb71 ? Is it unused everywhere ?

P4NK4J (commented):
Left a comment, let me know your thoughts

resources/views/layouts/app.blade.php
kickb71: No i vas working on another issue that is why commented

GauravGusain98 (approved):
Looks good

P4NK4J (approved):

Issue #944, opened 2021-09-07
Integrate cypress for integration testing 


mdshamoon:
**Describe the task**
Add support for cypress so that manual effort in testing gets removed and we could easily catch bugs through automation.

**Expected behavior**
Cypress should be integrated with some simple tests like Login

rathorevaibhav:
This is done. Closing.

PR #945, opened 2021-09-07
Added cypress support

mdshamoon:
Targets #944 

### Description
This PR adds cypress for integration testing in this repository

### Checklist:


.github/workflows/integration-testing.yml
```diff
@@ -0,0 +1,67 @@
...
+          php artisan db:seed
+
+      - name: Compile assets
+        run: |
+          npm install
+          npm run dev
+
+      - name: Run local server
+        run: |
+          php artisan serve &
```
P4NK4J, line 61: What's the use of '&' in this command ?

cypress/integration/login/login.spec.js
```diff
@@ -0,0 +1,15 @@
+describe("Login", () => {
+  const username = Cypress.env("username");
```
P4NK4J, line 2: We also have G-Suite login in Employee portal, just checking if this should also be included in login specs ?

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
+// This example commands.js shows you how to
+// create various custom commands and overwrite
```
P4NK4J, line 3: How we will be using this file ? Just wanted to confirm, we are not checking in a commented code file

P4NK4J (commented):
Left some questions, please have a look.

cypress.json.example
```diff
@@ -0,0 +1,12 @@
+{
+  "baseUrl": "http://127.0.0.1:8000",
+  "$schema": "https://on.cypress.io/cypress.schema.json",
+  "env": {
+    "username": "user@coloredcow.com",
+    "password": "12345678"
```
GauravGusain98, line 6: Is it fine to add these credentials here? As it is an open-source repo. 
If this can create a security issue then we should not add the credentials here.

GauravGusain98 (commented):
Please check the comments.

.github/workflows/integration-testing.yml
mdshamoon: `&` helps us to run commands in the background.

The actions written here are performed once the previous one finishes. In this particular action `Run local server` since it opens up the local server, it never finishes up. `&` helps us to move forward while running this process in the background and not waiting for it to finish.

cypress/integration/login/login.spec.js
mdshamoon: I think let's start with this. We will surely need to add g-suite login in this spec, but I will target it as a separate task.

cypress/support/commands.js
mdshamoon: We use this file to write custom commands in cypress like login and logout which will be used in multiple test files. I thought let's keep it as it is for now and when we will come to it we can clean it and write commands since it has some examples as code that we can reference.

cypress.json.example
mdshamoon: Yes, I think since it is seed data and not used anywhere in staging and production so it should be fine. I can't think of any security issues that it can have. Please let me know if you have a particular case in mind.

.github/workflows/integration-testing.yml
P4NK4J: Okay, that makes sense

cypress.json.example
P4NK4J: @GauravGusain98 Let me know your thoughts on this, I think it is good to go.

P4NK4J (approved):
Looks good to me 👍

P4NK4J:
@GauravGusain98 , Left one comments for your thoughts.

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
...
+// existing commands.
+//
+// For more comprehensive examples of custom
```
mdshamoon, line 6: This file will be used for writing cypress commands for login, logout, and other custom commands in further development. So keeping this feel for now.

cypress.json.example
GauravGusain98: Cool.

Issue #917, opened 2021-08-27
Make website/wordpress database optional


rathorevaibhav:
**Describe the task**
The setup of website/WordPress database has become mandatory otherwise we can't create any job opportunity.
https://user-images.githubusercontent.com/79009032/131099978-523138a1-872a-45d8-ab8d-880ef07452b4.png

**Expected behavior**
- [ ] Instead of enforcing another DB setup, make syncing with website an optional field.
- [ ] Create a boolean `.env` variable `WORDPRESS_ENABLED` that must be checked wherever corcel is used. 
- [ ] The variable must be added to `.env.example` and default value should be false
- [ ] Update documentation about this variable and what it does

rathorevaibhav:
@am-Anmol @GauravGusain98 this is not working on UAT.

rathorevaibhav:
This is also working on UAT. Closing.

PR #961, opened 2021-09-11
Make website database optional

am-Anmol:
Targets #917 

### Description

- Created an env variable as `WORDPRESS_ENABLED` and default value is `false`.
- Updated database config with `WORDPRESS_ENABLED`
- Updated installation documentation for enabling website integration and use WordPress database.

### Checklist:


GauravGusain98 (changes requested):
Add a check in the `Modules/User/Services/UserService.php` Line 25 as well. Also, confirm if you have added checks at all the places.

GauravGusain98 (approved):
Looks Good

PR #962, opened 2021-09-11
910 Update Migrations for Effort Tracking

GauravGusain98:
Targets #970 

### Description

- Replace `resource` with `team-member` at all the places in codebase
- Add a migration to make the changes as per the [DB Schema](https://docs.google.com/presentation/d/1jaByFL_NodzD1onxb7JGie3MFsNxfrMbdsEg9WFHVDk/edit#slide=id.geb9fea0699_1_47)

### Checklist:


Modules/Prospect/Resources/views/index.blade.php
```diff
@@ -17,7 +17,7 @@
                 <tr>
                     <th>Name</th>
                     <th>Resources</th>
-                    <th>Requirnments</th
+                    <th>Requirements</th
```
rathorevaibhav, line 20: 👍

database/migrations/2021_09_09_082632_update_tables_for_effort_tracking.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('projects', function (Blueprint $table) {
+            $table->enum('type', array_keys(config('project.type')))->after('id')->default(config('project.type.monthly'));
```
rathorevaibhav, line 17: let's not use enums. They make it difficult later to rename columns in that table. This should be just `string` or `varchar`.

Modules/Project/Resources/views/subviews/edit-project-team-members.blade.php
```diff
@@ -0,0 +1,53 @@
...
+                    <div class="col-4">
+                        <select v-model="projectTeamMember.pivot.designation" :name="`projectTeamMember[${index}][designation]`" class="form-control">
+                            <option value="">Select Designations</option>
+
+                            <option v-for="(teamMemberDesignation, key) in teamMembersDesignations" :value="key">@{{ teamMembersDesignation }}</option>
+                        </select>
+                    </div>
+
+                    <div class="col-4">
+                        <button v-on:click="removeProjectTeamMember(index)" type="button" class="btn btn-danger btn-sm mt-1 ml-2"> - </button>
```
rathorevaibhav, line 37: can we do this? May look a bit nicer on UI
```suggestion
                        <button v-on:click="removeProjectTeamMember(index)" type="button" class="btn btn-danger btn-sm mt-1 ml-2 text-white fz-14">Remove</button>
```

Modules/Project/Config/config.php
```diff
@@ -6,10 +6,13 @@
...
     ],
-
-    'resource_designations' => [
+    'team_member_designations' => [
         'project_manager' => 'Project Manager',
         'developer' => 'Developer',
         'designer' => 'Designer',
+    ],
+    'type' => [
+        'monthly' => 'Monthly',
```
rathorevaibhav, line 15:
```suggestion
        'monthly-billing' => 'Monthly Billing',
```

rathorevaibhav (commented):
Some minor comments. @GauravGusain98 have a look.

Modules/Project/Config/config.php
GauravGusain98: Cool, let's update the `type` column value to `monthly-billing` in the database schema as well.

rathorevaibhav: Done.

kuldeep3:
@rathorevaibhav Will it be `team-members` or just `members` in place of `resources` as from team-members it seems we have multiple teams in the project and `project-team-members` reflects that for a project there are teams and members is confine within the team.

rathorevaibhav:
@kuldeep3 not sure if `project members` would be straightforward to understand. Let's keep it `project team members` for now only. I can understand where you're coming from with the doubt with teams inside project, but let's keep it this way right now.

kuldeep3 (changes requested):

kuldeep3 (approved):
Looks good on code ⚡

Issue #834, opened 2021-01-13
Option to add/remove tags


rathorevaibhav:
**Describe the task**
In the HR dropdown, add a new menu item that will take to manage tags.

**Expected behavior**
The manage tag functionality will have the following functionality:
1. Page to list all the tag
2. Add a new tag
3. Edit a tag
4. Delete a tag

**References**
See how GitHub has options to manage labels

rathorevaibhav:
@am-Anmol this gives an error on UAT
https://uat.employee.coloredcow.com/hr/tags

rathorevaibhav:
Also, the tags look like this now. I'm not sure but could be related to your fix. Can you check?
![image](https://user-images.githubusercontent.com/12053186/132950863-1605a5c7-a94f-4b97-9c4f-4d856cd2175b.png)

rathorevaibhav:
The tags also show up on the single application page but now it's not coming anymore
<img width="465" alt="Screenshot 2021-09-11 at 8 36 53 PM" src="https://user-images.githubusercontent.com/12053186/132952363-700aea9f-ddf6-4bd8-9172-f716e4a17dff.png">

rathorevaibhav:
This is working well.

PR #971, opened 2021-09-11
tag name rename to name

am-Anmol:
Targets #834 

### Description

- Rename migration of `tag_name` to `name` in `Tags` table.
- Renamed where we are using tag_name to name.

### Checklist:


P4NK4J (approved):
Looks good

PR #972, opened 2021-09-11
Added the solution that is causing the in UAT

kickb71:
Solved the error causing UAT not to reject an application.
 issue #802


P4NK4J (approved):
Looks good

PR #978, opened 2021-09-14
Fixed issues related to environment variables

GauravGusain98:
Targets #917 

### Description

- Fetching the environment variables from config instead of env file
- Corrected some coding conventions

### Checklist:

### DIFF
--- a/app/Listeners/RemoveUserFromWebsite.php
+++ b/app/Listeners/RemoveUserFromWebsite.php
@@ -30,15 +30,10 @@ class RemoveUserFromWebsite
         }
 
         $wpUser = WPUser::where('user_email', $event->user->email)->first();
-        if (! $wpUser) {
-            return;
-        }
-
         $admin = WPUser::where('user_login', config('website.admin_login_name'))->first();
-        if (! $wpUser) {
-            return;
-        }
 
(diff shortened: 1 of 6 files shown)


app/Listeners/RemoveUserFromWebsite.php
```diff
@@ -25,18 +25,20 @@ public function __construct()
...
-            }
+        $wpUser = WPUser::where('user_email', $event->user->email)->first();
+        if (! $wpUser) {
+            return;
+        }
 
-            WPPost::where('post_author', $wpUser->ID)->update(['post_author' => $admin->ID]);
+        $admin = WPUser::where('user_login', config('website.admin_login_name'))->first();
+        if (! $wpUser) {
+            return;
```
kuldeep3, line 39: Why this check is called two times?

```
if (! $wpUser) {
    return;
}
```

kuldeep3 (changes requested):

app/Listeners/RemoveUserFromWebsite.php
GauravGusain98: I think the second check should be for admin. Let me correct it.

kuldeep3 (approved):
Looks good on code ⚡

Issue #954, opened 2021-09-10
Remove the currency signs from excel downloaded from monthly tax reports.


mohitsharma-22:
**Describe the bug**
Signs are making these values text and to make a total, I need to remove the signs manually. 

<img width="913" alt="Screenshot 2021-09-10 at 7 22 51 PM" src="https://user-images.githubusercontent.com/16541548/132869754-c6bf4f5b-4006-4b5b-a2a7-62f7357400e3.png">

srivastava-jyoti:
@rathorevaibhav this is working fine. Currency symbols are not shown in downloaded excel sheet.

mohitsharma-22:
@rathorevaibhav it is working fine.
<img width="872" alt="Screenshot 2021-09-25 at 4 43 53 PM" src="https://user-images.githubusercontent.com/16541548/134769458-b0cafb8e-db3e-4af4-b523-4a9bb5c03739.png">
.

PR #980, opened 2021-09-14
The symbol is removed while downloading

gautam071:
Targets #954 

### Description
This PR contains the solution for the removal of currency signs from excel downloaded from monthly tax reports.

### Checklist:


GauravGusain98 (approved):
Looks good

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -205,9 +205,9 @@ private function formatInvoicesForExport($invoices)
         return $invoices->map(function ($invoice) {
             return [
                 'Project' => $invoice->project->name,
-                'Amount' => $invoice->display_amount,
+                'Amount' => (float) str_replace(['$', '₹'], '', $invoice->display_amount),
                 'GST' => $invoice->gst,
-                'Amount (+GST)' => $invoice->invoiceAmount(),
+                'Amount (+GST)' => (float) str_replace(['$', '₹'], '', $invoice->invoiceAmount()),
                 'Received amount' => $invoice->amount_paid,
```
kuldeep3, line 211: Can we access `$invoice->amount`, instead of firstly fetching `$invoice->display_amount` (that comes with currency symbol) and then removing the currency symbol.

kuldeep3 (changes requested):

Modules/Invoice/Services/InvoiceService.php
kickb71: ```InvoiceAmount()``` comes with the GST added so the value will change if add the ```$invoice->display_amount```

GauravGusain98: @kickb71 Can you check the logic of how `diplay_name` is calculated and how it is different from the `$invoice->amount`. If the value are same then you can use `$invoice->amount`. 
You can find the logic in this file `Modules/Invoice/Entities/Invoice.php`

GauravGusain98 (changes requested):

Modules/Invoice/Services/CurrencyService.php
```diff
@@ -42,6 +42,6 @@ private function fetchExchangeRateInINR()
 
         $data = json_decode($response->getBody()->getContents(), true);
 
-        return round($data['quotes']['USDINR'], 2);
+        return round($data, 2);
```
GauravGusain98, line 45: What are these changes for?

kickb71: If I go into in finance module I get the error that's why I removed a temp fix.I have created a issue for this

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -205,9 +205,9 @@ private function formatInvoicesForExport($invoices)
         return $invoices->map(function ($invoice) {
             return [
                 'Project' => $invoice->project->name,
-                'Amount' => $invoice->display_amount,
+                'Amount' =>  $invoice->amount,
```
GauravGusain98, line 208:
```suggestion
                'Amount' => $invoice->amount,
```

GauravGusain98 (changes requested):
Some Minor fixes. @kuldeep3 Could you please check this PR once?

GauravGusain98 (approved):
Looks good. Just remove the space on line 208.

gautam071:
> Looks good. Just remove the space on line 208.

@GauravGusain98 Can U merge this PR after the monthly tax PR.

GauravGusain98:
> > Looks good. Just remove the space on line 208.
> 
> @GauravGusain98 Can U merge this PR after the monthly tax PR.

Sure. Just remember to make the change I requested in this [comment](https://github.com/ColoredCow/portal/pull/980#pullrequestreview-758242193)

gautam071:
> > > Looks good. Just remove the space on line 208.
> > 
> > 
> > @GauravGusain98 Can U merge this PR after the monthly tax PR.
> 
> Sure. Just remember to make the change I requested in this [comment](https://github.com/ColoredCow/portal/pull/980#pullrequestreview-758242193)

Changed the indentation error

GauravGusain98:
@kuldeep3 Can you review this PR?

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -205,9 +205,9 @@ private function formatInvoicesForExport($invoices)
         return $invoices->map(function ($invoice) {
             return [
                 'Project' => $invoice->project->name,
-                'Amount' => $invoice->display_amount,
+                'Amount' => $invoice->amount,
                 'GST' => $invoice->gst,
-                'Amount (+GST)' => $invoice->invoiceAmount(),
+                'Amount (+GST)' => (float) str_replace(['$', '₹'], '', $invoice->invoiceAmount()),
```
kuldeep3, line 210:
```suggestion
                'Amount (+GST)' => $invoice->amount + $invoice->gst,
```

Can you once try this suggestion and compare the result.

kuldeep3 (changes requested):

Modules/Invoice/Services/InvoiceService.php
gautam071: Sure

gautam071: It works but it is used in multiple places should I changed it everywhere. But the code will be repeated

GauravGusain98: Just use it here only for now.

Modules/Invoice/Services/InvoiceService.php
GauravGusain98: Why this change showing again?

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -205,9 +205,9 @@ private function formatInvoicesForExport($invoices)
         return $invoices->map(function ($invoice) {
             return [
                 'Project' => $invoice->project->name,
-                'Amount' => $invoice->display_amount,
+                'Amount' =>  $invoice->amount,
                 'GST' => $invoice->gst,
-                'Amount (+GST)' => $invoice->invoiceAmount(),
+                'Amount (+GST)' => (float) str_replace(['$', '₹'], '', $invoice->amount + $invoice->gst),
```
GauravGusain98, line 210: I don't think Kuldeep told you to change it like this. Why you are using str_replace now? There will be no symbol now as you are just adding two numbers.

gautam071: because the method @kuldeep3 mentioned is coming with the rupee or roller symbols

GauravGusain98: I don't think ` $invoice->amount + $invoice->gst` will return symbol. As they both are numbers. Can you please check again

GauravGusain98: The whole point of using `$invoice->amount + $invoice->gst` was to remove the extra function i.e `str_replace` which is being used.

gautam071: because GST has a symbol i have tried that too then only I added the str replace

GauravGusain98: @kuldeep3 do you have any questions about it?

kuldeep3: @gautam071 So, are you saying that we are saving GST with a symbol on our database? If this is the case, I think we need to fix that too.

gautam071: Yeah, But the symbol need to be displayed in the webpage

kuldeep3: Ok, please revert it to the original one, i.e. ` (float) str_replace(['$', '₹'], '', $invoice->invoiceAmount()),`

GauravGusain98: @gautam071 Please make the changes Kuldeep has suggested.

kuldeep3 (approved):

Issue #1008, opened 2021-09-18
Google sheet integration and getting effort hours for each person in the project (JS snippet)


mdshamoon:
Completed:

- [x] Created a command `sync:effortsheet` that gets all the active projects and fill an entry for each user in the project in the table `project_team_member_effort`
- [x] Added the command to run each day at 10 pm with a CRON job

Left to do:

- [x] Calculate the effort for the current day. Following conditions needs to be checked for this:
    - Check if there is a last entry for the project member. If there is an entry:
        - Check if the month has changed or not
            - If the month has changed the actual hours will be the total hours we have in the effortsheet for the current month
            - If the month did not changed then calculate the effort for current entry by subtracting the total effort from previous entry
    - If not
        - Actual effort will be the total effort

GauravGusain98:
This is implemented. Closing this now.

PR #1009, opened 2021-09-18
Effortsheet integration

mdshamoon:
Targets #1008 

### Description

- This PR creates a cron job that syncs the effort sheet data with the database.
- Added one more field in the user's table ( nickname ) that matches the portal user with the user name in the google sheet.

**Note**: Need to add google client id in env to link google sheets 

### Checklist:


Modules/User/Database/Migrations/2021_09_16_134242_update_user_table_with_nickname.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
```
GauravGusain98, line 2: Can you create this migration file again? As there is some migrations that are newer than this in the UAT as per the timestamp in the migration name.

GauravGusain98 (changes requested):
Looks good. Just make the required changes. @P4NK4J  Could you please review this PR once?

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -0,0 +1,106 @@
...
+                }
+
+                $projectTeamMember = $portalUser->projectTeamMembers()->where('project_id', $project->id)->first();
+
+                if (! $projectTeamMember) {
+                    continue;
+                }
+
+                $latestProjectTeamMemberEffort = $projectTeamMember->projectTeamMemberEffort()->orderBy('added_on', 'DESC')->first();
+                $actual_effort = $user[4];
```
P4NK4J, line 85: Is there a possibility to have key-value pairs for the array, rather than having numeric indexes being used? Let me know your thoughts

P4NK4J (commented):
Left a question, else looks good 👍

Modules/User/Database/Migrations/2021_09_16_134242_update_user_table_with_nickname.php
mdshamoon: Updated.

Modules/Project/Console/SyncEffortsheet.php
mdshamoon: Unfortunately, the array structure is coming up from the google sheets library and there is no possibility of converting it to a key-value pair in this case.

P4NK4J:
Looks good to me. @GauravGusain98 you can also take a look and then we should be good to merge the PR.

P4NK4J (approved):

GauravGusain98 (approved):
Looks good

PR #1011, opened 2021-09-19
Hotfix: App slots fix

rathorevaibhav:
Targets #1012 

### Description
1. Creating appointment slots when count is less than 15
2. Renaming of functions and variables for better understanding
3. Comments for better understanding of the code

### Checklist:


GauravGusain98 (approved):
Looks good.

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
```diff
@@ -105,35 +113,40 @@ public function appointmentSelected($data)
...
 
         foreach ($slots as $slot) {
+            // skip for Saturday and Sunday
             if ($slot['start_time']->dayOfWeek == 0 || $slot['start_time']->dayOfWeek == 6) {
                 continue;
             }
+            // create appointment slot if doesn't exist for that time
             if (! AppointmentSlot::where('user_id', $userId)
                 ->where('start_time', $slot['start_time'])
                 ->exists()) {
```
GauravGusain98, line 152: We should take this value using a function, instead of directly adding this inside an if condition

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
```diff
@@ -105,35 +113,40 @@ public function appointmentSelected($data)
...
         $slots = [];
-        $numberOfSlots = config('hr.daily-appointment-slots.total', 6);
+        $maxSlotsDaily = config('hr.daily-appointment-slots.total', 6);
 
+        // run from today till 20 days
         for ($j = 0; $j < 20; $j++) {
             $nextDay = (clone $startDate)->addDays(1);
-            for ($i = 0; $i < $numberOfSlots; $i++) {
+            // for every day, create the appointment slots
+            for ($i = 0; $i < $maxSlotsDaily; $i++) {
```
GauravGusain98, line 128: We should give $i and $j a meaningful name

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
```diff
@@ -57,8 +57,16 @@ public function showAppointments($data, $params)
 
         $freeSlots = $this->getUserFreeSlots($userId);
 
-        if ($freeSlots->isEmpty()) {
-            $this->fillScheduleForTheDay($userId);
+        if ($freeSlots->count() < 15) {
```
GauravGusain98, line 60: We can take the value `15` form config.

GauravGusain98 (changes requested):
Some minor fixes

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
```diff
@@ -105,35 +113,40 @@ public function appointmentSelected($data)
...
+        // run from today till 20 days
         for ($j = 0; $j < 20; $j++) {
             $nextDay = (clone $startDate)->addDays(1);
-            for ($i = 0; $i < $numberOfSlots; $i++) {
+            // for every day, create the appointment slots
+            for ($i = 0; $i < $maxSlotsDaily; $i++) {
                 $slots[] = [
                     'user_id' => $userId,
                     'start_time' => (clone $startDate),
                     'end_time' => (clone $startDate)->addMinutes(30),
```
GauravGusain98, line 132: We can user interview duration from config

Modules/AppointmentSlots/Services/AppointmentSlotsService.php
rathorevaibhav: Didn't understand your point. It's a one-line Eloquent query. Why should we have a one-liner function?

GauravGusain98: Is it correct to write eloquent query inside if condition? If its fine then I think we can ignore this point.

rathorevaibhav: I think for one-liner true/false query it should be fine. If it's getting complicated, a variable would be preferred though instead of a function.

GauravGusain98: cool

GauravGusain98 (approved):
Looks good

PR #1013, opened 2021-09-20
Feature/added domain names/ same as main website

srivastava-jyoti:
Targets #{1001}

Description: Add all domains which are existing on the ColoredCow website career page.

Task:

- [ ] To add domains in config file.


srivastava-jyoti:
Previous changes were also added to this PR

Issue #1001, opened 2021-09-16
Update Domain list on the portal


am-Anmol:
**Describe the task**
Add all domains which are existing on the ColoredCow website career page.

**References**
- Portal:
![image](https://user-images.githubusercontent.com/48876126/133610924-a03474e0-4257-4879-b1b9-b0ee76adfcaa.png)
- ColoredCow Website:
![image](https://user-images.githubusercontent.com/48876126/133611001-4f94fee2-53f2-4608-8616-efe3ee9ec5da.png)

rathorevaibhav:
This looks great. Closing.

PR #1014, opened 2021-09-20
Added domain names to config file same as main website

srivastava-jyoti:
Targets #1001

Description: Add all domains which are existing on the ColoredCow website career page.

Task:

- [x] To add domains in config file.


GauravGusain98 (approved):
Looks good

Issue #998, opened 2021-09-16
The API for the conversion of USD to INR not found


gautam071:
**Is your feature request related to a problem? Please describe.**
The API keys are not found in the local when opening the finance page

**Describe the solution you'd like**
    - Make the API key
    - Error message should pop up saying "the API server fails to produce the values"

**Additional context**
<img width="1440" alt="Screenshot 2021-09-16 at 2 10 33 PM" src="https://user-images.githubusercontent.com/73119570/133580193-6b63ebea-f3a6-4fde-a87c-276e1532cfed.png">

**Package Link**
1. https://github.com/orkhanahmadov/laravel-currencylayer

**Expected Outcome**
- [ ] The connection with external API should be made only if API key is available in the env file. If it's absent, then system should not make the API call and instead use a hardcoded value from config.

rathorevaibhav:
This works great! Thanks @bharatsingh100 🎉

PR #1031, opened 2021-09-24
added check for currency-exchange-rate api key present in the env

bharatsingh100:
Targets #998 

### Description

It fixes the issue 998, added the check whether the api key is present in the config or env, if it's not present, then the function returns a default value (configurable from env) 
### Checklist:


P4NK4J (approved):
Looks good 👍

Issue #1020, opened 2021-09-21
Job opportunity start date and end date validation


rathorevaibhav:
**Describe the task**
When editing or creating a job opportunity, the validation are not right in the start date and end date. Right now we can have a start date greater than end date.
![image](https://user-images.githubusercontent.com/12053186/134132456-623bea53-1fb7-41b3-a228-7e892e5d1af9.png)

**Expected behavior**
- [x] Start date and end date should not be required fields. A job can have a start date but no end date and vice versa.
- [x] A user can select a start date and end date in the past given the below checks pass.
    - [x] If start date is selected first, the user should not be able to select end date before start date.
    - [x] If end date is selected first, the user should not be able to select start date after end date.

rathorevaibhav:
@srivastava-jyoti there's one issue here:

On the edit opportunity page, when you do nothing (after page load), it allows end date to select before start date and vice versa. Once I edit any of the two fields, it works as expected.
![image](https://user-images.githubusercontent.com/12053186/136662051-642684f2-dbc5-4f32-8b46-f57bc9a7a80a.png)

PR #1032, opened 2021-09-25
Bugfix nonvalidation for previous start_date in update form 

srivastava-jyoti:
Targets #{1019}

Checklist:

- [x]  The date validation should not be when there is already an existing date.

- [ ]  The validation should only work if the job opportunity didn't have a date before.

### DIFF
--- a/resources/views/hr/job/edit.blade.php
+++ b/resources/views/hr/job/edit.blade.php
@@ -15,7 +15,7 @@
     <br><br>
     @include('status', ['errors' => $errors->all()])
     <h2 class="mb-3">Edit Opportunity</h2>
-    <form action="{{ $formAction }}" method="POST" id="update-form" novalidate>
+    <form action="{{ $formAction }}" method="POST" id="update-form">
         @csrf
         @method('PATCH')
         <div class="card mb-3">
@@ -59,7 +59,7 @@
                     </div>
                     <div class="col-md-3 form-group">
                         <label for="start_date" class="fz-14 leading-none text-secondary mb-1">Start Date</label>
-                        <input type="date" class="form-control" id="job_start_date" name="start_date" value="{{ old('start_date', $job->start_date) }}">     
+                        <input type="date" class="form-control" id="jobs_start_date" name="start_date" min="{{ old('start_date', $job->start_date) }}" value="{{ old('start_date', $job->start_date) }}">     
                     </div>
                     <div class="col-md-3 form-group">
                         <label for="end_date" class="fz-14 leading-none text-secondary mb-1">End Date</label>
--- a/resources/views/hr/job/edit.blade.php
+++ b/resources/views/hr/job/edit.blade.php
@@ -15,7 +15,7 @@
     <br><br>
     @include('status', ['errors' => $errors->all()])
     <h2 class="mb-3">Edit Opportunity</h2>
-    <form action="{{ $formAction }}" method="POST" id="update-form">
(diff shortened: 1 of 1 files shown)


srivastava-jyoti:
@ColoredCow/team-portal-reviewers in this task I have used novalidate attribute in the form to disable javascript validation which was previously working for the previous start_date. So, there is no warning coming for start_date and the form is updating now without throwing any error. And it is still throwing errors for required parts.

For the 2nd part of this task, since the previous dates are already disabled. Do I need to enable form validation for this part??

resources/views/hr/job/edit.blade.php
```diff
@@ -15,7 +15,7 @@
     <br><br>
     @include('status', ['errors' => $errors->all()])
     <h2 class="mb-3">Edit Opportunity</h2>
-    <form action="{{ $formAction }}" method="POST" id="update-form">
+    <form action="{{ $formAction }}" method="POST" id="update-form" novalidate>
```
rathorevaibhav, line 18: As per my understanding from [this link](https://www.w3schools.com/tags/att_form_novalidate.asp), this attribute disables validation for the whole form. Can you double-check for other required fields like title & description if you remove the entries in these fields and try submitting the form, then the form should throw validation error.

srivastava-jyoti: @rathorevaibhav the required fields are still throwing errors when I try to submit the form.
Screenshot (112)

rathorevaibhav: @srivastava-jyoti no the frontend validation should work for them like before. The form shouldn't submit at all.

The screenshot you shared is because of backend validation that runs after you've submitted.

rathorevaibhav: It seems the novalidate attribute is disabling the frontend validation for the form.

srivastava-jyoti: Yeah, novalidate attribute is disabling frontend-validation. I will correct this.

P4NK4J (approved):
looks good.

P4NK4J:
@rathorevaibhav I see this PR has review requested from your end as well, please take a look and if everything looks fine, please proceed with merging it.

resources/views/hr/job/edit.blade.php
```diff
@@ -59,7 +59,7 @@
                     </div>
                     <div class="col-md-3 form-group">
                         <label for="start_date" class="fz-14 leading-none text-secondary mb-1">Start Date</label>
-                        <input type="date" class="form-control" id="job_start_date" name="start_date" value="{{ old('start_date', $job->start_date) }}">     
+                        <input type="date" class="form-control" id="jobs_start_date" name="start_date" min="{{ old('start_date', $job->start_date) }}" value="{{ old('start_date', $job->start_date) }}">
```
GauravGusain98, line 62: Why you changed the name of id. `jobs_start_date` is not making sense as it is in a plural form which is incorrect as per the case.

GauravGusain98: If id is not being used anywhere then you can remove it.

GauravGusain98 (changes requested):
Please check the comments

resources/views/hr/job/edit.blade.php
srivastava-jyoti: I am deleting this pr and will work on #1020

PR #1046, opened 2021-09-29
Corrected validation for start & end date

srivastava-jyoti:
Targets #{1020}

 Checklist:

- [x] Removed Unused class
- [x] Added min and max values to start & end dates

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -1081,11 +1081,12 @@ $("#job_start_date").on("change", function() {
  	let startDate = $("#job_start_date").val();
 	$("#job_end_date").attr("min", startDate);
 });
+
 $("#job_end_date").on("change", function(){
 	let endDate = $("#job_end_date").val();
 	$("#job_start_date").attr("max", endDate);
 });
-   
+
 /*
    * HR Module JS code end
    */
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -1005,11 +1005,6 @@ function datePickerChart(){
 	  });
 }
  
-if ( document.getElementById("job_start_date") && document.getElementById("job_end_date") ){
-	var today = new Date().toISOString().split("T")[0];
-	document.getElementsByName("start_date")[0].setAttribute("min", today);
-	document.getElementsByName("end_date")[0].setAttribute("min", today);
-}
 
 function lineChart(){
 	  var value = $(".chart-data").data("target");
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -1081,6 +1081,14 @@ $(function () {
 		}
 	});
 });
+$('#job_start_date').on('change', function() {
+ 	let startDate = $('#job_start_date').val();
+	$("#job_end_date").attr('min', startDate);
+})
+$('#job_end_date').on('change', function(){
+	let endDate = $('#job_end_date').val();
+	$("#job_start_date").attr('max', endDate);
+})
    
 /*
    * HR Module JS code end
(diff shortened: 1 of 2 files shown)


resources/views/hr/job/edit.blade.php
```diff
@@ -59,11 +59,11 @@
                     </div>
                     <div class="col-md-3 form-group">
                         <label for="start_date" class="fz-14 leading-none text-secondary mb-1">Start Date</label>
-                        <input type="date" class="form-control" id="job_start_date" name="start_date" value="{{ old('start_date', $job->start_date) }}">     
+                        <input type="date" class="form-control" min ="{{ old('start_date', $job->start_date) }}" max="{{ old('end_date', $job->end_date) }}" value="{{ old('start_date', $job->start_date) }}">     
                     </div>
                     <div class="col-md-3 form-group">
                         <label for="end_date" class="fz-14 leading-none text-secondary mb-1">End Date</label>
-                        <input type="date" class="form-control" id="job_end_date" name="end_date" value="{{ old('end_date', $job->end_date) }}">
+                        <input type="date" class="form-control" min ="{{ old('start_date', $job->start_date) }}" value="{{ old('end_date', $job->end_date) }}">
```
GauravGusain98, line 66: I don't think these changes cover all the requirements of the issue. For example the below case is not covered here:
- [ ] A user can select a start date and end date in the past given the below checks pass.
    - [ ] If start date is selected first, the user should not be able to select end date before start date.
    - [ ] If end date is selected first, the user should not be able to select start date after end date.

GauravGusain98: I think Javascript will be required to handle this case.

GauravGusain98 (changes requested):
@srivastava-jyoti Please check the comments and make the required changes.

resources/views/hr/job/edit.blade.php
srivastava-jyoti: Okay, understood I have to make this dynamic. Suppose if the user selects end date first, then start_date should not show dates after that. Same thing goes to start date. But, I have a small query what should be the minimum value of end date then??

rathorevaibhav: @srivastava-jyoti answering your questions:
1. There's no limit on the start date and end date when both are blank and any one of them is selected first. It can be a past date, today, or in the future. The validation should only be applied to the other field after one of these fields has been selected.
2. Add dynamic validations in both create and edit

srivastava-jyoti: @rathorevaibhav Okay, understood.

resources/js/app.js
```diff
@@ -1081,6 +1076,15 @@ $(function () {
...
 });
+
+$("#job_start_date").on("change", function() {
+ 	let startDate = $("#job_start_date").val();
+	$("#job_end_date").attr("min", startDate);
+});
+$("#job_end_date").on("change", function(){
+	let endDate = $("#job_end_date").val();
+	$("#job_start_date").attr("max", endDate);
+});
```
GauravGusain98, line 1087:
```suggestion

$("#job_start_date").on("change", function() {
 	let startDate = $("#job_start_date").val();
	$("#job_end_date").attr("min", startDate);
});

$("#job_end_date").on("change", function() {
	let endDate = $("#job_end_date").val();
	$("#job_start_date").attr("max", endDate);
});
```

GauravGusain98 (changes requested):
Looks good. Just some minor fixes required.

GauravGusain98 (approved):
Looks good

PR #1047, opened 2021-09-29
Corrected twice fields created for university alias and contact details

srivastava-jyoti:
Targets #{1042}

Checklist:

- [x] Corrected Space between icon and text written
- [x] Shifted University alias button to down
- [x] Removed twice fields created for university alias
- [x] Removed twice fields created for contact info

### DIFF
--- a/Modules/HR/Resources/views/layouts/master.blade.php
+++ b/Modules/HR/Resources/views/layouts/master.blade.php
@@ -5,7 +5,6 @@
 @endsection
 
 @section('js_scripts')
-	{{-- <script src="{{ mix('/js/hr.js') }}"></script> --}}
 	<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.1.0/dist/js/bootstrap.bundle.min.js" integrity="sha384-U1DAWAznBHeqEIlVSCgzq+c9gqGAJn5c/t99JyeKa9xxaYpSvHU5awsuZVVFIhvj" crossorigin="anonymous"></script>
 @endsection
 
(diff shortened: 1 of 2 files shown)


srivastava-jyoti:
In edit.blade.php on inspecting the page in script tag hr.js file is called twice so, when I have commented that it is called only since and the field is showing in once now.

Modules/HR/Resources/views/universities/edit.blade.php
```diff
@@ -52,6 +51,7 @@
         <div id="create_alias_form_list">
            {{-- For inserting dynamic add alias form --}}
         </div>
+        <button type="button" class="btn btn-outline-primary add-aliases mb-3" id="add_contact_btn"><i class="fa fa-plus"></i> Add University Aliases</button>
```
GauravGusain98, line 54: Is this change required Or it is done by mistake?

Modules/HR/Resources/views/layouts/master.blade.php
```diff
@@ -5,7 +5,7 @@
 @endsection
 
 @section('js_scripts')
-	<script src="{{ mix('/js/hr.js') }}"></script>
+	{{-- <script src="{{ mix('/js/hr.js') }}"></script> --}}
```
GauravGusain98, line 8: You can remove this line instead of commenting. Leaving commented code is not a good practice.

GauravGusain98 (changes requested):
Some comments. Please check.

Modules/HR/Resources/views/universities/edit.blade.php
GauravGusain98: Cool

GauravGusain98 (approved):
Looks good

Issue #932, opened 2021-09-02
Projects and Clients Seeders


rathorevaibhav:
**Describe the task**
Write seeders for projects and clients so that it's easy for new people to set up projects and clients easily.

**Expected behavior**
- [x] Seeder for project (resources, repositories and other meta should be filled)
- [x] Seeder for client (basic fields and other relations tables should be filled)

PR #1051, opened 2021-10-01
Project and Client seeder implemented

vipulgoyal006:
Targets #932

### Description
Implemented Seeders for Clients table and Projects table for easy use.

### Checklist:


database/Seeders/ClientsTableSeeder.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @return void
+     */
+    public function run()
+    {
+        \DB::table('clients')->delete();
+
+        \DB::table('clients')->insert([
+            0 => [
+                'id' => 1,
```
GauravGusain98, line 20: If we have already an id with value 1 then this will not work.

database/Seeders/ClientsTableSeeder.php
```diff
@@ -0,0 +1,55 @@
...
+                'name' => 'Rahul',
+                'status' => 'Not Active',
```
GauravGusain98, line 22: Please confirm if the value for the status column is `Not Active` or `not-active`.

database/Seeders/ClientsTableSeeder.php
```diff
@@ -0,0 +1,55 @@
...
+                'channel_partner_id' => null,
+                'parent_organisation_id' => null,
+                'country' => 'India',
+                'state' => null,
+                'phone' => null,
+                'address' => null,
+                'pincode' => null,
+            ],
+            1 => [
+                'id' => 2,
```
GauravGusain98, line 37: You should not add id column as it is added by default

database/Seeders/ClientsTableSeeder.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Database\Seeders;
+
+use Illuminate\Database\Seeder;
+
+class ClientsTableSeeder extends Seeder
```
GauravGusain98, line 7: You should create this seeder in the Client Module. Also, we have a factory for clients, So you can use the factory to seed the clients in the database. 

To know how factories and seeders work you can use laravel documentation. Also, the factory and project concept is implemented in the project seeder in the Project Module. You can take reference from there as well.

database/Seeders/ProjectsTableSeeder.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Database\Seeders;
+
+use Illuminate\Database\Seeder;
+
+class ProjectsTableSeeder extends Seeder
```
GauravGusain98, line 7: There is already a seeder and factory created for projects. I don't think it is required.

GauravGusain98 (changes requested):
Please check the comments

database/Seeders/ClientsTableSeeder.php
vipulgoyal006: The options were Active and Inactive. I have set 'Inactive' as deafult now.

database/Seeders/ClientsTableSeeder.php
vipulgoyal006: As suggested, I have removed id from the Clientfactory.

database/Seeders/ClientsTableSeeder.php
vipulgoyal006: Done!

database/Seeders/ProjectsTableSeeder.php
vipulgoyal006: Implemented only ClientsSeeder now

database/Seeders/ClientsTableSeeder.php
vipulgoyal006: Used a factory to implement this seeder

Modules/Client/Database/Factories/ClientFactory.php
```diff
@@ -23,7 +23,19 @@ public function definition()
     {
         return [
             'name' => $this->faker->name,
-            'status' => 'active',
+            'status' => 'Inactive',
```
GauravGusain98, line 26:
```suggestion
            'status' => 'inactive',
```

Modules/Client/Database/Factories/ClientFactory.php
```diff
@@ -23,7 +23,19 @@ public function definition()
...
+            'key_account_manager_id' => $this->faker->unique($reset = true)->randomDigitNotNull,
+            'created_at' => now(),
+            'updated_at' => now(),
+            'is_channel_partner' => rand(0, 1),
+            'has_departments' => rand(0, 1),
```
GauravGusain98, line 31: You can let them be 0.

Modules/Client/Database/Factories/ClientFactory.php
```diff
@@ -23,7 +23,19 @@ public function definition()
...
+            'channel_partner_id' => rand(1, 1000),
+            'parent_organisation_id' => rand(1, 1000),
+            'country' => 'India',
+            'state' => null,
+            'phone' => null,
+            'address' => $this->faker->address,
+            'pincode' => null,
```
GauravGusain98, line 38: These fields are nullable so you can just remove them. You can keep the address field though.

GauravGusain98 (changes requested):
Looks good. Just some minor fixes.

Modules/Client/Database/Factories/ClientFactory.php
vipulgoyal006: Implemented the suggestion!

Modules/Client/Database/Factories/ClientFactory.php
vipulgoyal006: Done

Modules/Client/Database/Factories/ClientFactory.php
vipulgoyal006: Done

Modules/Client/Database/Factories/ClientFactory.php
```diff
@@ -23,7 +23,13 @@ public function definition()
     {
         return [
             'name' => $this->faker->name,
-            'status' => 'active',
+            'status' => 'inactive',
+            'key_account_manager_id' => $this->faker->unique($reset = true)->randomDigitNotNull,
+            'created_at' => now(),
+            'updated_at' => now(),
+            'is_channel_partner' => '0',
+            'has_departments' => '0',
```
GauravGusain98, line 31: These 2 columns are of int type
```suggestion
            'is_channel_partner' => 0,
            'has_departments' => 0,
```

GauravGusain98 (changes requested):
Please check the comments

GauravGusain98 (approved):
looks good now.

Issue #1000, opened 2021-09-16
Create commands for login and logout in cypress 


mdshamoon:
**Describe the task**
Create commands for login and logout in cypress so that it can be reused in multiple test files.

**Expected behavior**
We should have two commands in cypress namely- `login` and `logout` that can be called in any test case and perform the relevant actions.

**References**
https://docs.cypress.io/api/cypress-api/custom-commands

am-Anmol:
I'll be covering this issue in this PR #999

PR #1064, opened 2021-10-05
Added login and logout custom commands

am-Anmol:
Targets #1000 

### Description
Created custom `login` and `logout` commands in cypress, so that it can be called in any test case and perform the relevant actions.
- References: https://docs.cypress.io/api/cypress-api/custom-commands

### Checklist:

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -23,3 +23,13 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
+Cypress.Commands.add('Login', (user) => {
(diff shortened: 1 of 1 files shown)


am-Anmol:
> @am-Anmol Can we also use these commands in the login.spec file so that we know that these commands are working fine or not?

Yeah, we can use it in login.spec file also. 👍🏻

mdshamoon:
> > @am-Anmol Can we also use these commands in the login.spec file so that we know that these commands are working fine or not?
> 
> Yeah, we can use it in login.spec file also. 👍🏻

If possible can we use these commands in login.spec file in this PR only.

Issue #1053, opened 2021-10-01
Tags used by the system should not be allowed to delete


rathorevaibhav:
**Describe the task**
Disable delete for the default tags that are hard-coded in the code. These following tags are hard-coded and user shouldn't be allowed to delete them.
![image](https://user-images.githubusercontent.com/12053186/135629025-9bcffc0b-b8ee-4c8f-84be-05608fabc35a.png)

**Expected behavior**
- [ ] The delete option for these 6 tags should be disabled and not clickable. 
- [ ] If you hover on the delete icon, there should be a tooltip that says "You cannot delete a tag used by the system"

PR #1093, opened 2021-10-13
Tags used by system not allowed not delete

am-Anmol:
Targets #1053 

### Description

- Added system default tags in `config`.
- Updated delete button with checking if it is default tag or not then showing different delete button.
- image

### Checklist:


am-Anmol:
> Looks good

@GauravGusain98 Can we merge this now?

Issue #1137, opened 2021-10-25
Tags used by the system should not be allowed to edit


rathorevaibhav:
**Describe the task**
Coming from #1053

Right now the user can edit and rename a tag and can delete it.

Disable edit for the default tags that are hard-coded in the code. The following tags are hard-coded and user shouldn't be allowed to edit them.
![image](https://user-images.githubusercontent.com/12053186/135629025-9bcffc0b-b8ee-4c8f-84be-05608fabc35a.png)

**Expected behavior**
- [ ] The edit option for these 6 tags should be disabled and not clickable. 
- [ ] If you hover on the edit icon, there should be a tooltip that says "You cannot edit a tag used by the system"

srivastava-jyoti:
This works fine. Closing

PR #1151, opened 2021-10-28
Tags used by the system should not be allowed to edit

am-Anmol:
Targets #1137 

### Description
Updated system tags so that we can't edit and show a tooltip with not allowed to delete.

### Checklist:


GauravGusain98 (approved):
Looks fine

Issue #1113, opened 2021-10-18
Save Applicant WhatsApp Optin Time 


pankaj-ag:
**Is your feature request related to a problem? Please describe.**
We are integrating WhatsApp services in EP. For that, we need to save the Applicant optin in the DB so that we can communicate with them. 

**Describe the solution you'd like**
1. Create a new migration and add a field wa_optin_at in the applicant table. The type of this column will be timestamp and by default nullable.
2. Add a new field in the new application form (in the portal) as well. 
![image](https://user-images.githubusercontent.com/10670690/137671785-19bed8ed-75a1-48ba-828c-0629f9f7ebc1.png)

srivastava-jyoti:
Closing

PR #1156, opened 2021-11-01
Added whatsapp optin field

srivastava-jyoti:
Targets #1113


database/migrations/2021_10_30_084041_update_hr_applicants_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('hr_applicants', function (Blueprint $table) {
+            $table->time('wa_optin_at')->after('phone')->nullable();
```
GauravGusain98, line 17: This should be timestamp instead of time. It is also given in the issue description.

Modules/HR/Entities/Applicant.php
```diff
@@ -29,6 +29,7 @@ public static function _create($attr)
         ], [
             'name' => $attr['name'],
             'phone' => isset($attr['phone']) ? $attr['phone'] : null,
+            'wa_optin_at' => isset($attr['wa_optin_at']) ? $attr['wa_optin_at'] : null,
```
GauravGusain98, line 32: you need to add the current date as well along with the time in this column.

GauravGusain98 (changes requested):
Please check the comments.

GauravGusain98:
@rathorevaibhav Can we provide the date and time field for the WhatsApp option as shown below? And can the date be other than the current data?
image

rathorevaibhav:
@GauravGusain98 
1. The task was to capture the timestamp of opt-in, no? I didn't get your question
2. Yes, it can be some other date as well (we may require it specially when doing bulk upload)

GauravGusain98:
> @GauravGusain98
> 
> 1. The task was to capture the timestamp of opt-in, no? I didn't get your question
> 2. Yes, it can be some other date as well (we may require it specially when doing bulk upload)

1. Initially there was only a time field, So I thought we only need a time field in UI. 
2. Got it

GauravGusain98 (approved):
Looks good on code.

Issue #1159, opened 2021-11-02
Estimated hours fields should be non required. 


GauravGusain98:
**Describe the bug**
On the Edit Project and Add Project page, the `total estimated hours` and `monthly estimated hours` are non-required fields but we are getting an error when we leave these fields empty.

TASKS: 
- [x] Make the above-mentioned fields nullable
- [x] Update the `monthly_estimated_hours` column as nullable in database table

**To Reproduce**
NA

**Expected behavior**
The user should be able to leave these fields empty.

**Screenshots**
![image](https://user-images.githubusercontent.com/54376313/139801739-ef0ead6a-629f-4e0c-9603-03f7342f6a45.png)

![image](https://user-images.githubusercontent.com/54376313/139802213-083c6e6c-8f3a-48d4-bcf3-d2c700f29940.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

srivastava-jyoti:
This works fine. Closing.

PR #1161, opened 2021-11-02
non-req-fields

suntwalriya:
Targets #1159 

### Description
On the Edit Project page and Create Project Page, the total estimated hours and monthly estimated hours are non-required fields and these fields can be now left empty.

### Checklist:

### DIFF
--- a/Modules/Project/Database/Migrations/2021_09_09_082632_update_tables_for_effort_tracking.php
+++ b/Modules/Project/Database/Migrations/2021_09_09_082632_update_tables_for_effort_tracking.php
@@ -16,7 +16,7 @@ class UpdateTablesForEffortTracking extends Migration
         Schema::table('projects', function (Blueprint $table) {
             $table->string('type')->after('id')->default(array_keys(config('project.type'))[0]);
             $table->float('total_estimated_hours')->nullable()->after('effort_sheet_url');
-            $table->float('monthly_estimated_hours')->nullable()->after('total_estimated_hours');
+            $table->float('monthly_estimated_hours')->after('total_estimated_hours');
         });
 
         Schema::table('project_resources', function (Blueprint $table) {
--- a/Modules/Project/Database/Migrations/2021_09_09_082632_update_tables_for_effort_tracking.php
+++ b/Modules/Project/Database/Migrations/2021_09_09_082632_update_tables_for_effort_tracking.php
@@ -16,7 +16,7 @@ class UpdateTablesForEffortTracking extends Migration
         Schema::table('projects', function (Blueprint $table) {
             $table->string('type')->after('id')->default(array_keys(config('project.type'))[0]);
             $table->float('total_estimated_hours')->nullable()->after('effort_sheet_url');
-            $table->float('monthly_estimated_hours')->after('total_estimated_hours');
+            $table->float('monthly_estimated_hours')->nullable()->after('total_estimated_hours');
         });
 
         Schema::table('project_resources', function (Blueprint $table) {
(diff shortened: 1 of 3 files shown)


GauravGusain98 (commented):
Looks good. But we also decided to update the Table structure as well. Without that this code will not work as expected.

suntwalriya:
@GauravGusain98 I did that in my local, and it is working fine on my local.

GauravGusain98:
> @GauravGusain98 I did that in my local, and it is working fine on my local.

But we also need to change the structure of the production database. For that, we need the migration file changes. Please add that file in this PR.

Modules/Project/Database/Migrations/2021_09_09_082632_update_tables_for_effort_tracking.php
```diff
@@ -16,7 +16,7 @@ public function up()
         Schema::table('projects', function (Blueprint $table) {
             $table->string('type')->after('id')->default(array_keys(config('project.type'))[0]);
             $table->float('total_estimated_hours')->nullable()->after('effort_sheet_url');
-            $table->float('monthly_estimated_hours')->after('total_estimated_hours');
+            $table->float('monthly_estimated_hours')->nullable()->after('total_estimated_hours');
```
GauravGusain98, line 19: We don't update an existing migration. If we want to change the table structure then we need to create a new migration for those changes. 

This is because the old migrations have already been executed and will not be executed again. So, if we update anything in the old migration then no changes will be reflected.

GauravGusain98 (changes requested):
Please check the comments.

Modules/Project/Database/Migrations/2021_09_09_082632_update_tables_for_effort_tracking.php
suntwalriya: Done!

Modules/Project/Database/Migrations/2021_11_05_105704_make_monthly_estimated_hours_nullable_on_projects_table.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::dropIfExists('monthly_estimated_hours_nullable_on_projects');
```
GauravGusain98, line 28: You should not drop this column on rollback. Instead, change the column to not nullable.

Modules/Project/Database/Migrations/2021_11_05_105704_make_monthly_estimated_hours_nullable_on_projects_table.php
```diff
@@ -0,0 +1,30 @@
...
+class MakeMonthlyEstimatedHoursNullableOnProjectsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('monthly_estimated_hours_nullable_on_projects', function (Blueprint $table) {
```
GauravGusain98, line 16:
```suggestion
        Schema::table('monthly_estimated_hours_nullable_on_projects', function (Blueprint $table) {
```

GauravGusain98 (changes requested):
Please check the comments

GauravGusain98:
As per the discussion in the meeting, Please update the code changes. Updating the label to `reviewed`.

suntwalriya:
I made the discussed changes. Please check.

GauravGusain98 (approved):
Looks fine

Issue #831, opened 2021-01-13
Option to assign/remove labels to an application


rathorevaibhav:
**Describe the task**
We need to tag applications with labels for easy sorting and finding.

**Expected behavior**
In the application listing screen (both job and internship), there should be an option to add labels to an application.

1. In every row, show a three-dot icon before the applicant name that will open a dropdown. The dropdown will only have one option right now which would be "Assign labels"
2. When clicking on assign labels, it should open a modal that will have the functionality to select/unselect labels as checkbox.
3. On clicking on the save option inside the modal, the page reloads with updated labels for that application.

**References**
Similar to how GitHub allows option to assign/remove labels for an issue/PR.

srivastava-jyoti:
Closing.

PR #1191, opened 2021-11-25
Feature/assign/remove labels

srivastava-jyoti:
Targets #831


Modules/HR/Resources/views/application/render-application-row.blade.php
```diff
@@ -4,6 +4,7 @@
             <a href="/{{ Request::path() }}/{{ $application->id }}/edit"
                 class="mr-1">{{ $application->applicant->name }}
             </a>
+            <button style="outline: none;border: none; background-color: transparent;" title="Assign labels" data-toggle="modal" data-target="#assignlabelsmodal" type="button">{!! file_get_contents(public_path('icons/three-dots-vertical.svg')) !!}</button>
```
P4NK4J, line 7: Can we put the style in SCSS files instead of using inline ?

P4NK4J (commented):
Left a comment, please check.

Modules/HR/Resources/views/application/render-application-row.blade.php
srivastava-jyoti: Okay, will correct it.

resources/sass/app.scss
```diff
@@ -771,6 +771,12 @@ span.total-chart-count{
 .exclamation-icon{
 	color: $theme-red;
 }
+// assigning labels
+.assignlabels{
+	border: none;
+	outline: none; 
+	background-color: transparent;
+}
 // End Custom
```
P4NK4J, line 780: Do we need the comments here? I think it should be obvious with the class name. Also if you could use the Bootstrap classes instead of defining custom CSS, that would be great.

P4NK4J (commented):
Minor comment, let me know your thoughts.

resources/sass/app.scss
srivastava-jyoti: Removed comment. There is no bootstarp class for `border: none` and `background-color: transparent`. So, I have written it explictly.

P4NK4J (approved):
Looks good.

PR #1198, opened 2021-11-28
Effortsheet cubejs

AkhileshNegi:
Cube_JS integration to show project monthly billable hours


resources/views/knowledgecafe/library/books/show.blade.php
```diff
@@ -95,6 +95,9 @@
         :user = "{{ auth()->user() }}"
         />
     </div>
+    <div class="mt-3 w-75">
+        <effort-component/>
```
P4NK4J, line 99: @AkhileshNegi Is there any specific use case of using the KnowledgeCafe module in effortsheet integration with Cube.Js? Let me know if I am missing something.

P4NK4J (commented):
Left a comment, please take a look.

resources/views/knowledgecafe/library/books/show.blade.php
AkhileshNegi: @P4NK4J No there isn't. It's a miss from my end, I have reverted changes to it. Thanks

GauravGusain98 (approved):
Looks fine

PR #1224, opened 2021-12-16
Corrected Permissions Table seeding

srivastava-jyoti:
Issue: On running ` php artisan db:seed --class=PermissionsTableSeeder` previous values were again seeded in the database.

### DIFF
--- a/database/Seeders/PermissionsTableSeeder.php
+++ b/database/Seeders/PermissionsTableSeeder.php
@@ -16,7 +16,7 @@ class PermissionsTableSeeder extends Seeder
      */
     public function run()
     {
-        \DB::table('permissions')->delete();
+        DB::table('permissions');
         Artisan::call('permission:cache-reset');
 
         Permission::updateOrCreate(['name' => 'finance_reports.view']);
--- a/database/Seeders/PermissionsTableSeeder.php
+++ b/database/Seeders/PermissionsTableSeeder.php
@@ -5,6 +5,7 @@ namespace Database\Seeders;
 use Illuminate\Database\Seeder;
 use Illuminate\Support\Facades\Artisan;
 use Spatie\Permission\Models\Permission;
+use DB;
 
 class PermissionsTableSeeder extends Seeder
 {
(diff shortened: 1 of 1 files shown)


database/Seeders/PermissionsTableSeeder.php
```diff
@@ -15,8 +16,10 @@ class PermissionsTableSeeder extends Seeder
      */
     public function run()
     {
-        \DB::table('permissions')->delete();
+        DB::statement('SET FOREIGN_KEY_CHECKS=0;');
+        \DB::table('permissions')->truncate();
```
GauravGusain98, line 20: I think let's not solve this issue in this manner. AS if we delete the permissions and again create new ones (the permission will have new ids) then the roles cannot find the permissions to which they are mapped as we have deleted those permissions and now the id to which roles are mapped does not exists.

For this we can use the `createOrUpdate()` function of laravel ORM. 
[Reference documentation](https://laravel.com/docs/8.x/eloquent#upserts)

GauravGusain98: Also then we have to remove the delete statements from the seeder.

GauravGusain98 (changes requested):
Please check the comments

database/Seeders/PermissionsTableSeeder.php
srivastava-jyoti: Okay, I'll correct this.

database/Seeders/PermissionsTableSeeder.php
```diff
@@ -18,104 +19,105 @@ public function run()
         \DB::table('permissions')->delete();
```
GauravGusain98, line 19: This is not required. Please remove this line

database/Seeders/PermissionsTableSeeder.php
```diff
@@ -18,104 +19,105 @@ public function run()
...
         Artisan::call('permission:cache-reset');
```
GauravGusain98, line 20: Same here. This is also not required.

GauravGusain98 (approved):
Looks good.

Issue #1234, opened 2021-12-18
Create a reports table to store report URL and other data


GauravGusain98:
**Is your feature request related to a problem? Please describe.**
We need to store the embedded URL of the reports that we will be showing on the sales and marketing page. For this we will be needing a table to store the report details.

**Describe the solution you'd like**
- [x] Create a table `reports` using the Schema provided in this [Link](https://docs.google.com/presentation/d/19p69Ic4yyBQKjauHgXdsBZo3PFmQV5crSuZZfIYVtBc/edit#slide=id.g107fd8ecc8f_0_75)

**Describe alternatives you've considered**
NA

**Additional context**
- https://laravel.com/docs/8.x/migrations

dsajwan2020:
The  table reports is created using  the Schema provided in this [Link] (https://docs.google.com/presentation/d/19p69Ic4yyBQKjauHgXdsBZo3PFmQV5crSuZZfIYVtBc/edit#slide=id.g107fd8ecc8f_0_75)

The table reports is available in the database refer screen shot

![image](https://user-images.githubusercontent.com/96309135/162724190-c500c849-f12b-4df1-bf2f-4a8eff58a0f8.png)

Ayush8923:
This issue PR is merged. Closing it.

PR #1242, opened 2022-01-14
1234 create a reports table to store report url and other data

Ayush8923:
Targets #1234 

### Checklist:


Modules/SalesAutomation/Database/Migrations/2021_12_27_175619_create_reports_table.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('reports', function (Blueprint $table) {
+            $table->id();
+            $table->string('name')->nullable();
```
GauravGusain98, line 18: This field is a required field

GauravGusain98 (changes requested):
Minor feedback, please fix it.

Modules/SalesAutomation/Database/Migrations/2021_12_27_175619_create_reports_table.php
Ayush8923: u can tell how can i show the required field.Can you give some ideas.
Because as far as I think we can write the requested field as Nullable (False).

Modules/SalesAutomation/Database/Migrations/2021_12_27_175619_create_reports_table.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('reports', function (Blueprint $table) {
+            $table->id();
+            $table->string('name')->nullable(false)->change();
```
GauravGusain98, line 18:
```suggestion
            $table->string('name');
```

Modules/SalesAutomation/Database/Migrations/2021_12_27_175619_create_reports_table.php
```diff
@@ -0,0 +1,35 @@
...
+            $table->text('description')->nullable();
+            $table->string('url')->nullable(false)->change();
+            $table->string('type')->nullable(false)->change();
```
GauravGusain98, line 21:
```suggestion
            $table->string('url');
            $table->string('type');
```

Modules/SalesAutomation/Database/Migrations/2021_12_27_175619_create_reports_table.php
GauravGusain98: By default the field is `required`

GauravGusain98 (changes requested):
Please check the comments

Modules/SalesAutomation/Database/Migrations/2021_12_27_175619_create_reports_table.php
Ayush8923: @GauravGusain98, I want to ask you that whatever the URL and type is, as per the database schema is also a required field, so they have to provide the required field as well.

GauravGusain98: If you remove the nullable(), then the field will be consider as required by default

Modules/SalesAutomation/Database/Migrations/2021_12_27_175619_create_reports_table.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('reports', function (Blueprint $table) {
+            $table->id();
+            $table->string('name')->default(false);
```
GauravGusain98, line 18: First thing, the default value of the string cannot be false. Also In the schema default value is not mentioned anywhere so please don't add default values.

Modules/SalesAutomation/Database/Migrations/2021_12_27_175619_create_reports_table.php
```diff
@@ -0,0 +1,35 @@
...
+            $table->text('description')->nullable();
+            $table->string('url')->default(false);
+            $table->string('type')->default(false);
```
GauravGusain98, line 21: same here

GauravGusain98 (changes requested):
Please check the comments

GauravGusain98 (approved):
looks fine now

Issue #1195, opened 2021-11-27
Phase 2: Project details page


GauravGusain98:
**Issue's Linked**
#1196 

**Is your feature request related to a problem? Please describe.**
We need to create a project details page that will also show the health of the project.

**Describe the solution you'd like**
On the Project details page:
- [ ] Create a section for the project details with the title `Project Details` as shown in the design below
- [ ] `Check FTE` button should open the FTE page (https://uat.employee.coloredcow.com/effort-tracking/project/35)
- [ ] `Open Sheet` button should open the effort sheet URL in the new tab.  
- [ ] If there is no effort sheet URL or any other details available, then just show `Not Available` text.
- [ ] The buttons should have the color code `#6cb2eb`. you can also use `btn-info` class for the color.

![image](https://user-images.githubusercontent.com/54376313/143677375-1c3d235d-6d66-4a55-ae99-297cf7dc697a.png)

**Describe alternatives you've considered**
NA

**Additional context**
NA

Ayush8923:
This issue is complete. Closing this.

PR #1243, opened 2022-01-21
Phase 2: Project details page is done.

Ayush8923:
Targets #1195 

### Checklist:

### DIFF
--- a/Modules/Project/Services/ProjectService.php
+++ b/Modules/Project/Services/ProjectService.php
@@ -119,6 +119,7 @@ class ProjectService implements ProjectServiceContract
     public function getWorkingDays($startDate, $endDate)
     {
         $period = CarbonPeriod::create($startDate, $endDate);
+        $dates = [];
         $weekend = ['Saturday', 'Sunday'];
         foreach ($period as $date) {
             if (! in_array($date->format('l'), $weekend)) {
@@ -136,6 +137,8 @@ class ProjectService implements ProjectServiceContract
      */
     public function getTeamMembersDetails($teamMembers)
     {
+        $teamMembersEffort = [];
+        $users = [];
         foreach ($teamMembers as $teamMember) {
             $userDetails = $teamMember->getUserDetails;
             $efforts = $teamMember->projectTeamMemberEffort()->get();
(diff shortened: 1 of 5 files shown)


Ayush8923:
Project details page on my screen looks like this.
Screenshot (25)

GauravGusain98:
@Ayush8923 
The UI is not looking correct to me. Fix it.

Also, the FTE that we want to show in the UI should be of the project not for each user. You can check if there is any such function available in the project model. If not then you need to create one for calculating the overall FTE of a project i.e (Sum of all team member effort / Total expected effort till that day of the month).

GauravGusain98 (changes requested):
Please check the comment

Modules/Project/Services/ProjectService.php
```diff
@@ -42,6 +44,117 @@ public function store($data)
...
+            }
+        }
+        $expectedHours = $this->getExpectedHours(count($workingDays));
+        $totalFTE = ($totalEffort / $expectedHours);
+
+        return [
+            'project' => $project,
+            'totalFTE' => $totalFTE,
+        ];
+    }
```
GauravGusain98, line 77: This function should only return the values that are needed to show the project data. It should not do other operations like calculating FTE, for that we should create another function

GauravGusain98: Let's do one thing, create `getFteAttribute()` function inside the `project` module so that we can get the FTE by using project objects like `$project->fte`

Modules/Project/Services/ProjectService.php
```diff
@@ -42,6 +44,117 @@ public function store($data)
...
+
+        return $dates;
+    }
+
+    /**
+     * Get Team members details.
+     * @param  array $teamMembers Team Members.
+     * @return array
+     */
+    public function getTeamMembersDetails($teamMembers)
```
GauravGusain98, line 125: Lets move this as well in project model

GauravGusain98 (changes requested):
Please check the comments.

rathorevaibhav (approved):

Issue #829, opened 2020-12-07
Use curly brackets format for template variables


pokhiii:
**Describe the task**
We also need to restructure the template variables in the config use the curly brackets format. I think that's a wider used template variable convention.

**Expected behavior**
1. Instead of `|*variable*|`, we should replace the template variables with `{{variable}}`
2. The second one is a bit more cleaner approach
3. The functionality should work exactly same as before and the variables should get updated accordingly
4. All the saved templates will be required to be manually updated
5. Update the seeder files for the templates where the older variable template is used

dsajwan2020:
The changes are not reflected in the email received by the candidate

Modified the HR email template as referred in screenshot

![image](https://user-images.githubusercontent.com/96309135/161742110-2c502fb3-1bba-4bde-8a34-017db357a4f4.png)


Mail received in the candidate  gmail

![image](https://user-images.githubusercontent.com/96309135/161742723-cd7ad01d-de51-413e-b4b2-7875973cb756.png)

PR #1272, opened 2022-03-10
curly brackets format for template variables are completed.

Ayush8923:
Targets #829


rathorevaibhav (approved):
This is great work. Just one note that we need to change the existing database entries manually for these templates and use the new format for variables. Can we add a note in the task?

Issue #1180, opened 2021-11-22
Part:4 Store the contracts in AWS


gautam071:
**Issue**
- The original issue #816 

**Describe the task**
- [ ] Save the contracts pdf in the AWS S3.
- [ ] Save the file location/path in the contracts_file _path column.

**References**
- Check how it is done in other places and replicate the code accordingly.

rathorevaibhav:
This needs to be rethought. Closing.

PR #1279, opened 2022-03-15
Contracts has been stored in local disk

Shivam-Samant:
#1180

### DIFF
--- a/Modules/Project/Entities/ProjectContract.php
+++ b/Modules/Project/Entities/ProjectContract.php
@@ -8,7 +8,7 @@ class ProjectContract extends Model
 {
     protected $fillable = ['project_id', 'contract_file_path'];
 
-    public function ProjectContract()
+    public function project()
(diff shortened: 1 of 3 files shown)


Modules/Project/Entities/ProjectContract.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Modules\Project\Entities;
+
+use Illuminate\Database\Eloquent\Model;
+
+class ProjectContract extends Model
+{
+    protected $fillable = ['project_id', 'contract_file_path'];
+
+    public function ProjectContract()
```
rathorevaibhav, line 11: this should be 
```suggestion
    public function project()
```

rathorevaibhav (changes requested):
Just one feedback. Rest looking good.

rathorevaibhav (approved):

PR #1323, opened 2022-04-13
Added filters for applications

GauravGusain98:
### Changes
 - Added filter for graduation start year
 - Added filter for graduation end year
 - Added filter for Jobs

### DIFF
--- a/Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
+++ b/Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
@@ -68,9 +68,32 @@ abstract class ApplicationController extends Controller
         $applications = Application::join('hr_application_round', function ($join) {
             $join->on('hr_application_round.hr_application_id', '=', 'hr_applications.id')
                 ->where('hr_application_round.is_latest', true);
-        })
-            ->with(['applicant', 'job', 'tags', 'latestApplicationRound'])
-            ->whereHas('latestApplicationRound')
+        })->with(['applicant', 'job', 'tags', 'latestApplicationRound']);
+        foreach (array_keys(request()->all()) as $filterKeys) {
+            switch ($filterKeys) {
+                case 'start-year':
+                    $startYear = request()->all()['start-year'] ? (int) request()->all()['start-year'] : null;
+                    if ($startYear != null) {
+                        $applications = $applications->whereHas('applicant', function ($query) use ($startYear) {
+                            $query->where('graduation_year', '>=', $startYear);
+                        });
+                    }
+                    break;
+                case 'end-year':
+                    $endYear = request()->all()['end-year'] ? (int) request()->all()['end-year'] : null;
+                    if ($endYear != null) {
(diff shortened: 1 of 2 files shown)


Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -68,9 +68,32 @@ public function index()
...
+                case 'start-year':
+                    $startYear = request()->all()['start-year'] ? (int) request()->all()['start-year'] : null;
+                    if ($startYear != null) {
+                        $applications = $applications->whereHas('applicant', function ($query) use ($startYear) {
+                            $query->where('graduation_year', '>=', $startYear);
+                        });
+                    }
+                    break;
+                case 'end-year':
+                    $endYear = request()->all()['end-year'] ? (int) request()->all()['end-year'] : null;
```
rathorevaibhav, line 83: I think this should return null if not present
```suggestion
                    $endYear = request()->get('end-year');
```

rathorevaibhav: Please check and update at other places

resources/views/hr/application/index.blade.php
```diff
@@ -9,10 +9,16 @@
         <div class="col-md-3">
             <h1>Applications</h1>
         </div>
-        <div class="input-group mb-3 col-md-6" style="display: flex;">
+        <div class="input-group mb-1 col-md-6" style="display: flex;">
             <form method="GET" action="/{{ Request::path() }}">
                 <input type="hidden" name="status" class="form-control" id="search"
                     value="{{ config('constants.hr.status.' . request("status") . '.label') }}">
+                <input type="hidden" name="start-year" class="form-control" id="search"
+                    value=@if(request()->has('start-year')){{request()->get('start-year')}}@endif>
```
rathorevaibhav, line 17: I think this should work
```suggestion
                    value={{ old('start-year', request()->get('start-year')) }}>
```

resources/views/hr/application/index.blade.php
```diff
@@ -29,32 +35,79 @@
...
+                        </select>
+                    </div>
+                    <div class="mr-2 form-group">
+                        <label id="end-year">{!! __('End Year') !!}</label><br>
+                        <select class="fz-14 fz-lg-16 p-1 w-100 w-md-150 form-control rounded border-0" name="end-year" id="start-year"
+                            onchange="this.form.submit()">
+                            <option value="" {{ request()->has('start-year') ? '' : 'selected' }}>
+                                {!! __('All Years') !!}
+                            </option>
+                            @for ($i=2000; $i <= ((int) date('Y') + 10); $i++)
```
rathorevaibhav, line 72: can we make this an input field? I don't think setting year 2000 as lower limit would be good

rathorevaibhav: both start and end date fields as text inputs

rathorevaibhav (changes requested):

rathorevaibhav (approved):

PR #1358, opened 2022-05-05
Hotfix: Effortsheet Module Datatype fix

kuldeep3:
### Description
Screenshot 2022-05-05 at 8 20 00 PM

1. Replaced `0` with an empty array to avoid the error thrown as shown in the above screenshot.

### Checklist:


GauravGusain98 (approved):
looks fine

PR #1365, opened 2022-05-07
Client page refactoring

GauravGusain98:
Targets #1366

### Description

- Add sorting in client partners
- Add search bar in clients
- Fix navigation tabs for project
- Added permission for client edit page
- Fix issue related to project editing.

### Workflow
Cilent status flowchart

### Checklist:

### DIFF
--- a/Modules/Client/Resources/views/index.blade.php
+++ b/Modules/Client/Resources/views/index.blade.php
@@ -2,17 +2,33 @@
 @section('content')
 
 <div class="container" id="vueContainer">
-    @include('client::menu_header')
-    <br>
-    <div class="d-flex justify-content-between mb-2">
-        <h4 class="mb-1 pb-1">{{ config('client.status')[request()->input('status', 'active')] }} Clients ({{ $count }})</h4>
-        @can('clients.create')
-        <span>
-            <a  href= "{{ route('client.create') }}" class="btn btn-info text-white"> Add new client</a>
-        </span>
-        @endcan
+    @can('clients.create')
+    <div class="d-none d-md-flex justify-content-between my-2">
+        @include('client::menu_header')
+        <a  href= "{{ route('client.create') }}" class="btn btn-info text-white"> Add new client</a>
     </div>
-    
+    @endcan
+    <div class="d-md-flex justify-content-between mt-5 mb-2">
+        <h4 class="mb-1 pb-1">{{ config('client.status')[request()->input('status', 'active')] }} Clients ({{ $count }})</h4>
+        <div>
+            <form action="{{ route('client.index') }}" method="GET">
+                <div class="d-flex align-items-center">
+                    <input type=""hidden name="status" value="{{ request()->get('status') ?? 'active' }}">
+                    <input type="text" name="name" class="form-control" id="name" placeholder="Client name"
+                    value={{request()->get('name')}}>
+                    <button class="btn btn-info ml-2 text-white">Search</button> 
+                </div>
+            </form>
+        </div>
+    </div>
+    <div class='d-md-none mb-2'>
+        @can('clients.create')
+            <div class="d-flex flex-row-reverse">
+                <a  href= "{{ route('client.create') }}" class="btn btn-info text-white"> Add new client</a>
+            </div>
+        @endcan
+        @include('client::menu_header')
+    </div>
     <div>
         <table class="table table-bordered table-striped">
             <thead class="thead-dark">
--- a/Modules/Client/Resources/views/index.blade.php
+++ b/Modules/Client/Resources/views/index.blade.php
@@ -5,7 +5,7 @@
     @can('clients.create')
     <div class="d-none d-md-flex justify-content-between my-2">
         @include('client::menu_header')
-        <a  href= "{{ route('client.create') }}" class="btn btn-info text-white"> Add new client</a>
+        <a href= "{{ route('client.create') }}" class="btn btn-info text-white">Add new client</a>
     </div>
     @endcan
     <div class="d-md-flex justify-content-between mt-5 mb-2">
--- a/Modules/Client/Resources/views/index.blade.php
+++ b/Modules/Client/Resources/views/index.blade.php
@@ -23,7 +23,7 @@
     <div class='d-md-none mb-2'>
         @can('clients.create')
             <div class="d-flex flex-row-reverse">
-                <a href= "{{ route('client.create') }}" class="btn btn-info text-white">Add new client</a>
+                <a href= "{{ route('client.create') }}" class="btn btn-info text-white">Add client</a>
             </div>
         @endcan
         @include('client::menu_header')
--- a/Modules/Client/Resources/views/index.blade.php
+++ b/Modules/Client/Resources/views/index.blade.php
@@ -13,7 +13,7 @@
         <div>
             <form action="{{ route('client.index') }}" method="GET">
                 <div class="d-flex align-items-center">
-                    <input type=""hidden name="status" value="{{ request()->get('status') ?? 'active' }}">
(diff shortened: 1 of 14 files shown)


kuldeep3 (approved):
Looks good on code. Minor feedback. @rathorevaibhav Could you also go through the files once?

Modules/Client/Resources/views/index.blade.php
```diff
@@ -2,15 +2,33 @@
...
+    @endcan
+    <div class="d-md-flex justify-content-between mt-5 mb-2">
         <h4 class="mb-1 pb-1">{{ config('client.status')[request()->input('status', 'active')] }} Clients ({{ $count }})</h4>
-        <span>
-            <a  href= "{{ route('client.create') }}" class="btn btn-info text-white"> Add new client</a>
-        </span>
+        <div>
+            <form action="{{ route('client.index') }}" method="GET">
+                <div class="d-flex align-items-center">
+                    <input type="hidden" name="status" value="{{ request()->get('status') ?? 'active' }}">
```
rathorevaibhav, line 16: Does this work?
```suggestion
                    <input type="hidden" name="status" value="{{ request()->get('status', 'active') }}">
```

Modules/Client/Resources/views/index.blade.php
```diff
@@ -2,15 +2,33 @@
...
+                    <input type="text" name="name" class="form-control" id="name" placeholder="Client name"
+                    value={{request()->get('name')}}>
```
rathorevaibhav, line 18:
```suggestion
                    <input type="text" name="name" class="form-control" id="name" placeholder="Client name" value={{request()->get('name')}}>
```

Modules/Client/Resources/views/subviews/listing-client-row.blade.php
```diff
@@ -1,7 +1,7 @@
 <tr>
     <td> 
         <span class="{{ $level ? 'ml-2' : ''  }}">
-            <a href="{{ route('client.edit', $client) }}">
+            <a @can('clients.update' || 'clients.create' || 'clients.delete') href="{{ route('client.edit', $client) }}" @endcan>
```
rathorevaibhav, line 4: Let's do this:
```blade
@canany(['clients.update', 'clients.create', 'clients.delete'])
    <a href="{{ route('client.edit', $client) }}"></a>
@else
     <div></div>
@canany
```

Modules/Client/Services/ClientService.php
```diff
@@ -14,11 +14,14 @@ class ClientService implements ClientServiceContract
 {
     public function index()
```
rathorevaibhav, line 15: this should accept a data array from wherever it's being called
```suggestion
    public function index($data)
```

And we should pass `request()->all()` as the data from controller. We should avoid directly using `request()` inside service file.

Modules/Client/Services/ClientService.php
```diff
@@ -14,11 +14,14 @@ class ClientService implements ClientServiceContract
...
     {
-        $status = request()->input('status', 'active');
-        $clients = Client::status($status)->with(['linkedAsPartner' => function ($subQuery) use ($status) {
-            return $subQuery->status($status);
-        }, 'linkedAsDepartment' => function ($subQuery) use ($status) {
-            return $subQuery->status($status);
+        $filters = [
+            'status' => request()->input('status', 'active'),
+            'name' => request()->get('name'),
```
rathorevaibhav, line 19: This will change to `$data['status']` and `$data['name']`

Modules/Client/Services/ClientService.php
```diff
@@ -150,8 +153,13 @@ private function updateClientDetails($data, $client)
...
         $data['is_channel_partner'] = $data['is_channel_partner'] ?? false;
         $data['has_departments'] = $data['has_departments'] ?? false;
+        $is_data_updated = $client->update($data);
+
+        if ($data['status'] ?? 'active' == 'inactive') {
+            $client->projects()->update(['status' => 'inactive']);
+        }
 
-        return $client->update($data);
+        return $is_data_updated;
```
rathorevaibhav, line 162:
```suggestion
        return $isDataUpdated;
```

Modules/Project/Http/Requests/ProjectRequest.php
```diff
@@ -27,8 +27,20 @@ public function rules()
         $request = $this->create_project ?? $this->update_section;
 
         switch ($request) {
-            case 'create_project':
             case 'project_details':
+                $rules = [
+                    'name' => 'required|string',
+                    'client_id' => 'required|integer',
+                    'status' => 'sometimes|string',
```
rathorevaibhav, line 34: when are you not sending `status` input?

Modules/Project/Resources/views/menu_header.blade.php
```diff
@@ -1,9 +1,18 @@
 <ul class="nav nav-pills my-3">
+    @php
+        $request = request()->all();
+    @endphp
     <li class="nav-item mr-3">
-        <a class="nav-link {{ (request()->input('status', 'active') == 'active') ? 'active' : '' }}" href="{{ route('project.index', ['status' => 'active'])  }}">Active Projects</a>
+        @php
+            $request['status'] = 'active';
+        @endphp
+        <a class="nav-link {{ (request()->input('status', 'active') == 'active') ? 'active' : '' }}" href="{{ route('project.index', $request)  }}">Active Projects</a>
```
rathorevaibhav, line 9: change this as per above comment please

Modules/Project/Services/ProjectService.php
```diff
@@ -69,6 +71,11 @@ public function getClients()
         return Client::where('status', 'active')->get();
     }
 
+    public function getAllClients()
```
rathorevaibhav, line 74: I feel like this is an overdo. Can't we simply use `Client::all()` wherever we need it?

Modules/Project/Services/ProjectService.php
```diff
@@ -110,7 +117,7 @@ public function updateProjectData($data, $project)
 
     private function updateProjectDetails($data, $project)
     {
-        return $project->update([
+        $is_project_updated = $project->update([
```
rathorevaibhav, line 120:
```suggestion
        $isProjectUpdated = $project->update([
```

Modules/Project/Services/ProjectService.php
```diff
@@ -121,6 +128,14 @@ private function updateProjectDetails($data, $project)
             'end_date' => date('Y-m-d'),
             'effort_sheet_url' => $data['effort_sheet_url'] ?? null,
         ]);
+
+        if ($project->client->projects()->where('status', 'active')->count()) {
```
rathorevaibhav, line 132: why this check is needed? And does it need to be run for every project update?

rathorevaibhav (changes requested):

Modules/Project/Http/Requests/ProjectRequest.php
GauravGusain98: Oh! this should be required now. Earlier the rules for creating and editing were using same rules due to which this was required. But now I've created separate rules. So this should be updated accordingly

Modules/Project/Services/ProjectService.php
GauravGusain98: If a project is mark as inactive then we need to check if for the client any project is inactive or not. If not then we mark client as inactive

GauravGusain98: Yes, it need to be check everytime as the status gets updated every time

rathorevaibhav: let's check the current project status first. If it's already active, then we don't need to check if other projects are active. This will save one query.

Also, use `$project->client->projects()->where('status', 'active')->exists()` instead of count()

rathorevaibhav (approved):

PR #1371, opened 2022-05-07
changes made for client name validation

Sahsranshu1609:


deepankar-sajwan:
this  issue is fixed in another PR

Issue #1351, opened 2022-05-02
Disable create button on one click


mohitsharma-22:
**Describe the bug**
Create invoice button is clickable again, which resulting creating 2  invoices in the portal. 
<img width="506" alt="Screenshot 2022-05-02 at 12 46 59 PM" src="https://user-images.githubusercontent.com/16541548/166198241-aa90e387-525f-4295-b34f-9d3bc09a4bcc.png">

<img width="1200" alt="Screenshot 2022-05-02 at 12 42 17 PM" src="https://user-images.githubusercontent.com/16541548/166198142-7b8c8c07-f8de-4e45-babd-0dc4edd097c3.png">

dsajwan2020:
Hi @mohitsharma-22  can you please confirm, that this issue is fixed in the production environment.

mohitsharma-22:
@dsajwan2020 the crate button is getting disabled and it is working fine.
As discussed, we need to add a message, "Invoice created successfully" on the click of create button. . Please get that also implemented. Thanks

deepankar-sajwan:
Hi @gautam071 can you display the message "Invoice created successfully" on the click of create button as metioned by @mohitsharma-22.

deepankar-sajwan:
Message added when creating invoice.

![image](https://user-images.githubusercontent.com/96309135/170827367-0e2222a0-ad4b-4257-a76d-82da554c80c5.png)

PR #1373, opened 2022-05-07
Added the fix for creating the multiple entries with single click

gautam071:
Targets #1351

### Description

- When a user clicks on the create button on the invoice page multiple times. The record is saved multiple times. Added the fix for this issue in this PR.
### Checklist:


dsajwan2020:
Hi @GauravGusain98 can you please do the code review for this PR

GauravGusain98 (approved):
Looks fine.

Modules/Client/Resources/views/subviews/create-client-details.blade.php
```diff
@@ -59,5 +59,17 @@ class="checkbox-custom mb-1.9 mb-1.67 mr-3"
...
+    const saveClient = (button) => {
+        button.disabled = true;
+        if (!button.form.checkValidity()) {
+            button.disabled = false;
+            button.form.reportValidity();
+            return;
+        }
+        button.form.submit();
+    }
+</script>
```
GauravGusain98, line 75: I see that this functionality is being used at multiple place. can we make a general function for this .You may include that function inside the main app.js file and then you just need to call that function in the button.

GauravGusain98 (changes requested):
Just a minor change requested.

GauravGusain98 (approved):
Looks fine

PR #1376, opened 2022-05-07
Fix invoice number 

pankaj-ag:
Targets #1377

### Description
- Added a new seeder to set the invoice number based on the filename. 
- Added a new migration for invoice number in invoice table 
- Showed invoice number on the listing invoice page. 
```
php artisan module:seed Invoice  --class=ResetInvoiceNumberTableSeeder
```

### Checklist:

### DIFF
--- a/Modules/Invoice/Database/Seeders/ResetInvoiceNumberTableSeeder.php
+++ b/Modules/Invoice/Database/Seeders/ResetInvoiceNumberTableSeeder.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Modules\Invoice\Database\Seeders;
+
+use Illuminate\Database\Seeder;
+use Modules\Invoice\Entities\Invoice;
+use Illuminate\Database\Eloquent\Model;
+
+class ResetInvoiceNumberTableSeeder extends Seeder
+{
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
+        Model::unguard();
+
+        foreach (Invoice::all() as $invoice) {
+            $invoice->invoice_number = pathinfo($invoice->file_path, PATHINFO_FILENAME);
+            $invoice->save();
+        }
+    }
+}
--- a/Modules/Invoice/Resources/views/index.blade.php
+++ b/Modules/Invoice/Resources/views/index.blade.php
@@ -5,7 +5,7 @@
         <br>
 
         <div class="d-flex justify-content-between mb-2">
-            <h4 class="mb-1 pb-1"> Invoices</h4>
+            <h4 class="mb-1 pb-1">Invoices</h4>
             <span>
                 <a href="{{ route('invoice.create') }}" class="btn btn-info text-white">Add Invoice</a>
             </span>
--- a/Modules/Invoice/Resources/views/index.blade.php
+++ b/Modules/Invoice/Resources/views/index.blade.php
@@ -88,7 +88,7 @@
                                                         <input type="email" multiple class="form-control mt-2"
                                                             name="invoice_email" id="invoice_email" placeholder="CC Email"
                                                             data-toggle="tooltip" data-placement="bottom"
-                                                            title="Please enter a email in comma seprated">
+                                                            title="Please enter comma separated emails">
                                                         <button type="submit" class="btn btn-primary mt-4">Send</button>
                                                     </form>
                                                 </div>
--- a/Modules/Invoice/Resources/views/index.blade.php
+++ b/Modules/Invoice/Resources/views/index.blade.php
@@ -7,7 +7,7 @@
         <div class="d-flex justify-content-between mb-2">
             <h4 class="mb-1 pb-1"> Invoices</h4>
             <span>
-                <a href="{{ route('invoice.create') }}" class="btn btn-info text-white"> Add new invoice</a>
+                <a href="{{ route('invoice.create') }}" class="btn btn-info text-white">Add Invoice</a>
             </span>
         </div>
         <br>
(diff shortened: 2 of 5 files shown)


Modules/Invoice/Database/Seeders/ResetInvoiceNumberTableSeeder.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Modules\Invoice\Database\Seeders;
+
+use Illuminate\Database\Seeder;
+use Modules\Invoice\Entities\Invoice;
+use Illuminate\Database\Eloquent\Model;
+
+class ResetInvoiceNumberTableSeeder extends Seeder
```
rathorevaibhav, line 9: this looks like a one time seeder, right? In that case, let's create an issue tracking to delete this in future please. I had to clean up a lot of one-time seeder files today.

Modules/Invoice/Resources/views/index.blade.php
```diff
@@ -1,93 +1,107 @@
...
 @section('content')
+    <div class="container" id="vueContainer">
+        <br>
+        <br>
 
-<div class="container" id="vueContainer">
-    <br>
-    <br>
+        <div class="d-flex justify-content-between mb-2">
+            <h4 class="mb-1 pb-1"> Invoices</h4>
```
rathorevaibhav, line 8:
```suggestion
            <h4 class="mb-1 pb-1">Invoices</h4>
```

Modules/Invoice/Resources/views/index.blade.php
```diff
@@ -1,93 +1,107 @@
...
+            <span>
+                <a href="{{ route('invoice.create') }}" class="btn btn-info text-white"> Add new invoice</a>
```
rathorevaibhav, line 10:
```suggestion
                <a href="{{ route('invoice.create') }}" class="btn btn-info text-white">Add Invoice</a>
```

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -71,6 +71,8 @@ public function store($data)
         $data['receivable_date'] = $data['due_on'];
         $invoice = Invoice::create($data);
         $this->saveInvoiceFile($invoice, $data['invoice_file']);
+        // Todo: We need to update the logic for this.
```
rathorevaibhav, line 74: let's add a bit more detailing here as in what needs to be done in the to-do? Or maybe what is wrong with the current approach? I'm not sure when we will come to that, so just this line may not make a lot of sense to someone reading in future.

For example:
```php
// Todo: We need to update the logic to set invoice numbers. It should get 
// generated using a combination of invoice id, project id, and client id.
```

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -81,6 +83,7 @@ public function update($data, $id)
         $invoice->update($data);
         if (isset($data['invoice_file']) and $data['invoice_file']) {
             $this->saveInvoiceFile($invoice, $data['invoice_file']);
+            $this->setInvoiceNumber($invoice);
```
rathorevaibhav, line 86: should this go into InvoiceObserver? As it's happening in both update and delete?

Modules/Invoice/Resources/views/index.blade.php
```diff
@@ -1,93 +1,107 @@
...
+                                                        <input type="hidden" name="year" value="{{ $_GET['year'] }}">
+                                                        <label for="sender_invoice_email">Sender's email adress</label>
+                                                        <input type="email" class="form-control mt-2"
+                                                            name="sender_invoice_email" id="sender_invoice_email"
+                                                            value="{{ config('invoice.pending-invoice-mail.pending-invoice.email') }}">
+                                                        <label for="invoice_email" class="mt-4">CC Email</label>
+                                                        <input type="email" multiple class="form-control mt-2"
+                                                            name="invoice_email" id="invoice_email" placeholder="CC Email"
+                                                            data-toggle="tooltip" data-placement="bottom"
+                                                            title="Please enter a email in comma seprated">
```
rathorevaibhav, line 91:
```suggestion
                                                            title="Please enter comma separated emails">
```

Modules/Invoice/Resources/views/index.blade.php
pankaj-ag: I did not make this change. It's old one.

Modules/Invoice/Database/Seeders/ResetInvoiceNumberTableSeeder.php
pankaj-ag: Make sense. https://github.com/ColoredCow/portal/issues/1378

Modules/Invoice/Services/InvoiceService.php
pankaj-ag: I was also thinking to do that but I was not sure. @P4NK4J will work on this and clean it up.

Modules/Invoice/Services/InvoiceService.php
pankaj-ag: Done.

Modules/Invoice/Services/InvoiceService.php
rathorevaibhav: No no. I'm talking about this line only. calling the `setInvoiceNumber()` function from observer. I don't think we need to move the whole function there.

pankaj-ag: Can't do that right now. If you see we are adding the invoice number based on the file path in the database (which is wrong though) and we need to fix that as well.

rathorevaibhav (approved):

Issue #1342, opened 2022-04-28
Designing the algorithm for generating invoice no for the invoice


deepankar-sajwan:
To create the invoice report ([here](https://github.com/ColoredCow/portal/issues/1324)), we would need the following items 
1. **Invoice number**: It will have the format - Client Type - Client ID - Project ID - Invoice sequence
for example(IN-001-001-0001  for international it is EX-001-001-0001). Client ID, Client Type, and Project ID would be available in the database. To get the invoice sequence, we would have to add `1` to the count for each client and its project, how many invoices have been sent. The invoice number will be a different column in the invoices table, with a text/Varchar data type. It would be calculated and stored at the time an invoice is created. 

2. **IGST, CGST, SGST** -  Currently, we have Haryana as the billing state, that we can store in the config of invoices module. Upon invoice details listing, we will have to see if the client's state is the same as the billing state, the CGST value should be 9%(picked from config) and SGST value should be 9%(picked from config), IGST will remain null in this case. If the client's state is different from the billing state, the IGST value would be 18% (picked from config), and CGST and SGST will remain null in this case. The client's state can be picked from the database.

P4NK4J:
To create the invoice report ([here](https://github.com/ColoredCow/portal/issues/1324)), we would need the following items 
1. **Invoice number**: It will have the format - Client Type - Client ID - Project ID - Invoice sequence
for example(IN-001-001-0001  for international it is EX-001-001-0001). Client ID, Client Type, and Project ID would be available in the database. To get the invoice sequence, we would have to add `1` to the count for each client and its project, how many invoices have been sent. The invoice number will be a different column in the invoices table, with a text/Varchar data type. It would be calculated and stored at the time an invoice is created.

2. IGST, CGST, SGST -  Currently, we have Haryana as the billing state, that we can store in the config of invoices module. Upon invoice details listing, we will have to see if client's state is same as the billing state, the CGST value should be 9%(picked from config) and SGST value should be 9%(picked from config), IGST will remain null in this case. If the client's state is different from billing state, the IGST value would be 18% (picked from config), CGST and SGST will remain null in this case. Client's state can be picked from the database.

GauravGusain98:
### clarified Requirements

- Invoice Number format - <IN/EX><client ID><project Id><Six digit invoice id><MM><YY>
     - <IN/EX> - `IN` for indian clients, `EX` For international clients
     - Client Id - Id of client stored in DB
[...]

PR #1380, opened 2022-05-07
Generating invoice number for Invoices

Shivam-Samant:
#1342


GauravGusain98:
@Shivam-Samant The invoice number is not in correct format.
- We don't want hyphen or nay kind of special character in it
- The month and year numbers are missing.

Please check the requirements and make the changes accordingly.

GauravGusain98:
@Shivam-Samant I think there was another requirement as well to calculate gst and igst based on the client state. Are you not covering that in this PR

Shivam-Samant:
> @Shivam-Samant I think there was another requirement as well to calculate gst and igst based on the client state. Are you not covering that in this PR?

This will be done by @Ayush8923 in this PR #1389.

GauravGusain98 (approved):
LGTM

Issue #1385, opened 2022-05-10
Project details are not updated when removing Team member from project


deepankar-sajwan:
**Is your feature request related to a problem? Please describe.**
When we try to remove the team member associated with Project from Project detail page, error occurs on updating team mebers.

Refer screenshot

 ![image](https://user-images.githubusercontent.com/96309135/167592260-43fafcc7-7095-4a6e-b0bc-79f633a352a5.png)

![image](https://user-images.githubusercontent.com/96309135/167592306-9eef8a1c-a554-48b6-9239-bf4e0dbb977b.png)

![image](https://user-images.githubusercontent.com/96309135/167592341-f2ae3f54-3d02-4487-a094-9d6ee40482cc.png)

![image](https://user-images.githubusercontent.com/96309135/167592514-ad09193c-f77d-4fb8-ba22-33b3a179159f.png)

 

 

 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

deepankar-sajwan:
- Editing  the project details 

![image](https://user-images.githubusercontent.com/96309135/168762625-0ffc3157-c997-49b0-b0b8-269c251dcc51.png)

![image](https://user-images.githubusercontent.com/96309135/168762758-c85c63ba-713f-4ae1-a7bf-231c480be949.png)

![image](https://user-images.githubusercontent.com/96309135/168762822-2028bce8-6e3d-4d0f-a0d7-d7da48dd8de0.png)


![image](https://user-images.githubusercontent.com/96309135/168762888-e0bdb92f-bf06-449a-ab76-c76506231847.png)

deepankar-sajwan:
Issue fixed in https://github.com/ColoredCow/portal/pull/1390

PR #1388, opened 2022-05-10
Fix/team member not getting remove

GauravGusain98:
Targets #1385 

### Description

- Fixed issue related to team member not getting removed

**Expected Behaviour**
- When a team member is removed form project then his entry should not be deleted from database instead the ended on column should be updated.
- If the same member is added again then a new entry should be created for that member

### Checklist:

### DIFF
--- a/Modules/Project/Services/ProjectService.php
+++ b/Modules/Project/Services/ProjectService.php
@@ -140,10 +140,10 @@ class ProjectService implements ProjectServiceContract
         $projectTeamMembers = $project->getTeamMembers;
         $teamMembersData = $data['project_team_member'] ?? [];
         foreach ($projectTeamMembers as $member) {
-            $flag = False;
+            $flag = false;
             foreach ($teamMembersData as $teamMemberData) {
                 if ($member->id == $teamMemberData['project_team_member_id']) {
-                    $flag = True;
+                    $flag = true;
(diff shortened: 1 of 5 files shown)


Modules/Project/Services/ProjectService.php
```diff
@@ -135,14 +137,33 @@ private function updateProjectDetails($data, $project)
...
-        foreach ($projectTeamMembers as $projectTeamMember) {
-            $teamMembers[$projectTeamMember['team_member_id']] = ['designation' => $projectTeamMember['designation']];
+        $projectTeamMembers = $project->getTeamMembers;
+        $teamMembersData = $data['project_team_member'] ?? [];
+        foreach ($projectTeamMembers as $member) {
+            $flag = false;
+            foreach ($teamMembersData as $teamMemberData) {
+                if ($member->id == $teamMemberData['project_team_member_id']) {
+                    $flag = true;
+                    $tempArray = $teamMemberData;
```
kuldeep3, line 147: Can we use some better name than $tempArray 🤔

kuldeep3 (approved):
Looks good. Left some suggestions.

Modules/Project/Services/ProjectService.php
GauravGusain98: I think it will work

Issue #1381, opened 2022-05-08
Remove unwanted columns from Client table


pankaj-ag:
**Describe the task**
Currently, the client table has a few unused columns, like state, phone, address, and pincode. We are not using those columns anymore and using the client_address table for that. 

We need to remove those columns because they are confusing the new contributors to the project. 

**Expected behavior**
Unwanted columns should not be present in the clients table.

PR #1391, opened 2022-05-11
Unused columns removed from client table

Shivam-Samant:
#1381


Modules/Client/Entities/Client.php
```diff
@@ -15,7 +15,7 @@ class Client extends Model
 {
     use HasHierarchy, HasFactory, Filters;
 
-    protected $fillable = ['name', 'email', 'key_account_manager_id', 'status', 'country', 'state', 'phone', 'phone', 'address', 'pincode', 'is_channel_partner', 'has_departments', 'channel_partner_id', 'parent_organisation_id'];
```
P4NK4J, line 18: Why we are removing email and phone number here ?

app/Http/Requests/ClientRequest.php
```diff
@@ -25,11 +25,7 @@ public function rules()
     {
         return [
             'name' => 'required|string',
-            'emails' => 'nullable|string',
-            'phone' => 'nullable|string',
```
P4NK4J, line 29: Are these fields not required? In the issue description, I didn't find such a requirement. Let me know if I am missing something.

P4NK4J (commented):

Modules/Client/Entities/Client.php
Shivam-Samant: Because we use this in the client_contact_persons table.

app/Http/Requests/ClientRequest.php
Shivam-Samant: These fields are required fields. We use those fields in the clientFormRequest.php file instead of the clientRequest.php file

P4NK4J (approved):

PR #1408, opened 2022-05-14
Added the notice when creating and updating the clients and projects

gautam071:
Targets #1351 

### Description

- When a client is created or updated a notice will appear 

### Checklist:


GauravGusain98 (approved):
Looks fine

PR #1418, opened 2022-05-16
invoice no. sequence fix

### LINKED ISSUE #1417: The Invoice number generated for the invoice is showing incorrect client id and invoice number
**Describe the bug**
When adding new invoice number the client number shown in the invoice number is incorrect, also the sequence of invoice number is incorrect. The clinet id should be 060 , also invoice number should be 000002

The issue occured in production environment

Refer screenshot

image

Shivam-Samant:
#1417


P4NK4J (approved):
Looks good.

Issue #1297, opened 2022-04-04
Add validation to avoid adding duplicate Client and Project name in Portal


mohitsharma-22:

dsajwan2020:
**Issue**
1. At present there is no validation implemented for client name. The user can create more than one client with same name. This results in redundancy in the system. 

Refer screenshot

![image](https://user-images.githubusercontent.com/96309135/165265796-7f528953-ccfe-4544-a0b2-d1dd952b770b.png)

2. At present there is no validation implemented for project name. The user can create more than one project with same name. This results in redundancy in the system. 

Refer screenshot 

![image](https://user-images.githubusercontent.com/96309135/165265213-84c0cda0-9bd6-4ab1-b923-dbba4f43d635.png)


**Possible solution**

1. When new client is created, validation check should be implemented once user enters client name , if the client with similar name already exist, message should be displayed to the user  "This name already exist. Please try another name."


2. When new project is created, validation check should be implemented once user enters project name , if the projects with similar name already exist, message should be displayed to the user  "This name already exist. Please try another name."

deepankar-sajwan:
Hi @gautam071 the validation for duplicate client and  duplicate project needs to be implemented on two scenario
- When new project or client is created
- When we edit the existing client or project

deepankar-sajwan:
The feature is implemented

PR #1421, opened 2022-05-17
Validation is added for duplicate client and project name

Shivam-Samant:
#1297

### DIFF
--- a/Modules/Project/Http/Requests/ProjectRequest.php
+++ b/Modules/Project/Http/Requests/ProjectRequest.php
@@ -3,6 +3,7 @@
 namespace Modules\Project\Http\Requests;
 
 use Illuminate\Foundation\Http\FormRequest;
+use Modules\Project\Rules\ProjectNameExist;
 
 class ProjectRequest extends FormRequest
 {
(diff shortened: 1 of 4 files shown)


Modules/Project/Rules/ProjectNameExist.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Determine if the validation rule passes.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function passes($attribute, $value)
+    {
+        return ! (Project::where('name', $value)->exists());
```
GauravGusain98, line 29: It will always fail if we update the project details as then also we will receive same name.

Modules/Project/Http/Requests/ProjectRequest.php
```diff
@@ -42,7 +43,7 @@ public function rules()
 
             case 'create_project':
                 $rules = [
-                    'name' => 'required|string',
+                    'name' => ['required', 'string', new ProjectNameExist],
```
GauravGusain98, line 46: For creating you can just use the Laravel `exists` rule. It will also work the same

GauravGusain98 (changes requested):
Please check the comment

GauravGusain98 (approved):

PR #1422, opened 2022-05-18
Bugfix/1299/finance/invoice download with invoice number

abhishekkhanduri:
#{1299}

### Description
Invoice File download with invoice number in new tab.


Modules/Invoice/Resources/views/edit.blade.php
```diff
@@ -1,6 +1,9 @@
 @extends('invoice::layouts.master')
 @section('content')
-
+@php
+$invoice_file_path=$invoice->file_path;
+$filename=basename($invoice_file_path);
```
P4NK4J, line 5: How about `$filename=basename($invoice->file_path);` ?

package-lock.json
```diff
@@ -1,16983 +1,6 @@
 {
-    "name": "portal",
```
P4NK4J, line 2: Why we are removing this file ?

P4NK4J (changes requested):
Left some comments, please take a look

P4NK4J:
@abhishekkhanduri There are some comments left to resolve in your PR, please check and then assign for review. Also, post a response corresponding to the comments.

Modules/Invoice/Resources/views/edit.blade.php
abhishekkhanduri: Yes this would be perfect.

package-lock.json
abhishekkhanduri: I have not made any changes in this file on my own.

.husky/pre-commit
```diff
@@ -1,4 +0,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
```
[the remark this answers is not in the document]
abhishekkhanduri: This file will be created automatically when we runs npm install command.

package-lock.json
abhishekkhanduri: This file will be created automatically when we run install composer.

P4NK4J (approved):

Issue #1424, opened 2022-05-19
The content available in GST report is not as per the requiremnt


deepankar-sajwan:
**Describe the bug**
GST report does not have values as specified in [#1340](https://github.com/ColoredCow/portal/issues/1340)

Refer screenshot

![image](https://user-images.githubusercontent.com/96309135/169292634-71d0b5c4-f855-4942-bfcf-44990e5df00a.png)

deepankar-sajwan:
Hi @Ayush8923  there are still some changes that are not been implemented as per requirement

Refer screen shot

![image](https://user-images.githubusercontent.com/96309135/169681950-e225e75a-bbe3-48ed-a33d-1330cfcf8202.png)

GST Column

**Export** should be available as option for international client
**GST number** should be available for indian client, if GST number not available, then add **B2C** instead of GST number

Type Column

**Indian** as option for indian client
**Export** as option for international client

Cgst column -- Amout of cgst is not available when client was from harayana
sgst column -- Amount is sgst not available when client was from harayana

Ayush8923:
This issue is completed. Closing it.

PR #1433, opened 2022-05-20
Suggested changes in monthly gst report

Ayush8923:
Target #1424

### DIFF
--- a/Modules/Invoice/Entities/Invoice.php
+++ b/Modules/Invoice/Entities/Invoice.php
@@ -86,14 +86,6 @@ class Invoice extends Model
         return trim(optional($country)->currency_symbol . ' ' . $amount);
     }
 
-    public function isInvoiceAmount()
-    {
-        $country = optional($this->client)->country;
-        $amount = (int) $this->amount;
-
-        return trim(optional($country)->currency_symbol . ' ' . $amount);
-    }
-
     public function shouldHighlighted()
     {
         if ($this->status == 'paid') {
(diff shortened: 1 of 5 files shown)


Modules/Invoice/Entities/Invoice.php
```diff
@@ -86,14 +86,6 @@ public function invoiceAmount()
...
     }
 
-    public function isInvoiceAmount()
-    {
-        $country = optional($this->client)->country;
-        $amount = (int) $this->amount;
-
-        return trim(optional($country)->currency_symbol . ' ' . $amount);
-    }
-
```
P4NK4J, line 96: Why this is removed ? and why it was added in the first place. How this removal is being resolved now ?

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -265,23 +265,19 @@ private function monthlyReportInvoices($filters)
...
+                'TAXABLE AMOUNT' => $invoice->amount,
+                'IGST' => ClientAddress::select('*')->where('client_id', $invoice->client_id)->first() ? ((ClientAddress::select('*')->where('client_id', $invoice->client_id)->first()->state != 'Haryana') ? ((int) $invoice->display_amount * (int) config('invoice.invoice-details.igst')) / 100 : '') : '',
+                'CGST' => ClientAddress::select('*')->where('client_id', $invoice->client_id)->first() ? ((ClientAddress::select('*')->where('client_id', $invoice->client_id)->first()->state == 'Haryana') ? ((int) $invoice->display_amount * (int) config('invoice.invoice-details.cgst')) / 100 : '') : '',
+                'SGST' => ClientAddress::select('*')->where('client_id', $invoice->client_id)->first() ? ((ClientAddress::select('*')->where('client_id', $invoice->client_id)->first()->state == 'Haryana') ? ((int) $invoice->display_amount * (int) config('invoice.invoice-details.sgst')) / 100 : '') : '',
```
P4NK4J, line 280: Shouldn't `Haryana` be handled from the config as a billing state variable ?

P4NK4J (commented):

Modules/Invoice/Entities/Invoice.php
Ayush8923: I removed it as we no need it as per requirement.  and added earlier because at that time we only needed the amount without GST.

P4NK4J (approved):

Issue #1299, opened 2022-04-04
Invoice should be downloaded with the invoice number, with which it is uploaded


mohitsharma-22:
Currently, it is downloaded with counter number .501..502 etc..

deepankar-sajwan:
The issue is fixed please refer screenshot

File is available as per invoice no

![image](https://user-images.githubusercontent.com/96309135/169756167-270800e3-ea2a-4687-80bd-43af59e64ef6.png)


Download file has same name as invoice file name

![image](https://user-images.githubusercontent.com/96309135/169756448-d250d174-0f62-4639-a9f9-8173bb6c72a2.png)

PR #1443, opened 2022-05-21
release of invoice file download

abhishekkhanduri:
#1299 

### Description
Invoice Will be downloaded with invoice number after opening it in new tab.


P4NK4J (approved):
Looks good.

PR #1452, opened 2022-05-23
Fix logic to fetch effortsheet data

GauravGusain98:
### Changes

- Fetching effort sheet data based on start date and end_date of the effortsheet.
- Updating the data if the data is fetched again for same date

### DIFF
--- a/Modules/Project/Console/SyncEffortsheet.php
+++ b/Modules/Project/Console/SyncEffortsheet.php
@@ -82,10 +82,11 @@ class SyncEffortsheet extends Command
                     continue;
                 }
 
-                $projectMonth = Carbon::create($user[1])->month;
-                $currentMonth = now()->month;
+                $billingStartDate = Carbon::create($user[1]);
+                $billingEndDate = Carbon::create($user[2]);
+                $currentDate = now(config('constants.timezone.indian'))->today();
 
-                if ($projectMonth !== $currentMonth) {
+                if($currentDate < $billingStartDate || $currentDate > $billingEndDate) {
                     continue;
                 }
 
@@ -95,25 +96,28 @@ class SyncEffortsheet extends Command
                     continue;
                 }
 
-                $latestProjectTeamMemberEffort = $projectTeamMember->projectTeamMemberEffort()->orderBy('added_on', 'DESC')->first();
+                $latestProjectTeamMemberEffort = $projectTeamMember->projectTeamMemberEffort()
+                    ->where('added_on', '<', Carbon::now(config('constants.timezone.indian'))->format('Y-m-d'))
+                    ->orderBy('added_on', 'DESC')->first();
                 $actual_effort = $user[4];
 
                 if ($latestProjectTeamMemberEffort) {
                     $previous_effort_date = Carbon::parse($latestProjectTeamMemberEffort->added_on);
-
-                    if ($previous_effort_date->format('Y-m-d') == Carbon::now()->format('Y-m-d')) {
-                        continue;
-                    } elseif ($previous_effort_date->format('Y-m') == Carbon::now()->format('Y-m')) {
+                    if ($previous_effort_date->format('Y-m-d') >= $billingStartDate && $previous_effort_date->format('Y-m-d') <= $billingEndDate) {
                         $actual_effort -= $latestProjectTeamMemberEffort->total_effort_in_effortsheet;
                     }
                 }
 
-                ProjectTeamMemberEffort::create([
-                    'project_team_member_id' => $projectTeamMember->id,
-                    'actual_effort' => $actual_effort,
-                    'total_effort_in_effortsheet' => $user[4],
-                    'added_on' => now(),
-                ]);
+                ProjectTeamMemberEffort::updateOrCreate(
+                    [
+                        'project_team_member_id' => $projectTeamMember->id,
+                        'added_on' => Carbon::now(config('constants.timezone.indian'))->format('Y-m-d'),
+                    ],
+                    [
+                        'actual_effort' => $actual_effort,
+                        'total_effort_in_effortsheet' => $user[4],
+                    ]
+                );
             }
         }
     }


Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
 
-                if (! $portalUser) {
+                if (! $effortSheetURL) {
                     continue;
                 }
 
-                $projectMonth = Carbon::create($user[1])->month;
-                $currentMonth = now()->month;
+                $matchesId = [];
+                $matchesSheetId = preg_match('/.*[^-\w]([-\w]{25,})[^-\w]?.*/', $effortSheetURL, $matchesId);
```
P4NK4J, line 58: Can these be a bit readable ? Are these variables matching some ID or a matched ID is being stored in them ?

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
 
-                if ($projectMonth !== $currentMonth) {
+                if (! $matchesSheetId) {
                     continue;
                 }
 
-                $projectTeamMember = $portalUser->projectTeamMembers()->active()->where('project_id', $project->id)->first();
+                $sheetId = $matchesId[1];
```
P4NK4J, line 64: Whey we are taking hardcoded index here ?

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
+                $sheet = new Sheets();
+                $projectMembersCount = $project->teamMembers()->count();
+                $range = 'C2:G' . ($projectMembersCount + 1); // this will depend on the number of people on the project
```
P4NK4J, line 67: Can this range be taken from config ? I think different sheets may have different ranges

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
+                                    ->get();
+                } catch (Exception $e) {
                     continue;
                 }
 
-                $latestProjectTeamMemberEffort = $projectTeamMember->projectTeamMemberEffort()->orderBy('added_on', 'DESC')->first();
-                $actual_effort = $user[4];
+                foreach ($sheets as $user) {
+                    $userNickname = $user[0];
+                    $portalUsers = clone $users;
```
P4NK4J, line 79: How portal users are different from users ?

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
-                ProjectTeamMemberEffort::create([
-                    'project_team_member_id' => $projectTeamMember->id,
-                    'actual_effort' => $actual_effort,
-                    'total_effort_in_effortsheet' => $user[4],
-                    'added_on' => now(),
-                ]);
+                    $latestProjectTeamMemberEffort = $projectTeamMember->projectTeamMemberEffort()
+                        ->where('added_on', '<', Carbon::now(config('constants.timezone.indian'))->format('Y-m-d'))
+                        ->orderBy('added_on', 'DESC')->first();
+                    $actual_effort = $user[4];
```
P4NK4J, line 103: Why the indexing is hardcoded here ?

P4NK4J, line 103: Why variable naming convention is changed suddenly ?

P4NK4J (changes requested):

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
-            }
-
-            foreach ($sheets as $user) {
-                $userNickname = $user[0];
-                $portalUsers = clone $users;
-                $portalUser = $portalUsers->where('nickname', $userNickname)->first();
+                $effortSheetURL = $project->effort_sheet_url;
 
-                if (! $portalUser) {
+                if (! $effortSheetURL) {
```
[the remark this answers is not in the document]
GauravGusain98: Done

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: This will be changed and will be calculated dynamically in upcoming functionalities. So for this will work.

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
+                    $portalUser = $portalUsers->where('nickname', $userNickname)->first();
+
+                    if (! $portalUser) {
+                        continue;
+                    }
 
-                if ($latestProjectTeamMemberEffort) {
-                    $previous_effort_date = Carbon::parse($latestProjectTeamMemberEffort->added_on);
+                    $billingStartDate = Carbon::create($user[1]);
+                    $billingEndDate = Carbon::create($user[2]);
```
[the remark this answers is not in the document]
GauravGusain98: This is also the data coming from sheet.

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: user is the data of user coming from sheet and protal user is the object of user model

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
-                ]);
+                    $latestProjectTeamMemberEffort = $projectTeamMember->projectTeamMemberEffort()
+                        ->where('added_on', '<', Carbon::now(config('constants.timezone.indian'))->format('Y-m-d'))
+                        ->orderBy('added_on', 'DESC')->first();
+                    $actualEffort = $user[4];
+
+                    if ($latestProjectTeamMemberEffort) {
+                        $previousEffortDate = Carbon::parse($latestProjectTeamMemberEffort->added_on);
+
+                        if ($previousEffortDate->format('Y-m-d') >= $billingStartDate->format('Y-m-d') && $previousEffortDate->format('Y-m-d') <= $billingEndDate->format('Y-m-d')) {
```
rathorevaibhav, line 108: you can use carbon functions to compare two carbon objects. No format required in that case.
https://stackoverflow.com/questions/29684111/how-to-compare-two-carbon-timestamps

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
+                            $actualEffort -= $latestProjectTeamMemberEffort->total_effort_in_effortsheet;
+                        }
+                    }
+
+                    ProjectTeamMemberEffort::updateOrCreate(
+                        [
+                            'project_team_member_id' => $projectTeamMember->id,
+                            'added_on' => $currentDate->format('Y-m-d'),
```
rathorevaibhav, line 116: You can use datecasting in Laravel to avoid formatting everytime for a datetime field. Simply pass the carbon object and it will do the rest.
https://laravel.com/docs/9.x/eloquent-mutators#date-casting

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
 
-                if (! $portalUser) {
+                if (! $effortSheetUrl) {
                     continue;
                 }
 
-                $projectMonth = Carbon::create($user[1])->month;
-                $currentMonth = now()->month;
+                $correctedEffortsheetUrl = [];
+                $isSyntaxMatching = preg_match('/.*[^-\w]([-\w]{25,})[^-\w]?.*/', $effortSheetUrl, $correctedEffortsheetUrl);
```
rathorevaibhav, line 58: can you put a one-line or a bit longer comment to help understand what you're trying to do. Assuming the reader knows what syntax is there in the effortsheet that you're trying to match might not be effective.

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
 
-                if ($projectMonth !== $currentMonth) {
+                if (! $isSyntaxMatching) {
                     continue;
                 }
 
-                $projectTeamMember = $portalUser->projectTeamMembers()->active()->where('project_id', $project->id)->first();
+                $sheetId = $correctedEffortsheetUrl[1];
```
rathorevaibhav, line 64: any reason for not picking up `$correctedEffortsheetUrl[0]`?

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
+                    $sheets = $sheet->spreadsheet($sheetId)
+                                    ->range($range)
+                                    ->get();
+                } catch (Exception $e) {
                     continue;
                 }
 
-                $latestProjectTeamMemberEffort = $projectTeamMember->projectTeamMemberEffort()->orderBy('added_on', 'DESC')->first();
-                $actual_effort = $user[4];
+                foreach ($sheets as $user) {
```
rathorevaibhav, line 77: naming convention looks a bit weird. From `$sheets` I took it as an array of sheets but it seems like an array of users. Can we correct one of them?
```php
// either this
foreach ($sheets as $sheet) {

// or this
foreach ($users as $user) {
```

Or maybe some different name if these two are not right.

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
+
+                    if (! $portalUser) {
+                        continue;
+                    }
 
-                if ($latestProjectTeamMemberEffort) {
-                    $previous_effort_date = Carbon::parse($latestProjectTeamMemberEffort->added_on);
+                    $billingStartDate = Carbon::create($user[1]);
+                    $billingEndDate = Carbon::create($user[2]);
+                    $currentDate = now(config('constants.timezone.indian'))->today();
```
rathorevaibhav, line 88: why are we hardcoding indian timezone? Can't we pick it up from env `APP_TIMEZONE`?

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
-                }
 
-                ProjectTeamMemberEffort::create([
-                    'project_team_member_id' => $projectTeamMember->id,
-                    'actual_effort' => $actual_effort,
-                    'total_effort_in_effortsheet' => $user[4],
-                    'added_on' => now(),
-                ]);
+                    $latestProjectTeamMemberEffort = $projectTeamMember->projectTeamMemberEffort()
+                        ->where('added_on', '<', Carbon::now(config('constants.timezone.indian'))->format('Y-m-d'))
```
rathorevaibhav, line 101: can we make it consistent? Somewhere we've used the helper `now()` and somewhere we're using `Carbon::now()`. Though it doesn't matter, I think let's keep one so that it does not escalate more in future.

Modules/Project/Console/SyncEffortsheet.php
P4NK4J: Do upcoming functionalities mean when the effortsheet will be moved from spreadsheet to portal? `For this will work  ` does it mean is it being created for a specific case ?

Modules/Project/Console/SyncEffortsheet.php
P4NK4J: Can we rename them to `$projectResources` for the sheet user and keep is users for the portal ?

Modules/Project/Console/SyncEffortsheet.php
```diff
@@ -47,73 +47,82 @@ public function handle()
...
+                            'added_on' => $currentDate->format('Y-m-d'),
+                        ],
+                        [
+                            'actual_effort' => $actualEffort,
+                            'total_effort_in_effortsheet' => $user[4],
+                        ]
+                    );
+                }
+            } catch (Exception $e) {
+                continue;
```
P4NK4J, line 125: Are we displaying the exception or any user-related error here? I mean if functionality breaks suddenly, what would happen here ?

P4NK4J (commented):
Left some comments

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: done

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: We are updating the column type to date. So it is not required.

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: Actually in the URL, there is a gid parameter for identifying the sheet. So we are just removing that so that we only get the first sheet assuming it is the effortsheet for the current month.

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: As we are getting the URL in the first index.

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: Done

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: As we wan to update the effortsheet according to indian timezone. We don't want to get the back date for the case when the data gets updated in a time range where date changes for different timezones. 

Not using app timezone as it may change in future IST. So using the timezone that is already present in the codebase for IST.

Modules/Project/Console/SyncEffortsheet.php
GauravGusain98: done

Modules/Project/Console/SyncEffortsheet.php
P4NK4J: Resolved in offline discussion

P4NK4J (approved):
Looks good to me, please check the comments posted by @rathorevaibhav

Issue #1441, opened 2022-05-21
When adding team members in the project details error occurs if designation of team member not added


deepankar-sajwan:
**Describe the bug**
In updating project details , when adding details of the team members if we do not select designation of the team member, and save the form error is displayed on the screen.

Refer screenshot

![image](https://user-images.githubusercontent.com/96309135/169640031-84252d8a-c896-4fa2-9cbe-3d2a6b17794d.png)

deepankar-sajwan:
The error  message still exist 

![image](https://user-images.githubusercontent.com/96309135/171168680-32bd76a2-394d-4291-a23c-656b473bf23e.png)

deepankar-sajwan:
The issue has been fixed

![image](https://user-images.githubusercontent.com/96309135/171615736-5dd824f3-ca1d-4e70-96dc-218150ea4d7d.png)


![image](https://user-images.githubusercontent.com/96309135/171615772-6e59d8a7-c7ce-4099-a48d-3b6c9e3a72e3.png)


![image](https://user-images.githubusercontent.com/96309135/171615823-5d05178b-6743-48e9-bae9-913b1d4e4274.png)

PR #1489, opened 2022-05-31
Changes in error message according to the form data

abhishekkhanduri:
Targets #1441

### Description
Error messages are added according to the validations in the Project team members form.


P4NK4J (changes requested):
@abhishekkhanduri I see the linting checks are failing right now, please correct them and then assign for review

Issue #1506, opened 2022-06-03
Enable save button if error occured while project details editing


GauravGusain98:
**Describe the bug**
When an error comes while editing project details then the save button gets disabled. The user then needs to refresh the whole page to get it enabled. 

**To Reproduce**
1. Go to the project edit page
2. Got to a team member
3. Add a new team member and do not enter any field
4. Save the form
5. Error comes and the save button gets disabled

**Expected behavior**
If an error comes the save button should get enabled.

**Screenshots**
<img width="1387" alt="image" src="https://user-images.githubusercontent.com/54376313/171837788-a5c19579-1295-4d32-ab07-181d38367922.png">

**Additional context**
NA

PR #1512, opened 2022-06-04
enable save button

JCArya:
Targets #1506
Enable the save button if an error occurred while project details editing

### DIFF
--- a/Modules/Project/Resources/views/edit.blade.php
+++ b/Modules/Project/Resources/views/edit.blade.php
@@ -82,9 +82,9 @@
                 },
 
                 updateProjectForm: async function(formId) {
-                    $('.save-btn').attr('disabled', true);
+                    $('.save-btn').attr('enable', true);
                     let formData = new FormData(document.getElementById(formId));
-                    $('.save-btn').removeClass('btn-primary').addClass('btn-dark');
+                    $('.save-btn').removeClass('btn-primary').addClass('btn-primary');
                     await axios.post('{{ route('project.update', $project) }}', formData)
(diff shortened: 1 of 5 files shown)


Modules/Project/Resources/views/edit.blade.php
```diff
@@ -82,9 +82,9 @@ class="text-theme-body text-decoration-none mb-2 mb-xl-4 d-inline-flex align-ite
                 },
 
                 updateProjectForm: async function(formId) {
-                    $('.save-btn').attr('disabled', true);
+                    $('.save-btn').attr('enable', true);
                     let formData = new FormData(document.getElementById(formId));
-                    $('.save-btn').removeClass('btn-primary').addClass('btn-dark');
+                    $('.save-btn').removeClass('btn-primary').addClass('btn-primary');
                     await axios.post('{{ route('project.update', $project) }}', formData)
```
GauravGusain98, line 88: This will make the enable button always enable, but we want to make it disable until we get response from backend and if any error comes then we want to make the button enable again.

GauravGusain98: Also only keep the changes of this file and don't include the changes related to `composer.lock`, `package-lock.json`, `package.json`, `routes/web.php`

GauravGusain98 (changes requested):
Check the comments and make the changes as requested.

JCArya:
closeing this PR and some changes

JCArya:
closing this PR since it has a lot of unwanted changes.

PR #1516, opened 2022-06-06
Enable save button issue 1506

JCArya:
Targets #{1506}
Enable the save button if an error occurred while project details editing


GauravGusain98 (approved):
Looks good

PR #1525, opened 2022-06-07
fix we click on close button or cross icon

JCArya:
Targets #{1523}
Modal not closing when we click on the close button or cross icon


JCArya:
Due to some problem in the file I had close the PR

PR #1526, opened 2022-06-07
fix when we click on close button

JCArya:
Targets #{1523}

Modal not closing when we click on the close button or cross icon


GauravGusain98 (approved):
Looks fine

PR #1555, opened 2022-06-15
FIx issue related to invoice generation

GauravGusain98:
### Changes
- Fix issue where client_id is not getting auto update on creation of new client
- Fix issue related to sequence number throwing error when no earlier invoice exist.
- Billing Level not being shown in the projects details page


P4NK4J (approved):
Looks good.

PR #1556, opened 2022-06-15
Added badge for primary project while invoice creation

GauravGusain98:
### Changes
- If a client don't have a primary project then we should show a warning that no primary project set for the client while creating invoice
- Corrected calculation to fetch the project level billing and client level billing projects for the client

### DIFF
--- a/Modules/Client/Entities/Client.php
+++ b/Modules/Client/Entities/Client.php
@@ -46,7 +46,7 @@ class Client extends Model
 
     public function projectLevelBillingProjects()
     {
-        return $this->hasMany(Project::class)->where('projects.status', '!=', 'inactive')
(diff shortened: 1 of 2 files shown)


Modules/Client/Entities/Client.php
```diff
@@ -46,7 +46,7 @@ public function projects()
 
     public function projectLevelBillingProjects()
     {
-        return $this->hasMany(Project::class)->where('projects.status', '!=', 'inactive')
+        return $this->hasMany(Project::class)->select('projects.*')->where('projects.status', '!=', 'inactive')
```
GauravGusain98, line 49: Needed as we were getting id as null. So explicitly fetching only the data related to projects table

P4NK4J (approved):

Issue #1290, opened 2022-03-30
Add a filter job candidates based on their respective college or university, and  date of application.


deepankar-sajwan:
**Is your feature request related to a problem? Please describe.**
There is no process for filtering candidates based on their respective college or university, also their date of application. 
The filter can be used to check how many job applicants, belonging to a particular University or college, secured a job. 
We can also evaluate, the amount of time spent in the recruitment process of those job applicants.

**Describe the solution you'd like**
Add a filter, which can filter the job applicant based on their university or college, along with the recruitment time duration. 
This information can be used for making the recruitment process more time-efficient since we can give preference to job applicants of a particular college or university based on the time duration of recruitment.

dsajwan2020:
![image](https://user-images.githubusercontent.com/96309135/160768269-b6443a32-ea27-4a47-a41b-26e66cf448ef.png)

P4NK4J:
Closing this issue, since it's deployed on the production

PR #1570, opened 2022-06-17
Add-University-dropdown

JCArya:
Targets #1290
Add a filter of job candidates based on their respective college or university


P4NK4J (approved):
Looks good.

Issue #1540, opened 2022-06-10
Pending invoice list should include invoice having status as overdue


deepankar-sajwan:
**Describe the task**
At present the list of pending invoices, which is sent to finance department every monday, includes invoices that have  status as **sent**. Since invoice **overdue** status are also pending invoices. Hence they should be included in pending invoices list.

**Expected behavior**
Invoice having status as **overdue** should be included in the pending invoice list.

mohitsharma-22:
@gautam071 @deepankar-sajwan sharing the email structure for your reference
<img width="516" alt="Screenshot 2022-06-17 at 12 28 01 PM" src="https://user-images.githubusercontent.com/16541548/174243696-6027e774-ba8d-4d78-8dcf-3c2ae5a2d22d.png">
.

gautam071:
Hey @mohitsharma-22 

Thanks for sharing.

PR #1579, opened 2022-06-18
Fix/pending invoice

gautam071:
Targets #1540 

### Description
- Added the overdue column for pending invoices

### Checklist:

### DIFF
--- a/Modules/Invoice/Services/InvoiceService.php
+++ b/Modules/Invoice/Services/InvoiceService.php
@@ -3,6 +3,7 @@
 namespace Modules\Invoice\Services;
 
 use App\Models\Country;
+use Carbon\Carbon;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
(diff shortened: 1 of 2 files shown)


P4NK4J (approved):

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -124,7 +125,7 @@ public function delete($invoiceID)
 
     public function getUnpaidInvoices()
     {
-        return Invoice::status('sent')->with(['client', 'project'])->get();
+        return Invoice::status('sent')->with(['client', 'project'])->where('due_on', '>', Carbon::today())->get();
```
GauravGusain98, line 128: Do we need to get the invoices whose due date is in the future? 

I think the condition should be opposite. Please confirm.

GauravGusain98 (commented):
Please check the comment.

PR #1588, opened 2022-06-20
Project Members History table added

abhishekkhanduri:
### Description
Added project member history table

### Checklist:

### DIFF
--- a/Modules/Project/Entities/ProjectTeamMember.php
+++ b/Modules/Project/Entities/ProjectTeamMember.php
@@ -13,6 +13,13 @@ class ProjectTeamMember extends Model
 
     protected $guarded = [];
 
+    protected $dates = [
+        'started_on',
+        'ended_on',
+        'created_at',
+        'updated_at',
+    ];
+
     protected static function newFactory()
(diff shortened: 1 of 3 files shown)


Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -13,6 +13,13 @@ class ProjectTeamMember extends Model
 
     protected $guarded = [];
 
+    protected $dates = [
+        'started_on',
```
GauravGusain98, line 17: we don't have started_on column in table. So we don't need this. Please remove

GauravGusain98 (changes requested):
Small change requested, else looks good

Modules/Project/Entities/ProjectTeamMember.php
abhishekkhanduri: Ok, I will remove it.

abhishekkhanduri:
image
Table looks like this

GauravGusain98 (approved):
Looks good

Issue #1568, opened 2022-06-17
Phase 2 : Hiring guidelines and resources


P4NK4J:
**Describe the task**
On the guidelines landing page, we can have different cards. Each card can have job labels on them and will open up a view for resources corresponding to the job opportunity. Sharing the screenshot for UI reference.

**Expected behavior**
1. Create a grid with different cards having job opportunity labels on them.
2. Clicking on the job opportunity card, the user should be redirected to a common view page, that will have content in the future. As of now, we can keep it blank. 

**References**
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/35001775/174228037-22600996-2184-4c11-be12-1cd06338b85d.png">

PR #1603, opened 2022-06-22
Added job cards to the resources listing page

JCArya:
Targets #1568
 
Description:
added cards related to the job and create a view of every card

### DIFF
--- a/Modules/HR/Http/Controllers/ResourcesController.php
+++ b/Modules/HR/Http/Controllers/ResourcesController.php
@@ -11,11 +11,12 @@ class ResourcesController extends Controller
     public function index()
     {
         $jobs = Job::select('title')->get();
-        return view('hr::guidelines-resources.index', compact('jobs'));    
+        
+        return view('hr::guidelines-resources.index', compact('jobs'));   
     }
 
     public function show()
     {
-        return view('hr::guidelines-resources.show');
+        return view('hr::guidelines-resources.show';
     }
 }
--- a/Modules/HR/Http/Controllers/ResourcesController.php
+++ b/Modules/HR/Http/Controllers/ResourcesController.php
@@ -4,7 +4,6 @@ namespace Modules\HR\Http\Controllers;
 
 use Illuminate\Routing\Controller;
 use Modules\HR\Entities\Job;
-use Illuminate\Http\Request;
 
 
 class ResourcesController extends Controller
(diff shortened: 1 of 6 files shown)


resources/views/layouts/navbar.blade.php
```diff
@@ -25,7 +25,7 @@
                 @endcan
                 <a class="dropdown-item" href="{{route('userappointmentslots.show',auth()->id())}}">Appointment Slots</a>
                 <a class="dropdown-item" href="{{ route('hr.tags.index') }}">{{ __('Manage Tags') }}</a>
-                <a class="dropdown-item disabled" href="{{ route('resources.index') }}">Guidelines And Resources</a>
+                <a class="dropdown-item " href="{{ route('resources.index') }}">Guidelines And Resources</a>
```
P4NK4J, line 28: Let's keep it disabled for a while, we can enable it once the page is ready to use

Modules/HR/Resources/views/guidelines-resources/show.blade.php
```diff
@@ -0,0 +1,14 @@
+@extends('hr::layouts.master')
+@section('content')
+<div class="row col-gap justify-content-center">
+    <div class="col-md-8">
+        <h3>Resources for Laravel Developer</h3>
+    </div>
+    <div class="col-md-2">
+        <a href="#" class="btn btn-success btn-block btn-lg">Add Resources</a>
```
P4NK4J, line 8: Let's add a comment here `//ToDo: Functionality yet to be created`

Modules/HR/Http/Controllers/ResourcesController.php
```diff
@@ -3,11 +3,19 @@
...
 
 use Illuminate\Routing\Controller;
+use Modules\HR\Entities\Job;
 
 class ResourcesController extends Controller
 {
     public function index()
     {
-        return view('hr::guidelines-resources.resources');
+        $jobs = Job::select('title')->get();
```
P4NK4J, line 12: Why we are using `$jobs = Job::select('title')->get();`, what would happen if we use `all()` ?

P4NK4J (changes requested):
Left some comments, please take a look

resources/views/layouts/navbar.blade.php
JCArya: okk

P4NK4J (approved):

Issue #1561, opened 2022-06-15
Search filter on application listing page only takes first word of input


P4NK4J:
**Describe the bug**
When we search applications on the listing page using keywords, the results are displayed based on the first letter of the input string. We need to fix the search query, it should also consider the space-separated strings. 

**To Reproduce**
1. Go to 'HR->Recruitment'
2. On the search form, enter a keyword, let's say the applicant's full name. 
3. You will see the listing would appear based on the results having the first word of the input string.

**Expected behavior**
Application should be filtered based on the complete input provided in the search bar

**Screenshots**
Input string
<img width="1430" alt="image" src="https://user-images.githubusercontent.com/35001775/173834932-8c10bd7e-920a-4e88-9bdb-8fc7adc2e1e2.png">

Results
![Uploading image.png…]()

PR #1621, opened 2022-06-25
Fixed search filter on application listing page only takes first word of input...

Bahugunajii:
Target #1561


P4NK4J:
@Bahugunajii The search filters are not working as expected. The value of search input is not being restored if we switch the tab from open applications to closed or something else. Could you fix that and assign it for review? For reference, you can check out the job filter, like its value is restored in the filter parameters upon tab switching as well. Let me know if you have any questions.

P4NK4J (approved):
Looks good.

Issue #1619, opened 2022-06-25
Phase 4 Hiring guidelines and resources


P4NK4J:
**Describe the task**
We need to create migrations for the resources table database schema added in this issue (https://github.com/ColoredCow/portal/issues/1612)

**Expected behavior**
Create the migrations as per the DB schema

PR #1628, opened 2022-06-27
create migrations for resources and guidelines

JCArya:
Targets #1619

 Description
we are creating two migrations for hr_resouces and hr_resources_categories


P4NK4J (approved):
looks good.

Issue #1542, opened 2022-06-10
Functionality to mark an application "on hold" and send an email to the candidate


P4NK4J:
**Is your feature request related to a problem? Please describe.**
There are instances when the recruitment team is not confident in calling a job applicant for an interview. Then the recruitment team can place the application on hold, and send a mail to the candidate conveying the same message. This will save time and energy for all the persons involved in the recruitment process. 
If the recruitment team is unable to evaluate the application of the candidate, based on their evaluation criteria then put the application on hold. Also, convey the same message to the candidate through email.

**Describe the solution you'd like**
1. On the edit application page, if an application is under the resume screening phase, provide a button saying `Put on Hold`, next to the Take action and Reject button. 
2. Clicking on which, a popup would appear that will ask the evaluating person if he/she wants to put the application on hold. Clicking on Yes, the application would move to the on-hold tab. 
3. Once an application is put on hold, an email should be triggered for the candidate. We need to create an email template for this, it can be treated as another issue.

Ayush8923:
This issue PR is merged. Closing it.

PR #1647, opened 2022-06-28
Functionality to mark an application "on hold" and send an email to the candidate

Bahugunajii:
Targets #1542


P4NK4J:
@Bahugunajii Sharing a couple of observations:

- [ ] Can we adjust the width of the dropdown so that the buttons look a bit presentable in the UI



- [ ] Can we have the variables mapping the mail template for putting an application on hold, sharing the mail template below:
```
Dear |*applicant_name*|, 

Thanks for applying to ColoredCow for the position |*job_title*|. We don't have a relevant opening for the position right now. We have kept, your application is kept under waiting and we will reach out to you if there are any possibilities we can explore together.

Thanks,
HR team,
ColoredCow
```
|*applicant_name*| and  |*job_title*| will be mapped in the backend based on the application details. You can see other mail templates for reference.

- [ ] On putting an application to hold, the status and label of the application are not changed. Once an application is made on hold, the page should redirect to the application listing page, and the application should be moved to the on-hold section with the `on-hold` label on it. Currently, it sits in the `Open` listing with the `new application` tag.

Issue #1623, opened 2022-06-25
Renaming Monthly Tax report file which is exported through Export to Excel button


deepankar-sajwan:
**Describe the task**
In current implementation the Monthly Tax report is exported, through Export to Excel button as TaxReportExport file.
TaxReportExport  file name  does not  include month and year for which it is geneated. 

**Expected behavior**
The  TaxReportExport  file name  should include month, and year for which it is geneated. For instance if Monthly tax report
is generated for Jun 2022, the file name should be TaxReportExport-06-2022

**References**
Specify any reference material / links

PR #1661, opened 2022-06-29
Fixed Renaming Monthly Tax report file which is exported through Export to Excel button

Bahugunajii:
Targets #1623


P4NK4J (approved):
Looks good.

PR #1691, opened 2022-07-02
Feature: Cron jobs listing

gautam071:
Targets #1692 

### Description
- Displayed all the corn job details with the run immediately button only for the super admin
Screenshot 2022-07-02 at 6 23 17 PM

### Checklist:


resources/views/settings/index.blade.php
```diff
@@ -36,6 +36,16 @@
                 </div>
             </div>
         @endif
+
+        @if (auth()->user()->hasAnyPermission(['super-admin']))
```
GauravGusain98, line 40: I think this should be role instead of permission as we don't have super-admin as permission

GauravGusain98 (changes requested):
Minor changes requested. Please check

app/Http/Controllers/Settings/CronController.php
```diff
@@ -0,0 +1,49 @@
...
+use App\Http\Controllers\Controller;
+use Carbon\Carbon;
+use Cron\CronExpression;
+use Illuminate\Support\Facades\Artisan;
+
+class CronController extends Controller
+{
+    public function index()
+    {
+        app()->make(\Illuminate\Contracts\Console\Kernel::class);
```
rathorevaibhav, line 14: what does this line do?

app/Http/Controllers/Settings/CronController.php
```diff
@@ -0,0 +1,49 @@
...
+        $schedule = app()->make(\Illuminate\Console\Scheduling\Schedule::class);
+
+        $events = collect($schedule->events())->map(function ($event) {
```
rathorevaibhav, line 17: can you add a 1-2 liner comment on what exactly is happening here?

app/Http/Controllers/Settings/CronController.php
```diff
@@ -0,0 +1,49 @@
...
+            $command = str_after($event->command, '\'artisan\' ');
+            $commandDescription = '';
+            foreach (Artisan::all() as $artisanCommand => $commandDetails) {
+                if ($artisanCommand == $command) {
+                    $commandDescription = $commandDetails->getDescription();
+                    break;
+                }
+            }
+            $cron = CronExpression::factory($event->expression);
+            $date = Carbon::now();
```
rathorevaibhav, line 27: what's the use of this variable?

app/Http/Controllers/Settings/CronController.php
```diff
@@ -0,0 +1,49 @@
...
+                'command' => $command,
+                'next_run_at' => $cron->getNextRunDate()->format('d, M, Y, g:i A'),
+                'description' => $commandDescription,
+            ];
+        });
+
+        return view('settings.cron.index')->with('events', $events);
+    }
+
+    public function run($command)
```
rathorevaibhav, line 43: can we typehint the parameter?

resources/views/settings/index.blade.php
```diff
@@ -36,6 +36,16 @@
                 </div>
             </div>
         @endif
+
+        @if (auth()->user()->hasAnyRole(['super-admin']))
```
rathorevaibhav, line 40: Let's use the `@role` directive here

routes/web.php
```diff
@@ -126,6 +126,9 @@
         });
 
         Route::get('/nda-template', 'NDAAgreementController@index')->name('setting.agreement.nda');
+
+        Route::get('/cron', 'CronController@index')->name('settings.cron');
```
rathorevaibhav, line 130: What happens if a non super-admin user directly types the URL on browser? Will he/she go to that page?

rathorevaibhav: I believe you need to add Laravel Policy at controller level

rathorevaibhav (changes requested):

app/Http/Controllers/Settings/CronController.php
rathorevaibhav: any update on my question @gautam071 ? I don't see there is a variable assigned here so wanted to understand what's happening here.

app/Http/Controllers/Settings/CronController.php
rathorevaibhav: @gautam071 you missed my comment 👆

routes/web.php
```diff
@@ -126,6 +126,9 @@
...
+        Route::get('/cron/{command}', 'CronController@run')->name('settings.cron.run');
```
rathorevaibhav, line 131: What happens if a non super-admin user directly types the URL on browser? Will he/she go to that page?

resources/views/settings/index.blade.php
```diff
@@ -36,6 +36,16 @@
                 </div>
             </div>
         @endif
+
+        @role('super-admin')
```
rathorevaibhav, line 40: there should be permissions based checks not role based.

app/Http/Controllers/Settings/CronController.php
```diff
@@ -0,0 +1,55 @@
...
+                    break;
+                }
+            }
+            // CRON expression parser that can determine whether or not a CRON expression.
+            $cron = CronExpression::factory($event->expression);
+
+            return (object) [
+                'expression' => $event->expression,
+                'command' => $command,
+                'next_run_at' => $cron->getNextRunDate()->format('d, M, Y, g:i A'), // Converts the cron expression to a human readable date.
```
rathorevaibhav, line 40: can we also add a row in the UI to show `last_run_at`, if that is something that is supported in-built. That would be handy also? Sorry I'm adding this requirement at the end!

app/Http/Controllers/Settings/CronController.php
```diff
@@ -0,0 +1,55 @@
...
+
+        return view('settings.cron.index')->with('events', $events);
+    }
+
+    public function run(string $command)
+    {
+        $this->authorize('run', 'super-admin');
+        Artisan::call($command);
+
+        return redirect()->route('settings.cron')->with('success', 'Command executed successfully');
```
rathorevaibhav, line 53: will the page keep loading until the artisan command is finished? If yes, do we disable the button so that user does not click it again?

rathorevaibhav (changes requested):

app/Http/Controllers/Settings/CronController.php
```diff
@@ -0,0 +1,55 @@
...
+
+use App\Http\Controllers\Controller;
+use Cron\CronExpression;
+use Illuminate\Support\Facades\Artisan;
+
+class CronController extends Controller
+{
+    public function index()
+    {
+        $this->authorize('index', 'super-admin');
```
rathorevaibhav, line 13: This should be a permission-based check and not a role-based check.

app/Http/Controllers/Settings/CronController.php
gautam071: The Cron variable is used to convert the corn expression to a readable format

app/Http/Controllers/Settings/CronController.php
gautam071: I have added a comment in the code on what exactly happens

gautam071: You can look [here](https://github.com/ColoredCow/portal/blob/ea01fa5a42f9d462a03b294ba0421bc21516d531/app/Http/Controllers/Settings/CronController.php#L18)

app/Http/Controllers/Settings/CronController.php
gautam071: @rathorevaibhav This gets all the data from the kernel file.
I just added it globally because we don't have to use this every time.
like this 
`app()->make(\Illuminate\Contracts\Console\Kernel::class)->make(\Illuminate\Console\Scheduling\Schedule::class);`

routes/web.php
gautam071: It will show an error of 403 forbidden

resources/views/settings/index.blade.php
gautam071: @rathorevaibhav As Gaurav suggested I went with roles. Changing back to permissions

app/Http/Controllers/Settings/CronController.php
gautam071: Okay! 👍

app/Http/Controllers/Settings/CronController.php
gautam071: I am collecting the events from the kernel file schedule function mapping through each function and getting the required details

Issue #1686, opened 2022-07-02
Change search and add client button similar to active clients


P4NK4J:
**Describe the task**
Currently, the `search` and `add client` buttons are not in alignment with the other buttons. We would like the background color of both these buttons similar to the active clients.
<img width="1417" alt="image" src="https://user-images.githubusercontent.com/35001775/176991823-44586897-ce6b-4710-82a8-787a6388a372.png">

Tushar403:
sir can you explain it in detail please

P4NK4J:
@Tushar403 In the screenshot above, you can see that the color of `Add Client` and `search` button is different from the other buttons in portal. It should be the same as `Active Client` button. Let me know if this is clear, or you need more details.

Tushar403:
ok sir, thank you for explaining

deepankar-sajwan:
The task has been done please refer screenshot

![Uploading image.png…]()

PR #1699, opened 2022-07-04
color_changed in add client and search button

Tushar403:
Targets #1686 

### Description
changed background color of both add client and search buttons similar to the active clients.

### Checklist:

### DIFF
--- a/Modules/Client/Resources/views/index.blade.php
+++ b/Modules/Client/Resources/views/index.blade.php
@@ -12,7 +12,7 @@
     <div class="d-none d-md-flex justify-content-between my-2">
         @include('client::menu_header')
         @can('clients.create')
-            <a href= "{{ route('client.create') }}" class="btn btn-info text-white">Add client</a>
+            <a href= "{{ route('client.create') }}" class="btn btn-info text-white active">Add client</a>
         @endcan
     </div>
     <div class="d-md-flex justify-content-between mt-5 mb-2">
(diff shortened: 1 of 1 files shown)


Modules/Client/Resources/views/index.blade.php
```diff
@@ -58,10 +58,14 @@
 
                 @empty
                     <tr>
-                        <td colspan="2">
+                        <td>  </td>
+                        <td> - </td>
+                        <td>  </td>
```
P4NK4J, line 63: Why you are removing colspan and adding manual rows ?

Modules/Client/Resources/views/index.blade.php
```diff
@@ -58,10 +58,14 @@
...
                             <p class="my-4 text-left">No {{ config('client.status')[request()->input('status', 'active')] }} clients found.</p>
                         <td>
                     </tr>
+                   
+
```
rathorevaibhav, line 68: please remove extra lines.

P4NK4J (commented):
left a comment, please check

Modules/Client/Resources/views/index.blade.php
```diff
@@ -12,7 +12,7 @@
     <div class="d-none d-md-flex justify-content-between my-2">
         @include('client::menu_header')
         @can('clients.create')
-            <a href= "{{ route('client.create') }}" class="btn btn-info text-white">Add client</a>
+            <a href= "{{ route('client.create') }}" class="btn btn-info text-white active">Add client</a>
```
rathorevaibhav, line 15: what does `active` class do? Why have you added it here?

Modules/Client/Resources/views/index.blade.php
rathorevaibhav, line 63: can you explain these changes? I don't understand the reason for these changes from the issue description.

rathorevaibhav (changes requested):

Modules/Client/Resources/views/index.blade.php
Tushar403: because i got confused while performig a task, i thought i have to make table also.

Tushar403: i fixed this issue in master branch

Issue #1664, opened 2022-06-30
Update information available in help icon when changing status of client from inactive to active


deepankar-sajwan:
**Describe the task**
The  information mentioned in the help icon available in Status of the  Inactive Client, there is no information about  how 
to change Status of Client from Inactive to Active

**Expected behavior**
The following information should be added in the help icon

To change status to active please update status of any project associated with the client as active

PR #1703, opened 2022-07-04
Update information available in help icon when changing status of client from inactive to active

Ayushigupta2:
Targets #1664 

### Description
Added the information about how to change Status of Client from Inactive to Active.

### Checklist:


P4NK4J (approved):
Looks good.

PR #1705, opened 2022-07-04
color_changed in add client and search button

Tushar403:
Targets #1686 

### Description
changed background color of both add client and search buttons similar to the active clients.

### Checklist:

### DIFF
--- a/Modules/Client/Resources/views/index.blade.php
+++ b/Modules/Client/Resources/views/index.blade.php
@@ -12,7 +12,7 @@
     <div class="d-none d-md-flex justify-content-between my-2">
         @include('client::menu_header')
         @can('clients.create')
-            <a href= "{{ route('client.create') }}" class="btn btn-info text-white active">Add client</a>
+            <a href= "{{ route('client.create') }}" class="btn btn-primary text-white">Add client</a>
         @endcan
     </div>
     <div class="d-md-flex justify-content-between mt-5 mb-2">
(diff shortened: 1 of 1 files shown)


P4NK4J (commented):
What's the use of active class here?

Tushar403:
> What's the use of active class here?

because to apply a dark colour we have to use active class here .

Modules/Client/Resources/views/index.blade.php
```diff
@@ -12,7 +12,7 @@
     <div class="d-none d-md-flex justify-content-between my-2">
         @include('client::menu_header')
         @can('clients.create')
-            <a href= "{{ route('client.create') }}" class="btn btn-info text-white">Add client</a>
+            <a href= "{{ route('client.create') }}" class="btn btn-info text-white active">Add client</a>
```
P4NK4J, line 15: Let's use `btn btn-primary text-white` instead of `active` class, since active class is generally used for navigation or switching purposes, in bootstrap

P4NK4J (changes requested):
Left a comment, please check.

Modules/Client/Resources/views/index.blade.php
Tushar403: so do i have to make changes in this branch .

P4NK4J: > so do i have to make changes in this branch .

Yes of course, code reviews and feedback cycles work on a single branch for a single issue, you don't create different branches each time. Make changes to this code, and push them, assign for review again. If everything looks good, this branch would be merged into main and deployed for testing on UAT

P4NK4J: > done with my changes you can check now.

Can you request for review again ?

P4NK4J (approved):

Issue #1673, opened 2022-07-01
Navigation Dropdown UI


P4NK4J:
**Describe the bug**
Currently, while we are on the projects listing page, if we hover on HR the dropdown UI is a bit off. The project listing page can be accessed from navigation through the CRM menu item. 
We want the dropdown to appear like the one shared below
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/35001775/176849636-187cb7a2-9a20-48a3-a2e2-345895504e5f.png">
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/35001775/176849783-890f9c64-16eb-4f07-99b5-f484d58a979b.png">

P4NK4J:
Closing the issue, its tested deployed and working fine

PR #1707, opened 2022-07-04
DropDown UI Changes

aquibmoin786:
#1673

### Description
Changed Navigation Dropdown UI
### Checklist:

### DIFF
--- a/Modules/Project/Resources/views/index.blade.php
+++ b/Modules/Project/Resources/views/index.blade.php
@@ -50,9 +50,9 @@
         <table class="table table-bordered table-striped">
             <thead class="thead-dark">
                 <tr>
-                    <th>Client/Project Name</th>
-                    <th>Team Members</th>
-                    <th>Velocity (Hours)</th>
+                    <th class="w-33p sticky-top">Client/Project Name</th>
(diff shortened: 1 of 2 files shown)


Modules/Project/Resources/views/index.blade.php
```diff
@@ -50,9 +50,9 @@
         <table class="table table-bordered table-striped">
             <thead class="thead-dark">
                 <tr>
-                    <th class="w-33p sticky-top">Client/Project Name</th>
-                    <th class="sticky-top">Team Members</th>
-                    <th class="sticky-top">Velocity (Hours)</th>
+                    <th>Client/Project Name</th>
+                    <th>Team Members</th>
+                    <th>Velocity (Hours)</th>
```
kuldeep3, line 55: If we remove the sticky class from these headings, these will not show for the users whose projects lies at the bottom of the page. 

You need to preserve this sticky header and then fix the issue with the dropdown. You can have a look at z-index property in CSS, might be useful here.

kuldeep3 (changes requested):

kuldeep3 (approved):

Issue #1720, opened 2022-07-05
Updates to installation guide regarding forbidden error in Apache


rathorevaibhav:
**Describe the task**
When setting up the portal, @amisha296 faced an error that throws a page saying "Forbidden". There were some changes to the virtual host file that made it work.

**Expected behavior**
An optional step should be added to the installation guide at the time of virtual host step, with the proper error message in case a user sees and what to change/modify in the virtual host file.

deepankar-sajwan:
The issue has been fixed refer screenshot

![image](https://user-images.githubusercontent.com/96309135/179158230-e92c1e6e-c64b-44da-93b5-57951a24e0ab.png)

PR #1731, opened 2022-07-05
Updation in installation guidelines

amisha296:
Targets #1720 

### Description
I added an extra optional step in the installation guide I made some changes in the virtual host file(added a line).
It was required because if you are not using the latest version of apache you could come up with the error "403 Forbidden", by adding the steps of the change that I have made one can get rid of this error.

### Checklist:

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -174,18 +174,11 @@ Before you start following the guidelines, make sure to go through the [prerequi
                 </Directory>
             </VirtualHost>
             ```
-            - After making the above code, if you are getting the "403 Forbidden" error. Consider the following changes:
+            - After adding the above code, if you are getting the "403 Forbidden" error, try to resolve it through the following addition in your code:
             ```apacheconf
             <Directory>
                  # some code above
-                 Allow from all
-            </Directory>
-            ```
-            Change to:
-            ```apacheconf
-            <Directory>
-                 # some code above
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -174,6 +174,22 @@ Before you start following the guidelines, make sure to go through the [prerequi
                 </Directory>
             </VirtualHost>
             ```
+            - After making the above code, if you are getting the "403 Forbidden" error. Consider the following changes:
```
rathorevaibhav, line 177:
```suggestion
            - After making the above code, if you are getting the "403 Forbidden" error, try adding the following line in your virtual host configuration:
```

docs/installation.md
```diff
@@ -174,6 +174,22 @@ Before you start following the guidelines, make sure to go through the [prerequi
...
+            </Directory>
+            ```
+            Change to:
+            ```apacheconf
+            <Directory>
+                 # some code above
+                 Allow from all
+                 Require all granted
+            </Directory>
+            ```
```
rathorevaibhav, line 191: We can just mention the line to add. Let's skip writing both before/after so that the documentation size is smaller
```apacheconf
<Directory>
     # some code above
     Allow from all    # add this line
     # some code below
</Directory>
```

rathorevaibhav (approved):

PR #1751, opened 2022-07-06
Commented out email functionality

P4NK4J:
After putting an application on hold, we need to think of the workflow. The mail going to candidates is not needed anymore. Till the time there is thought process is in place. The mail trigger is prevented.


GauravGusain98 (approved):
Looks fine

Issue #1719, opened 2022-07-05
Switch from hasAnyPermission to canAny


rathorevaibhav:
**Describe the task**
Based on research, I found that `hasAnyPermission` or `hasPermissionTo` functions check in the DB directly whether a user has a permission or not. They do not consider Laravel Gates and policies.

**Expected behavior**
1. Replace all occurrences of `hasAnyPermission` with `canAny`
2. Replace all occurrences of `hasPermissionTo` with `can`
3. In the blade files, use `@can` or `@canAny` as needed

**References**
1. https://laravel.com/docs/9.x/authorization
2. https://spatie.be/docs/laravel-permission/v5/basic-usage/direct-permissions

rathorevaibhav:
This is fixed. Closing.

PR #1754, opened 2022-07-07
Switched to canAny

shardulgusain:
#1719

### Description
 1.hasAnyPermission is replaced with  canAny
 2.hasPermissionPermission functions check in the DB directly whether a user has a permission or not. They do not consider Laravel Gates and policies, hence replaced with canAny 

### Checklist:


P4NK4J (approved):
Looks good.

PR #1766, opened 2022-07-07
Change invoice table email column type

GauravGusain98:
### Changes
- As we are plannig to keep count of reminder email, so changing the type and name of the column


P4NK4J (commented):
Looks good.

P4NK4J (approved):

Issue #1747, opened 2022-07-06
Phase 8: Hiring guidelines and resources


JCArya:
Describe the task
we need to show a listing of categories in the table.

Expected behavior
when we will be adding resources we need to show categories in the table.

PR #1771, opened 2022-07-08
Show categories in the table

JCArya:
Targets #1747

Description
show all the categories and resources in the table.


Modules/HR/Resources/views/guidelines-resources/show.blade.php
```diff
@@ -16,9 +16,29 @@
...
+                <th>Categories</th>
+                <th>Actions</th>
+            </tr>
+        </thead>
+        <tbody>
+            @foreach($categories as $category)
+                <tr>
+                    <td>{{ $category->id }}</td>
+                    <td>
+                        <a href="#">{{ $category->name }}</a>
```
P4NK4J, line 32: This data should be fetched from hr_resources_table, where there is a resource URL present corresponding to a category and a job ID. right now, it fetches all the categories and renders them.

P4NK4J (changes requested):

PR #1791, opened 2022-07-09
Change in days calculation logic for different client billing dates

abhishekkhanduri:
### Description
Changes in logic in calculating working days ( now its based upon client billing date).

### Checklist:


Modules/Client/Entities/Client.php
```diff
@@ -197,4 +197,38 @@ public function getEffortSheetUrlAttribute()
...
+
+            return today(config('constants.timezone.indian'))->subMonth()->addDays($billingDate - today(config('constants.timezone.indian'))->day);
+        }
+
+        if (today(config('constants.timezone.indian'))->day >= $billingDate) {
+            return today(config('constants.timezone.indian'))->startOfMonth()->addDays($billingDate->subday());
+        }
+    }
+
+    public function getEndBillingDateAttribute()
```
GauravGusain98, line 217:
```suggestion
    public function getClientMonthEndDateAttribute()
```

Modules/Client/Entities/Client.php
```diff
@@ -197,4 +197,38 @@ public function getEffortSheetUrlAttribute()
             }
         }
     }
+
+    public function getStartBillingDateAttribute()
```
GauravGusain98, line 201:
```suggestion
    public function getClientMonthStartDateAttribute()
```

GauravGusain98 (changes requested):
Please check the comment

Issue #1788, opened 2022-07-09
Add coming soon template to Leads page


P4NK4J:
**Describe the task**
Currently, the Leads page on the portal is having an off-themed landing page. We want to change it into a coming soon page template, with Employee portal theme, branding and navigation
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/35001775/178101901-3b6b1662-3dc9-460d-9647-1d6c00ab0b2f.png">

Ayush8923:
This issue is completed. Closing it.

Ayush8923:
This issue is completed. Closing it.

PR #1797, opened 2022-07-11
Feature/1788/Add_coming_soon_template_to_leads

Sachinbisht27:
Targets #1788

### Description
I changed lead page theme to ColoredCow main theme and added a coming soon template to it.

### Checklist:

### DIFF
--- a/Modules/Lead/Resources/views/index.blade.php
+++ b/Modules/Lead/Resources/views/index.blade.php
@@ -1,4 +1,4 @@
-@extends('layouts.app')
+@extends('lead::layouts.master')
 
 @section('content')
 <div class="task-content">
--- a/Modules/Lead/Resources/views/layouts/master.blade.php
+++ b/Modules/Lead/Resources/views/layouts/master.blade.php
@@ -1,8 +1,6 @@
 @extends('layouts.app')
 @section('js_scripts')
-{{-- <script src="{{ mix('/js/user.js') }}" defer></script> --}}
 @endsection
 
 @section('css_scripts')
-{{-- <link href="{{ mix('/css/user.css') }}" rel="stylesheet"> --}}
 @endsection


P4NK4J (changes requested):
@Sachinbisht27 I don't think you need to create header and everything again. That's already created. You just need to see how you can include them here, probably it would be including any master/app blade file that will bring the changes here. Please correct this and then assign for review.

Sachinbisht27:
@P4NK4J I don't get you. Does i have to make changes in master/app blade file? And What are those major changes if they are not creating headers. Please explain.

P4NK4J:
@Sachinbisht27 If you take a latest pull from master branch, run `npm run dev` and then access the `Task` module from navigation, it will give you some reference, how you can make the leads page similar to it. Let me know if that makes sense

image

Modules/AppointmentSlots/Resources/views/layouts/master.blade.php
```diff
@@ -1,4 +1,4 @@
-@extends('layouts.app')
+@extends('')
```
P4NK4J, line 1: why this is changed ?

P4NK4J (changes requested):
@Sachinbisht27 Left some comments, also could you share the screenshot of how it is looking on the UI.

P4NK4J (changes requested):
Left a comment, please check

Sachinbisht27:
@P4NK4J At first I used HTML code of main coloredcow portal to get the theme for the lead page, then you said me that we don't have to add headers and all again and again then I learned about @extends & @yield used in laravel to get content from other pages and used @extends('layouts.app') in master.blade.php file.
This is my first time on laravel and i did some mistakes.
Now I understand that why we use @extends (which is a laravel tamplate which defines its own sections using @yield) and @yield (used to define section in layout and used to get content from one page to another page).
Attaching new screenshot of master.blade.php file. 
Screenshot (9)

Does I have to make more changes?

P4NK4J:
@Sachinbisht27 No worries, we always learn from our mistakes. Glad you researched a lot about these helpers. If you feel like everything is fine from your end, please push the code and assign it for review. The portal team reviewers would take a look and take action accordingly.

Modules/AppointmentSlots/Resources/views/layouts/master.blade.php
Sachinbisht27: I don't remember when I did these changes in AppointmentSlots.
I don't know sir, I think by mistake.

Sachinbisht27: I am correcting it and adding new PR

P4NK4J (commented):
A minor comment, else looks good

Sachinbisht27:
> @Sachinbisht27 If you take a latest pull from master branch, run `npm run dev` and then access the `Task` module from navigation, it will give you some reference, how you can make the leads page similar to it. Let me know if that makes sense
> 
> image

@P4NK4J  
I searched everywhere and tried so much to align ComingSoon!! text in the middle of the page. But it always aligned in the center of the row not in the center of the page.
Till this morning I haven't did this method which you suggested to do pull req. and all..
Now I saw the [ div class="task-content" ] i did't saw this bootstrap class anywhere.
But now by taking your reference task is done. 

I have a question.
please suggest me any reference where i can learn about <div class="task-content">.

Sachinbisht27:
please suggest me any reference where i can learn about div class="task-content"

P4NK4J:
@Sachinbisht27 `text-center` is a class in Bootstrap used to align the text in the center of the section. In CSS you would right it to
```
div {
text-align: center;
}
```
I told you to replace the `style="margin-top:6cm` with bootstrap class `mt`, if you write it like `mt-2`, it means, that `margin-top: 2*spacer (spacer will be defined in your variables.scss)`.

Let me know if that answers your question.

Sachinbisht27:
> @Sachinbisht27 `text-center` is a class in Bootstrap used to align the text in the center of the section. In CSS you would right it to
> 
> ```
> div {
> text-align: center;
> }
> ```
> 
> I told you to replace the `style="margin-top:6cm` with bootstrap class `mt`, if you write it like `mt-2`, it means, that `margin-top: 2*spacer (spacer will be defined in your variables.scss)`.
> 
> Let me know if that answers your question.

@P4NK4J I tried to replace style with mt already but it allows me to increase by mt-25 only after that the div length ends and the text goes up as it is.

Attaching some SS:

1) When i used mt-25
Screenshot (13)


2) When i used mt-26
Screenshot (14)

P4NK4J:
@Sachinbisht27 mt-26, just see inspect in the browser how much margin is being applied. Why on a different screen is not correct, and how you can use bootstrap media queries to resolve the same. You can refer to Tasks page coming soon template as well.

Sachinbisht27:
@P4NK4J i have added new PR as i take reference from task module from nav.

Issue #1761, opened 2022-07-07
Books page, replace JS alert with modal


P4NK4J:
**Describe the task**
Currently, when we click on delete button for a book, a JS alert appears for confirmation. We would like to replace it with a modal, that would ask users for deletion in the same manner, with a `Yes` and `Cancel` button. The background of Yes button should be red and text should be white in color. For the Cancel button, it should be gray background with white text
<img width="1433" alt="image" src="https://user-images.githubusercontent.com/35001775/177724442-76c59163-ef18-47f0-ba6c-bf2633461e73.png">

Ayush8923:
This issue is completed. Closing it.

PR #1801, opened 2022-07-11
change in color

akriti12negi:
#1761

### Description
 change in color 

### Checklist:


P4NK4J (changes requested):
@akriti12negi A couple of feedbacks:
1. The modal should only have a question `Are you Sure`? right now, there is a heading portal.says, replace it with the confirmation question. 
2. Clicking upon yes, the book doesn't get deleted, please check
3. Also, the PR has merged conflicts in it, please fix them
image

P4NK4J (changes requested):
@akriti12negi The modal is not disappearing after a book is deleted. Also, the modal title should not be `confirmation question` itself. Simply, remove the modal header and footer. Keep the buttons and the `Are you sure` question in the body itself. Let me know if that makes sense.
image

PR #1717, opened 2022-07-05
Disable save button in add category popup on guidelines page 

shardulgusain:
#1713

### Description
Disabled save button in add category popup on guidelines page.
Added a function in category.modal.blade.php file that is checking whether the text in the form tab is empty or not and enabling/disabling the button accordingly .

### Checklist:

- [x] I have performed a self-review of my own code.

PR #1806, opened 2022-07-12
Disabled save button in add category popup on guidelines page

shardulgusain:
#1717 

### Description
Implemented a function that will enable the save button only if user enter something on the form tab of add category window inside Guidelines and Resources option.

### Checklist:

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -956,7 +956,7 @@ $(document).ready(function() {
 	  $(document).on("change", ".send-mail-to-applicant", toggleApplicantMailEditor);
 });
 
-document.getElementById("name").onkeyup = function(){ButtonEnable()};;
+$(document).getElementById("name").onkeyup = function(){ButtonEnable()};
 
 function ButtonEnable() {
 	if(document.getElementById("name").value=="") { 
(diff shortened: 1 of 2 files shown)


resources/js/app.js
```diff
@@ -955,6 +955,16 @@ $(document).ready(function() {
 	  $(document).on("click", ".toggle-block-display", toggleBlockDisplay);
 	  $(document).on("change", ".send-mail-to-applicant", toggleApplicantMailEditor);
 });
+
+$(document).getElementById("name").onkeyup = ButtonEnable();
```
GauravGusain98, line 959: The name of the id looks a bit generic. By looking at the name it's not getting clear, which element is this or what this element do? 

Please update the id name to a more specific name.

GauravGusain98: Maybe `categoryName` would be a better name.

GauravGusain98 (changes requested):
Please check the comment.

resources/js/app.js
shardulgusain: Ok. Done.

GauravGusain98 (approved):
Looks fine

Issue #1789, opened 2022-07-09
On the HR recruitment reports page, add a count of total verified applications


P4NK4J:
**Describe the task**
On the HR recruitment page, we have a count of applications appearing on daily basis. We would like to have a label next to Analytics heading, named `total verified applications - |*COUNT*|`

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/35001775/178102058-0465c784-243f-40a2-873d-642f8ddcf770.png">

LavishRawat:
@P4NK4J Do I only have to add Label (total verified applications - |*COUNT*|) or a count function next to Total Verified Applications?

P4NK4J:
@LavishRawat Let's count the applications only those who are added after July 6th.

Ayush8923:
This issue is completed. Closing it.

PR #1810, opened 2022-07-12
add-a-count-of-total-verified-applications

LavishRawat:
Targets #1789 

### Description
On the HR recruitment reports page, add a count of total verified applications 

### Checklist:


resources/views/hr/recruitment/reports.blade.php
```diff
@@ -5,7 +5,8 @@
 <div class="chart-container" align="center">
     <div class="bar-chart-container">
         <div align="left">
-         <h2>Analytics </h2>  
+         <span class="chart-heading mr-1">Analytics</span>
+         <span class="total-chart">Total Verified Applications-|*COUNT*|</span>
```
P4NK4J, line 9: |*COUNT*| is a method to clarify that is a dynamic section, where the total applications count will come from database. Please correct this.

P4NK4J (changes requested):
@LavishRawat After these changes, the UI of applications report count is distorted. Also, when you select the date from datepicker inputs and click on view, its throwing an error
image
image

The correct UI of date counter graph should be as per shared in the screenshot.
image

Modules/HR/Http/Controllers/Recruitment/ReportsController.php
```diff
@@ -30,14 +31,26 @@ public function index()
 
         $data = [];
 
+        $record1 = Application::select(
+            \DB::raw('is_verified')
+        )
+        ->get();
```
kuldeep3, line 37: Can we create query that fetches only those applications where is_verified is set to 1 🤔 ? That will reduce the below for-loop

Modules/HR/Http/Controllers/Recruitment/ReportsController.php
```diff
@@ -58,13 +71,25 @@ public function searchBydate(Request $req)
...
+
+        $countIsVerified = 0;
+        foreach ($record1 as $row) {
```
kuldeep3, line 80: Same as previous feedback. Moreover, we can create a function for it and then call that function here to avoid duplicacy.

resources/views/hr/recruitment/reports.blade.php
```diff
@@ -4,20 +4,17 @@
...
-                <form action="{{route('recruitment.report')}}" method="POST" align="right">
-                    {{csrf_field()}}
-                    <input type="date" name="report_start_date" id='StartDate'> to
-                    <input type="date" name="report_end_date" id="EndDate">
-                    <input type="submit" class="btn btn-sm btn-primary text-white" value="View">
-                </form>
-                <br>
-            </div>
+        <div align="left">
+        <span class="chart-heading mr-5">Analytics</span><span class="total-chart">Total Verified Applications:<?php echo $countIsVerified ?></span>
```
kuldeep3, line 8: Indentation

kuldeep3 (changes requested):

P4NK4J (commented):
@LavishRawat I have added a minor comment on the issue as well. Right now, we are querying all the applications, we need to check applications created after July 6th. I missed mentioning it in the initial requirements, if that can be done now, that would be great. Because the verification functionality was on that date only, applications before that, won't be verified, and querying them would be an unnecessary performance burden.
Let me know your thoughts.

LavishRawat:
okay , I will try to do that.

Modules/HR/Http/Controllers/Recruitment/ReportsController.php
```diff
@@ -67,4 +70,13 @@ public function searchBydate(Request $req)
 
         return view('hr.recruitment.reports', $data, compact('todayCount'));
     }
+
+    private function verifiedApplications()
+    {
+        $from = date('2022-07-06');
```
kuldeep3, line 76: Can we pass this date from config instead of hardcoding here?

Modules/HR/Http/Controllers/Recruitment/ReportsController.php
```diff
@@ -30,14 +31,16 @@ public function index()
 
         $data = [];
 
+        $verifiedApplicationCount = $this->verifiedApplications();
```
kuldeep3, line 34: If we are only getting count of verified applications here, let's rename the function to getVerifiedApplicationsCount()

kuldeep3 (approved):

P4NK4J (approved):
Looks good.

PR #1820, opened 2022-07-13
Parse invoice payment message

GauravGusain98:
### Changes

-  Added functionality to parse messages that we received in mobile when a client payment is processed. 
- For Indian client the template of a message is different from international client.
- Fixes naming convention of ids


P4NK4J (approved):
Looks good.

PR #1824, opened 2022-07-13
Follow up email template 

P4NK4J:
Checked in some cleanups with a follow-up email template blade file.


GauravGusain98 (approved):
Looks fine

PR #1830, opened 2022-07-13
1794 title input alignment hr reporting

chetanbhatt10:
Targets #{1794}

### Description
changed alignment on title and date in hr reporting 

### Checklist:

### DIFF
--- a/resources/sass/app.scss
+++ b/resources/sass/app.scss
@@ -832,7 +832,6 @@ span.total-chart-count {
   -webkit-transform: translate(-50%, -50%);
   transform: translate(-50%, -50%);
 }
-
 // HR module report analytics ends
 // End Custom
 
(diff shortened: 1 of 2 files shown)


resources/sass/app.scss
```diff
@@ -1,796 +1,859 @@
...
+}
+.chart-container .bar-chart-container .report-date-filter {
+  float: right;
+}
+.card.report-detail-section {
+  clear: both;
+  margin-top: 20px;
+  float: left;
+  width: 100%;
+}
```
P4NK4J, line 847: Let's use Bootstrap classes wherever it is possible. We are loading bootstrap files in the browser, so adding already created classes in the SCSS file would increase the compiled file size, that may impact performance a bit.

P4NK4J (changes requested):
Left a comment, please check.

P4NK4J (changes requested):
@chetanbhatt10 Let's add some space between Analytics and the datepicker inputs, from top and bottom as well. Other than that, please change the background of the reporting container back to white
image

Issue #1816, opened 2022-07-12
Project details page should have a refresh button to sync latest efforts for the project


deepankar-sajwan:
**Describe the task**
In current implemntation the efforts for projects are synced from Efforthseet available in google drive to portal at 10pm everyday. In case the  any team member updates the efforts, after 10 pm then efforts are not reflected for that team member in the portal.   

Steps to reach  project details page

Go to CRM and click on the Projects
Click on any Project

**Expected behavior**

- A refresh button should be available in the Project details page, so that updated efforts for any project are available.

- The font size and background colour of refresh button should be same as Edit button

- Please refer screenshot for reference for refresh button 

![image](https://user-images.githubusercontent.com/96309135/178505706-41aac1b5-5e92-4ee9-99d1-0ad80a9b207f.png)

**References**
Specify any reference material / links

deepankar-sajwan:
The task is complete refere screenshot

![image](https://user-images.githubusercontent.com/96309135/180228966-1d410430-8fcf-444b-b5ff-ea9c70ae9140.png)

![image](https://user-images.githubusercontent.com/96309135/180229274-73b4ab7f-cdb3-4c74-a198-ddd9c8ed4a4a.png)

PR #1854, opened 2022-07-14
Added a refresh button to sync latest efforts for the projects in pro…

Bahugunajii:
### Target
#1816 

### Description
- Added a refresh button for individual projects in order to sync the latest efforts for the project in the project details page

### Checklist:


kuldeep3 (commented):
Any reason to create a new function called refreshEfforts?

I think right now there is a cron that runs twice a day and triggers a function that fetches the efforts for the project. Can't we utilise the same function on click of the button?

Moreover, on the function naming..it should be getEffortForProject where project would be an argument passed to the function.

GauravGusain98:
> Any reason to create a new function called refreshEfforts?
> 
> I think right now there is a cron that runs twice a day and triggers a function that fetches the efforts for the project. Can't we utilise the same function on click of the button?
> 
> Moreover, on the function naming..it should be getEffortForProject where project would be an argument passed to the function.

@kuldeep3 We have removed the code from the cron and made this function that will get effort of the project. We have then used this function in the cron

kuldeep3 (approved):

PR #1879, opened 2022-07-16
sort clients by name

parasnegi2002:
Targets #{1861}

### Description
 clients  name in ascending order
### Checklist:

### DIFF
--- a/Modules/Project/Services/ProjectService.php
+++ b/Modules/Project/Services/ProjectService.php
@@ -117,7 +117,7 @@ class ProjectService implements ProjectServiceContract
 
     public function getClients()
     {
-        return Client::where('status', 'active')->get();
+        return Client::orderBy('name')->where('status', 'active')->get();
     }
 
     public function getTeamMembers()
--- a/Modules/Project/Services/ProjectService.php
+++ b/Modules/Project/Services/ProjectService.php
@@ -117,7 +117,7 @@ class ProjectService implements ProjectServiceContract
 
     public function getClients()
     {
-        return Client::orderBy('name')->where('status', 'active')->get();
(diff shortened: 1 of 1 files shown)


Modules/Project/Services/ProjectService.php
```diff
@@ -117,7 +117,7 @@ public function store($data)
 
     public function getClients()
     {
-        return Client::where('status', 'active')->get();
+        return Client::orderBy('name')->where('status', 'active')->get();
```
P4NK4J, line 120: Please refer conventional laravel query, I think where clause should come before `orderBy`. If we translate it into SQL query, then selection is made based on a condition and then ordering is done. Let me know your thoughts?

P4NK4J (commented):
Left a question for you, please take a look.

Modules/Project/Services/ProjectService.php
parasnegi2002: sir there is the difference between sortBy and orderBy.
#sortBy = we sort / arrange the fetched data in ascending order according to the "id" of the data using sortByAcse("id").

#orderBy = if we use "orderBy()" rather than "sortBy()"
we sort / arrange the data before fetching it in descending order according to the "id" of the data using orderBy('id','desc'.)

P4NK4J: @parasnegi2002 Thanks for sharing the details, my question is shouldn't the query structure be `
`return Client::where('status', 'active')->orderBy('name')->get();`

With conventional structure we order the data once a condition is met, let me know what you think?

P4NK4J (approved):
Looks good.

Issue #1848, opened 2022-07-14
Team member available in the project are removed, when project status is changed from Active to Halted


deepankar-sajwan:
**Describe the bug**
When a project is started ,its status is Active, team members are available in the Project. In case if we change status of project from Active to Halted then team members are not available in project detail page.

**To Reproduce**
1. Go to CRM
2. Click on Projects, then click on Add new project
3. Fill form and select status as Active, then click on Save button 
4. Click on the project created in step 3
5. Click on  Edit button and then click on Project Team members tabs
6. Click on Add new team member
7.Enter name ,designation and efforts for team members
8. Click on Save button
9. Go to project details tab and change status to halted, then click on save button
10. Click on <- back icon 

**Expected behavior**
The team member should not be removed from project details, if status of project is updated from Active to Halted

**Screenshots**
![image](https://user-images.githubusercontent.com/96309135/178958324-c62680c5-74ef-4f10-910f-b3c9e7120c07.png)

![image](https://user-images.githubusercontent.com/96309135/178958371-8d0359e8-e1d5-4602-8e10-2c7d918ff7ae.png)

![image](https://user-images.githubusercontent.com/96309135/178958421-cac79395-3d8f-4568-a580-48493605baeb.png)

![image](https://user-images.githubusercontent.com/96309135/178958468-d2043bcb-69c5-417d-954e-2dce7087b75b.png)

![image](https://user-images.githubusercontent.com/96309135/179181334-000dfa28-1f25-4c23-a229-7d4fc693168b.png)

Kirannegi02:
@amanbhatt2022 I am working on this issue.

deepankar-sajwan:
Task is complete

Refer screenshot

![image](https://user-images.githubusercontent.com/96309135/180796354-0997a0b4-6a93-4100-b51b-cd2c3ab2a84a.png)


![image](https://user-images.githubusercontent.com/96309135/180796427-d360a011-a643-4bab-9538-48eadd7d545c.png)


![image](https://user-images.githubusercontent.com/96309135/180796534-3a009bf0-9f59-4305-b7b7-d27a315dd4cf.png)

PR #1892, opened 2022-07-17
Showing the team members in halted project

Kirannegi02:
Targets #1848

### Description
- Showing all the team members in halted projects.

### Checklist:


Modules/Project/Entities/Project.php
```diff
@@ -47,12 +47,12 @@ public function tasks()
 
     public function getTeamMembers()
     {
-        return $this->hasMany(ProjectTeamMember::class)->whereNULL('ended_on');
+        return $this->hasMany(ProjectTeamMember::class)->whereNotNULL('ended_on');
```
GauravGusain98, line 50: Nothing should be changed in this file. We need  only those team members which are currently assigned to the project

GauravGusain98 (changes requested):
Please check the comment and make the changes as per the discussion held offline

Modules/Project/Services/ProjectService.php
```diff
@@ -189,7 +189,7 @@ private function updateProjectDetails($data, $project)
         if ($data['status'] == 'active') {
             $project->client->update(['status' => 'active']);
         } else {
-            if (! $project->client->projects()->where('status', 'active')->exists()) {
+            if (! $project->client->projects()->whereIn('status', ['active', 'halted'])->exists()) {
```
GauravGusain98, line 192: This does not look correct as in any case the team members are getting unassigned

GauravGusain98 (changes requested):
Please check the comments

Modules/Project/Services/ProjectService.php
```diff
@@ -186,8 +186,8 @@ private function updateProjectDetails($data, $project)
         }
 
         $this->saveOrUpdateProjectContract($data, $project);
-        if ($data['status'] == 'active') {
-            $project->client->update(['status' => 'active']);
+        if (in_array($data['status'], ['active', 'halted'])) {
+            $project->client->update(['status' => $data['status']]);
```
GauravGusain98, line 190: This will mark the client as halted in case `data['status']` is halted but we don't have any halted status for the client. Maybe you can mark the client as active in this case

GauravGusain98 (changes requested):
Please check the comment

GauravGusain98 (approved):
Looks fine

Issue #1772, opened 2022-07-08
Error in moving on-hold applications to resume screening


P4NK4J:
**Describe the task**
When we move any application from **on-hold to resume screening**, its status is changed as an in-progress application in the database. While it should have a `new` status and `new application` tag associated. This fix is required only if we move any application from `on-hold` to `resume screening`, which was never evaluated. For the applications whose resume screening is done, they should not be part of this feature.

To solve this issue you can follow these subtasks.

- [x]  Find the application listing page where the `on-hold` application is shown.
- [x]  Find the route file(web.php) in which the route is defined for when you move the **on-hold application** to resume screening.
- [x]  From the route file(web.php) you get the function for the particular controller.
- [x]  In this function you need to make a logic for if the application status is `on-hold` then put the `new` status. You can refer to the application functions where the status is being changed.

PR #1894, opened 2022-07-18
Feature/1772/fixing new status

ASHISH-KOHLI:
#1772

### Description
new application tag is associated when an application is sent form on hold to resume screening 

### Checklist:

### DIFF
--- a/Modules/HR/Entities/ApplicationRound.php
+++ b/Modules/HR/Entities/ApplicationRound.php
@@ -66,7 +66,8 @@ class ApplicationRound extends Model
                 if ($application->status=='on-hold') {
                     $application->untag('on-hold');
                     $application->tag('new-application');
-                }else{
+                }
+                else{
                     $application->untag('new-application');
                     $application->tag('in-progress');
                 }
--- a/Modules/HR/Entities/ApplicationRound.php
+++ b/Modules/HR/Entities/ApplicationRound.php
@@ -63,8 +63,8 @@ class ApplicationRound extends Model
                 break;
 
             case 'confirm':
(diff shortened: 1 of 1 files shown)


P4NK4J (changes requested):
@ASHISH-KOHLI This is not as per the requirements. I moved an on-hold application to Resume Screening round, and its status is still showing in-progress, and awaiting confirmation. The application status should be `new` and it should have corresponding tag.

ASHISH-KOHLI:
@P4NK4J okay i will fix it

P4NK4J:
@ASHISH-KOHLI I tested with an application, it seems that moving it to Resume Screening is not correct. Right now, its showing accepted in Resume screening, and then again in Resume screening application is available for evaluation. 
The workflow should be, if an application is in `on-hold`, and from there if its moved to Resume screening, it should appear as a fresh application, like shared in the second screenshot. 
You need to change the status from `on-hold` to `new` in the database and use tags correspondingly
Screenshot 2022-07-19 at 11 38 06 AM

image

P4NK4J (changes requested):
Left some feedback, please check.

viveknegi84:
Sorry for that.😬

P4NK4J (changes requested):
@ASHISH-KOHLI While I am moving an application, its not available for evaluation in resume screening, and in the timeline it is showing that resume screening has already conducted. Please take a look and find how the data is present for new applications, we need to replicate the same in these cases as well (only if an application is moved to resume screening from on hold)
Feel free to reach out if you have any questions.
image

P4NK4J (approved):
Looks good

Issue #1624, opened 2022-06-25
Adding new Team member functionality should be disabled for Inactive Project and Halted Project


deepankar-sajwan:
**Describe the task**
In current implementation for Inactive Project, Halted Project the Adding new Team member functionality is available. The Inactive project and Halted Project should not have functionality to add new team member.

Refer screenshot for Halted Project

![image](https://user-images.githubusercontent.com/96309135/178756450-9204d052-d168-440b-a5c6-eafe69ddf934.png)

Click on Edit button and click on Team members Tab

![image](https://user-images.githubusercontent.com/96309135/178760352-cbfbb6f1-8f7e-4b5b-9c9b-8184d8dce6d5.png)

**Expected behavior**
Disable the Adding new Team member functionality from front end for Inactive project and Halted Project

Ayush8923:
This issue PR is merged. Closing it.

PR #1903, opened 2022-07-19
disabled add new team member functionality

Simmyrwt:
Targets #1624

### Description

disable add new team member functionality in inactive and halted project
### Checklist:

### DIFF
--- a/Modules/Project/Resources/views/subviews/edit-project-team-members.blade.php
+++ b/Modules/Project/Resources/views/subviews/edit-project-team-members.blade.php
@@ -68,9 +68,13 @@
                     </div>
                 </div>
 
+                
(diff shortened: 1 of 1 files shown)


Modules/Project/Resources/views/subviews/edit-project-team-members.blade.php
```diff
@@ -68,9 +68,13 @@
                     </div>
                 </div>
 
+                
+                @if((request()->get('projects') == 'inactive-projects') || (request()->get('projects') == 'halted-projects'))
```
GauravGusain98, line 72: This is not correct. We are not getting any parameters with name projects in the request. You need to fetch the status of project from DB instead of finding the status in request.

GauravGusain98 (changes requested):
Please correct the logic

Ayush8923:
Closing it because no one is working on it.

Issue #1554, opened 2022-06-14
Segment name in Manage Evaluation to be mapped with Evaluation Round


P4NK4J:
**Describe the task**
Currently while creating a new evaluation segment, we don't have the facility to map it with the application round. 

**Expected behavior**
Add a dropdown having all the application rounds when a segment is created in the `create segment` modal. If a user selects the round corresponding to the segment, it would be mapped in the database automatically. Refer to the screenshot below for the modal
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/35001775/173597241-957441ec-25ee-44d8-b17a-7fdcb3044f3e.png">

deepankar-sajwan:
The task is complete please refer screenshot


![image](https://user-images.githubusercontent.com/96309135/180226997-0b39de44-30f1-4c74-ae86-e4f062cb6e41.png)


![image](https://user-images.githubusercontent.com/96309135/180227037-45d9fcd4-6d18-4e22-a5b7-65ef8b140fe2.png)

PR #1921, opened 2022-07-20
Evaluation round mapped

NirmalNegi07:
#1554

### Description
There was no dropdown option of rounds in create segment form, so i add that facility to it. Also, now hr_rounds table is mapped with evaluation_segment table. 

### Checklist:

### DIFF
--- a/Modules/HR/Http/Controllers/EvaluationController.php
+++ b/Modules/HR/Http/Controllers/EvaluationController.php
@@ -53,10 +53,10 @@ class EvaluationController extends Controller
      */
     public function createSegment(Request $request)
     {
-        $segment_id = Round::select('*')->where('name', $request->rounds)->first()->id;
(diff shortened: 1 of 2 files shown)


Modules/HR/Http/Controllers/EvaluationController.php
```diff
@@ -50,7 +53,11 @@ public function segmentParameters(Request $request, $segmentID)
      */
     public function createSegment(Request $request)
     {
-        $segment = Segment::create(['name' => $request->name]);
+        $segment_id = Round::select('*')->where('name', $request->rounds)->first()->id;
```
P4NK4J, line 56:
```suggestion
        $segmentId = Round::select('*')->where('name', $request->rounds)->first()->id;
```

P4NK4J (approved):
Left a minor suggestion, else looks good.

Issue #1898, opened 2022-07-18
The filter for checking the application count in Analytic page is not working as per specification


deepankar-sajwan:
**Describe the task**
The filter used to check the number of application received  between duration of days, does not include information of end date.

Please refer screenshot

The data available from july 11th  till July 18th
![image](https://user-images.githubusercontent.com/96309135/179502803-09bdb021-b4ed-45ce-8420-6ff8f080912e.png)

Selected  date  from july 11th  till July 16th
![image](https://user-images.githubusercontent.com/96309135/179502853-2bb90006-2625-4599-8c0f-f001bb9f26bd.png)

The data available from july 11th  till July 15th July
![image](https://user-images.githubusercontent.com/96309135/179502898-90437c5c-bc9b-49ef-bf55-3a97ca7fb527.png)

Data for 16th July is not available

**Expected behavior**
A clear and concise description of what needs to be done.

deepankar-sajwan:
The  issue is fixed please rere screenshot

![image](https://user-images.githubusercontent.com/96309135/180245577-f61b7b55-7e2d-4b53-898c-6193b87ea13d.png)

![image](https://user-images.githubusercontent.com/96309135/180245750-dd1d4d38-e5c9-4396-ae23-802050a67d78.png)

PR #1925, opened 2022-07-20
the-filter-for-application-count-is-not-working

negirishabh28:
Targets #1898 

### Description
The filter for checking the application count in Analytic page is not working as per specification

### Checklist:


P4NK4J (approved):
Looks good.

Issue #1836, opened 2022-07-13
Pagination in Project listing page


deepankar-sajwan:
**Describe the task**
As the number of projects increase the projects displayed in the project listing page will also increase. 
This result in slow page loading, as the number of records are huge. 

Refer screenshot for project listing page

![image](https://user-images.githubusercontent.com/96309135/178799360-4e37c6c8-343e-4cd2-b8f1-9dd7dd945804.png)

**Expected behavior**
Pagination should be applied in the project listing page, so that it results in faster page loading. Display 10 records in each page.

**References**
Specify any reference material / links

Abhishek25081999:
@deepankar-sajwan Sir as I have to paginate the inactive and haulted project pages but I'm unable to change the status of an active project into haulted or inactive project.

Abhishek25081999:
@deepankar-sajwan Sorry it's Halted project instead of Haulted.

PR #1930, opened 2022-07-21
Pagination in Project listing page

Abhishek25081999:
Targets #1836 

### Description
Pagination should be applied in the project listing page, so that it results in faster page loading. Display 10 records in each page.

### Checklist:

### DIFF
--- a/Modules/Project/Resources/views/index.blade.php
+++ b/Modules/Project/Resources/views/index.blade.php
@@ -109,5 +109,6 @@
             </tbody>
         </table>
     </div>
+    {{ $clients->links()}}
 </div>
 @endsection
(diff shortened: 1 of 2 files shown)


Modules/Project/Resources/views/index.blade.php
```diff
@@ -109,5 +109,6 @@
             </tbody>
         </table>
     </div>
+    {{ $clients->links()}}
```
P4NK4J, line 112: Why are we using this ?

P4NK4J (commented):
left a question, please check.

Modules/Project/Resources/views/index.blade.php
Abhishek25081999: sir, we are using this link to show the pagination, which is already set in config. file.

P4NK4J: Why we are using `$clients` for project listing pagination, can we make it `$project->links()` instead ?

P4NK4J (commented):
Left a suggestion, please check.

P4NK4J (approved):
Based on our offline conversation, I got to know that we are using client object to display 10 records of client and all the projects associated with that. Hence approving the PR.

Issue #1831, opened 2022-07-13
Bulk upload applications in HR jobs


P4NK4J:
**Is your feature request related to a problem? Please describe.**
Currently, we have an application mechanism in the portal for the website only. We would require functionality to upload the applications from an excel sheet.

**Describe the solution you'd like**
We can give a bulk upload feature, where the applications can be uploaded from an excel sheet. There is already an existing implementation created (start researching with this [route](https://github.com/ColoredCow/portal/blob/master/Modules/HR/Routes/web.php#:~:text=Route%3A%3Apost(%27/excel%2Dimport%27%2C%20%27ApplicantController%40importExcel%27)%2D%3Ename(%27hr.applications.excel%2Dimport%27)%3B)). We will start with tweaking the particular section.

**Dev process to follow:**
1. Figure out a way to run the current implementation, and see to which extent it is working
2. Once we get an idea of the code changes we need to do, create issues for fixing the uploading of applications, and what's working and what's not.  **BREAKING IT DOWN IS A MUST**
3. After the applications are created, and proper mapping is done in the database, we need to figure out the best way to upload the resume corresponding to each application. We do have an existing functionality that moves portal resumes to CC websites in WordPress. We need to check if moving the resumes to S3 would be the best option (yet to be researched)
4. Bulk uploaded applications will have channels specified (database structure would be completed after issue #1818  is resolved
5. Post bulk upload, we will also include a telephonic interview round in the evaluation
6. Extensive testing would be required whether the application is created as per the website flow.
7. Post this import functionality, we will implement a telephonic interview round for this

**You are not alone**
- Reach out for help on the portal hangout channel, to get help in understanding the workflow, breakdown of tasks 

**GitHub workflow**
It would be a huge functionality, hence we will do it in a feature branch
1. Create a `feature/bulk-import-applications` branch, this branch would be like the `master` branch of the portal.
9. Breakdown the bulk upload into different phases and each phase would have a PR which upon review will be merged into  `feature/bulk-import-applications`, the main branch would then be merged into master, once the functionality is completed.
10. We will create different issues for breakdown items.

Bahugunajii:
The issue stated above could be broken down into the following phases:
P1: Add an import excel button beside the `Add new application` button.
P2: On clicking the button, a modal should appear containing mandatory fields like `job` and `upload excel file` 
P3: Once the import is started, all applications should be mapped in the `hr_applications` table
[...]

PR #1934, opened 2022-07-21
Added a bulk applications upload feature

Bahugunajii:
### Targets
 #1831 

### Description
- A new feature is being added where one can upload applications in bulk from excel
- An additional round named "Telephonic Interview" is also being added
- Application Evaluation is also possible corresponding to this particular round

### Checklist:


P4NK4J (changes requested):
A couple of testing feedbacks
1. Update sample excel file for correct data format
2. MoveFilesToWordPress.php, we don’t need to call that when we are using bulk import
3. The following error appears when an application is directly moved from Resume Screening to Telephonic interview. It seems that it looks out for marks corresponding to round 1 as well.
image

Modules/HR/Http/Controllers/EvaluationController.php
```diff
@@ -180,10 +180,18 @@ public function update($applicationRoundId)
...
-        return Segment::whereHas('round')
-            ->orWhereNull('round_id')
+        $query = null;
+        if ($roundId === 14) {
+            $query = Segment::whereHas('round')
+            ->where('round_id', $roundId);
+        } else {
+            $query = Segment::whereHas('round')
+            ->where('round_id', '!=', 14);
+        }
```
P4NK4J, line 192: Let's not use round ID here, it may vary for different databases. Instead, can we check it with Round Name? Because Round Names can be updated from the database if are inconsistent, but ID will create issues with the other rows as well ?

Modules/HR/Resources/views/application/render-application-row.blade.php
```diff
@@ -53,79 +54,80 @@
...
+                    @endif
+                    <span>
+                        {{ $tag->name }}
 
-				@if($tag->slug == 'need-follow-up' && $attampt =
-				optional($application->latestApplicationRound->followUps)->count())
-				. attempt: {{$attampt}}
-				@endif
+                        @if ($tag->slug == 'need-follow-up' &&
+                            ($attampt = optional($application->latestApplicationRound->followUps)->count()))
```
P4NK4J, line 126: Let's rename this to $attempt

P4NK4J (changes requested):
Minor suggestions, else looks good.

Modules/HR/Config/config.php
```diff
@@ -3,7 +3,8 @@
 return [
     'name' => 'HR',
     'applicationEvaluation' => [
-        'cutoffScore' => 6,
+        'cutoffScore-resume-screening' => 6,
+        'cutoffScore-telephonic-interview' => 5,
```
GauravGusain98, line 7: Let's follow one convention only
```suggestion
        'cut-off-score-resume-screening' => 6,
        'cut-off-score-telephonic-interview' => 5,
```

Please fix this at other places as well

resources/views/hr/application/evaluation-form.blade.php
```diff
@@ -21,42 +21,51 @@
         </div>
     </div> --}}
     {{-- this depends on the application round. currently hard-coded for resume-screening --}}
-    <div class="evaluation-stage" id="evaluationStage1">
-        @include('hr::evaluation.evaluation-form.resume-screening.feeling')
+    @if ($applicationRound->hr_round_id == 14)
```
GauravGusain98, line 24: Lets not harcode the id of a round as it may change in future. Maybe you can use the round name instead

GauravGusain98 (changes requested):
Minor suggestions, else looks fine.

P4NK4J (approved):
Looks good.

GauravGusain98 (approved):
looks good

Issue #1940, opened 2022-07-21
Create functionality to store the content of the email  sent to client requesting for payment  realted to project work


deepankar-sajwan:
Finance department sent email realted to invoice to client in following scenario
Requesting payment for a project work, done by the company for the client

At present there is no implementation for storing  content sent to client, for scenario mentioned above.

Ayush8923:
This issue PR is merged. Closing it.

PR #1963, opened 2022-07-22
Feature/store emails in invoice mails table

amisha296:
Targets #1940 

### Description
I have stored invoices in the invoice_mails table for type invoice.

### Checklist:


Modules/Invoice/Entities/Invoice.php
```diff
@@ -200,4 +200,10 @@ public function confirmationMail()
         return $this->invoiceMail()->where('type', config('invoice.mail-type.payment-confirmation.slug'))
         ->first();
     }
+
+    public function invoiceOfInvoiceTypeMail()
```
P4NK4J, line 204: Can we rename it to `getInvoiceTypeMail` ? Currently, it seems like we are fetching an invoice from invoice type email, while with the function definition it seems to be fetching the Invoice email only

P4NK4J (commented):
Left a suggestion, please check.

GauravGusain98:
Merging in the base branch. Will review together

Issue #1953, opened 2022-07-22
Change color of total count text and icon


P4NK4J:
**Describe the task**
We need to change the color of the total count button and its icon. We need to match it with the same color as others

<img width="1307" alt="image" src="https://user-images.githubusercontent.com/35001775/180368708-d9033344-6765-40d7-b530-ecfff516fb9d.png">

deepankar-sajwan:
The task is complete please refer screenshot

![image](https://user-images.githubusercontent.com/96309135/180494178-378fa019-3d42-41f6-b6aa-c383ee738743.png)

PR #1965, opened 2022-07-22
Created a div to make all the text and icon color identical in the menu

viveknegi84:
Targets #1953 

### Description
Changes include creation of a new div for the hr menu so that all the text and icons in the menu are identical in color.

### DIFF
--- a/resources/views/hr/menu.blade.php
+++ b/resources/views/hr/menu.blade.php
@@ -1,3 +1,4 @@
+<div class ="hr-menu">
 <ul class="nav nav-pills">
     <li class="nav-item">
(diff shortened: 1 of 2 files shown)


resources/views/hr/menu.blade.php
```diff
@@ -1,3 +1,4 @@
+<div class ="hr-menu">
```
P4NK4J, line 1: Is it possible doing it with `text-color` class of bootstrap ? defining a similar class and then adding it to app.scss is an additional step. Please check if its possible to use text color class of bootstrap.

P4NK4J (commented):
left a comment, please check.

P4NK4J (approved):

Issue #1635, opened 2022-06-28
Single invoice for all the  projects associated with the client


deepankar-sajwan:
**Describe the task**
In case client request single invoice for different project. For example Client has two projects  Project A and Project B, and client request single invoice including both Project A and Project B

**Expected behavior**
Single invoice  should be available for all the projects associated with the client

PR #1984, opened 2022-07-23
Project Level Invoice Generation

GauravGusain98:
Targets #1635 

### Description

- Functionality to Show client level and project level ready to send Invoices
- Functionality to create project level billing invoice

### Checklist:


Modules/Invoice/Resources/views/index.blade.php
```diff
@@ -148,26 +162,30 @@
...
-                                <td>{{ $loop->iteration }}</td>
+                                <td>{{ $index }}</td>
                                 <td>
                                     {{ $client->name . ' Projects' }}
                                 </td>
                                 <td>{{ config('constants.currency.' . $client->currency . '.symbol') . '' . $client->billingDetails->service_rates . __('/hour')}}</td>
-                                <td>{{ $client->getClientLevelProjectsBillableHoursForTerm($month, $year) }}</td>
+                                <td>
+                                    {{ $client->getClientLevelProjectsBillableHoursForInvoice() }}
+                                    {{-- <i class="pt-1 ml-1 fa fa-external-link-square" data-toggle="modal" data-target="#InvoiceDetailsForClient{{ $client->id }}"></i> --}}
```
GauravGusain98, line 183: Commented as this feature is only built for client level projects

Modules/Client/Entities/Client.php
```diff
@@ -208,68 +220,112 @@ public function getEffortSheetUrlAttribute()
         }
     }
 
-    public function getClientMonthStartDateAttribute()
+    public function getClientMonthStartDateAttribute($monthsToSubtract)
```
rathorevaibhav, line 223:
```suggestion
    public function getMonthStartDateAttribute($monthsToSubtract)
```

Modules/Client/Entities/Client.php
```diff
@@ -208,68 +220,112 @@ public function getEffortSheetUrlAttribute()
...
     {
+        $monthsToSubtract = $monthsToSubtract ?? 0;
```
rathorevaibhav, line 225: why default value is zero here and `1` at other places?

Also, let's remove this line and set default value at function param level.

Modules/Client/Entities/Client.php
```diff
@@ -208,68 +220,112 @@ public function getEffortSheetUrlAttribute()
...
-
-                return today(config('constants.timezone.indian'))->subMonth()->addDays($billingDate - today(config('constants.timezone.indian'))->day);
-            }
+            return now(config('constants.timezone.indian'))->subMonthsNoOverflow($monthsToSubtract)->startOfMonth();
+        }
 
-            if (today(config('constants.timezone.indian'))->day >= $billingDate) {
-                return today(config('constants.timezone.indian'))->startOfMonth()->addDays($billingDate - 1);
+        if (today(config('constants.timezone.indian'))->day < $billingDate) {
+            if (today(config('constants.timezone.indian'))->subMonthsNoOverflow($monthsToSubtract + 1)->addDays($billingDate - today(config('constants.timezone.indian'))->day) > today(config('constants.timezone.indian'))->subMonth()->endOfMonth()) {
```
rathorevaibhav, line 233: I don't think we need to hardcode `constants.timezone.indian` anywhere if we set the APP_TIMEZONE in the env file to IST.

We would need a single timezone in the application and I think we can keep it IST. I believe we also need to ensure what is the timezone for our DB server. It should be IST as well.

What do you think?

rathorevaibhav: I just checked our servers. We have APP_TIMEZONE set to `Asia/Calcutta`.

Maybe it'll be better if you make these three changes in your PR:
1. Set the default value of app timezone in the .env.example file to asia/kolkata
2. Set the backup value in `app.php` for timezone to asia/kolkata instead of UTC
3. As a part of deployment step, mention to change OS server and DB server timezone to same as app timezone

rathorevaibhav (changes requested):

Modules/Client/Entities/Client.php
```diff
@@ -175,16 +180,23 @@ public function getNextInvoiceNumberAttribute()
...
-        return $monthEnd->diffInDaysFiltered(function (Carbon $date) {
+        return $monthEndDate->diffInDaysFiltered(function (Carbon $date) {
             return ! $date->isWeekend();
         }, $monthStartDate);
     }
 
+    public function getWorkingDays($startDate, $endDate)
+    {
+        return $endDate->addDay()->diffInDaysFiltered(function (Carbon $date) {
+            return ! $date->isWeekend();
```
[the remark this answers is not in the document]
GauravGusain98: done

Modules/Invoice/Routes/web.php
```diff
@@ -22,7 +22,6 @@
     Route::get('/{invoice}/edit', 'InvoiceController@edit')->name('invoice.edit');
     Route::post('/{invoice}/update', 'InvoiceController@update')->name('invoice.update');
     Route::delete('/{invoiceId}/delete', 'InvoiceController@destroy')->name('invoice.delete');
-    Route::post('/generate-invoice-client', 'InvoiceController@generateInvoiceForClient')->name('invoice.generate-invoice-for-client');
```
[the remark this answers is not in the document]
GauravGusain98: It is not being used anywhere

Modules/Project/Entities/Project.php
```diff
@@ -166,4 +192,25 @@ public function getLastUpdatedAtAttribute()
...
+
+    public function invoices()
+    {
+        return $this->hasMany(Invoice::class);
+    }
+
+    public function scopeInvoiceReadyToSend($query)
+    {
+        return $query->whereDoesntHave('invoices', function ($query) {
+            return $query->whereMonth('sent_on', now(config('constants.timezone.indian')))->whereYear('sent_on', now(config('constants.timezone.indian')));
```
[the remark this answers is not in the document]
GauravGusain98: They are store in IST

Modules/Client/Entities/Client.php
GauravGusain98: Its because for the client we usually take the dates for the current Month but at other places the functions are mostly related to invoice, so the default value was set to 1.

GauravGusain98: The reason for not setting the value in the parameter at most places was that when we use this accessor like `$client->month_start_date` then the value that is set to $monthToSubtract is NULL instead of 0. And many places where the default value is 1 there the result comes incorrect. 

Apart from this at some places where there are 2 parameter, we are getting error from Larastan, and as a suggestion it is removing the value from the parameter.

rathorevaibhav (approved):

Issue #1837, opened 2022-07-13
Phone no field available in client  contact person should also accept  character as input


deepankar-sajwan:
**Describe the task**
When new client is added the details realted to client contact person are added, which also includes Phone no. 
In current implemntation Phone no field only accepts number as input, also phone only acccept 10 digit phone number only, but there are some client which have code number  such as (061) 707-2HG.

Steps to reach Client Phone no field
- Go to CRM and click on clients
- Click on Add client
- Enter client name and click on Save button
- Click on the client name, mentioned in above step
- Click on Contact Persons Tab

Please refer screenshot for contact person tab

![image](https://user-images.githubusercontent.com/96309135/180715525-009702e1-da7f-4c45-93ca-57c9e8cba330.png)

**Expected behavior**
Phone no field should also accept, alphabets and symbols like -  ( )  in addition the phone no can accept  input maximum  upto 20 aphalumeric character .

deepankar-sajwan:
Provided input as 20 alphanumeric characters (1234567890abcdefghij) in Phone no field, the error is displayed clicking save and an Exit button

![image](https://user-images.githubusercontent.com/96309135/180802599-5b189686-e8f7-4291-be48-7c63fba24223.png)


![image](https://user-images.githubusercontent.com/96309135/180802885-cbf5e27c-3b2f-48af-ab8a-26968b614b96.png)

deepankar-sajwan:
Input number (616)-674-2HMG-12345 (20 alphanumeric character) accepted in Phone no field

![image](https://user-images.githubusercontent.com/96309135/181051560-8b4c2d39-7140-4fbe-ae66-146e1df307fc.png)


![image](https://user-images.githubusercontent.com/96309135/181051991-f6f41452-bdc1-4757-b941-ad3dfcfeabbf.png)

PR #1993, opened 2022-07-25
made-phone no field alphanumeric

Kirannegi02:
Targets #1837

### Description
Phone no field is accepting, alphabets and symbols like - ( ) in addition the phone no is accepting input maximum upto 20 alphanumeric character .

### Checklist:


pokhiii (approved):
lgtm

Issue #1978, opened 2022-07-23
All project back button functionality is accessible from outside the button area


deepankar-sajwan:
**Describe the task**
All Projects button is used for getting back to page which contains listing of all the projects. Currently if you click on the horizontal line lying ,along with the All project button then, you will land on page containing listing of all the projects

Refer screenshot  click on the area  highlighted by black colour

Click on project available in project listing page

![image](https://user-images.githubusercontent.com/96309135/180593610-a57729ac-175d-4181-8a5b-760582d70bb2.png)

Click on Check FTE button in project details page

![image](https://user-images.githubusercontent.com/96309135/180593624-291b8874-dc59-4411-8d68-1b59a3d11c09.png)

click on the area  highlighted by black colour

![image](https://user-images.githubusercontent.com/96309135/180593318-395738ab-7cf8-45e7-bafb-ee8c6ca70a9f.png)

Prerequiste steps  for checking <-All project functionality
- Go to CRM, click on clients 
- Click on Add client button, fill the form and create a client.
- Go to CRM, click on projects
- Click on Add new project button, fill the form and create a project.
- Click on the project created in previous step
- Click on Check FTE button available in project details button

**Expected behavior**
User should be able to land on all projects listing page only when  clicking on <- All projects

**References**
Specify any reference material / links

deepankar-sajwan:
All projects button is not accessible from outside its area

PR #1996, opened 2022-07-25
all projects listing page only when clicking on All projects

JCArya:
Targets #1978

Description
we need to show all projects in click on the All project button.


P4NK4J (approved):
Looks good.

Issue #1971, opened 2022-07-22
Salary breakdown based on gross salary


rathorevaibhav:
**Describe the task**
As we have added the gross salary field, we now need the breakdown of the monthly gross salary for a month.

Here's what to be added:
1. Basic Salary
2. HRA
3. Transport allowance
4. Other Allowance
5. Food Allowance
6. Total Salary

**Expected behavior**
- [x] These fields to be calculated based on the Salary settings
- [x] The salary settings page to be made visible for super admins under Global settings tab
- [x] The above fields should be read only and should dynamically change (reactive) as the user edits the gross salary.
- [x] If the gross salary is already set, these fields should be calculated on page load.

deepankar-sajwan:
The salary is generated as per the setting done for the salary

![image](https://user-images.githubusercontent.com/96309135/181066121-b0de9203-5cd9-47fd-9e16-b7c2c097b8c4.png)



![image](https://user-images.githubusercontent.com/96309135/181065829-91d3678c-7070-4192-b0ef-886f18254538.png)

PR #2000, opened 2022-07-25
Employee salary breakdown using salary setting page data

abhishekkhanduri:
#1971

### Description

- employee salary breakdown on the basis of salary setting data
- salary setting page created under global setting tab and only visible to Super Admins
- If the gross salary is already set then the value will be calculated at page load

### Checklist:

### Commands to run:
```sh
php artisan module:seed Salary --class=SalaryPermissionTableSeeder
```


GauravGusain98 (approved):
Looks good. Just fix the CI checks

Issue #1998, opened 2022-07-25
The project details page should diplay client name 


deepankar-sajwan:
**Describe the task**
Currently, the project details do not contain the name of the client with which the project is associated.

Please refer screenshot for the project details page

![image](https://user-images.githubusercontent.com/96309135/180775373-8e36b972-1d88-4dc6-9387-193431d64837.png)

**Expected behavior**
The project details page should contain the client name with which the project is associated. Please keep format same as other field available in project details page.

![image](https://user-images.githubusercontent.com/96309135/180775539-b558341d-7a9e-4ad4-bcb6-2926a1464ae6.png)

![image](https://user-images.githubusercontent.com/96309135/180776010-de21ede0-b062-44d0-9595-48bac1354093.png)

deepankar-sajwan:
The task is complete 

![image](https://user-images.githubusercontent.com/96309135/181049253-765789ca-1e34-4d04-a4ac-e1c2fbf9a380.png)

PR #2002, opened 2022-07-25
Display client name in project details page

Kirannegi02:
Targets #1998

### Description
- The project details page displayed client name.
-  Resolved error message in input field (phone no).

### Checklist:

### DIFF
--- a/Modules/Client/Http/Requests/ClientFormsRequest.php
+++ b/Modules/Client/Http/Requests/ClientFormsRequest.php
@@ -27,7 +27,7 @@ class ClientFormsRequest extends FormRequest
         return [
             'client_contact_persons.*.name' => 'nullable|max:120',
             'client_contact_persons.*.email' => 'required|email',
-            'client_contact_persons.*.phone' => 'nullable|max:20',
+            'client_contact_persons.*.phone' => 'nullable|digits_between:4,10|integer',
         ];
     }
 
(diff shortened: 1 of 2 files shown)


Modules/Client/Http/Requests/ClientFormsRequest.php
```diff
@@ -27,7 +27,7 @@ private function contactPersonsValidation()
         return [
             'client_contact_persons.*.name' => 'nullable|max:120',
             'client_contact_persons.*.email' => 'required|email',
-            'client_contact_persons.*.phone' => 'nullable|digits_between:4,10|integer',
+            'client_contact_persons.*.phone' => 'nullable|max:20',
```
P4NK4J, line 30: A phone number should ideally be 10 digits, I see here it is 20, let me know if I am missing something, or we need to change it to valid phone number

P4NK4J (changes requested):

Modules/Client/Http/Requests/ClientFormsRequest.php
Kirannegi02: It was mentioned in the requirement that the phone no can accept input maximum upto 20 alphanumeric character and there are some clients which have code number such as (061) 707-2HG.

GauravGusain98 (approved):
Looks fine

P4NK4J (approved):

PR #2018, opened 2022-07-26
#1971 Testing feedbacks

rathorevaibhav:
Targets #1971

### Description
1. Added Form Request to validate data. The form was failing due to the default CSRF token being added in `request->all()` which then was failing due to config check
2. Made some adjustments in UI to show the error notification

### Checklist:


GauravGusain98 (approved):
looks fine

PR #2025, opened 2022-07-27
Client policy fixes

rathorevaibhav:
### Description
1. The issue was not allowing admins with clients.update permission to see client details page
2. Updated client policy to allow users with clients.update & clients.delete permission to see clients
3. Updated client.update and key account manager check to check for only one possibility
4. Removed Key Account Manager to delete client

### Checklist:


GauravGusain98 (approved):
Looks fine

Issue #1908, opened 2022-07-19
The filter by  status functionality should display, applicantions based on the application status selected for filter


deepankar-sajwan:
**Describe the task**
Currently the functionality for  filtering the applications based on different status of the  applications not working.

Please refer screenshot

![image](https://user-images.githubusercontent.com/96309135/179725917-0268495f-e686-45a8-aefc-26b7cca48d72.png)

Procedure to create new application
1. Go to recruitment under HR
2. Click on Add new application
3. Similarly click on Internship applications
4. Click on Add new application

**Expected behavior**
On clicking apply button in filter by status,  should display the  applications based on the status selected in the filter.   On clicking Close button the filter should be reset. 
The required changes are required in Job applications and Internship applications 

**References**
Specify any reference material / links

P4NK4J:
**Action items:**

- [ ] Enable Apply button to send formData to the backend
- [ ] Send the form data to the index function of ApplicationController
- [ ] Check if the data is received in the controller
- [ ] Write a Laravel query to filter out the applications based on the status keyword
- [ ] Confirm if we are getting the filtered applications or not in the front end.

PR #2038, opened 2022-07-27
Filter Functionality

junaid1010:
Targets #1908 

### Checklist:

### DIFF
--- a/Modules/HR/Resources/views/application/index.blade.php
+++ b/Modules/HR/Resources/views/application/index.blade.php
@@ -44,14 +44,27 @@
     <div class="d-flex align-items-center justify-content-between">
         <ul class="nav nav-pills mb-2">
             <li class="nav-item">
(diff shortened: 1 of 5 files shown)


P4NK4J:
@junaid1010 A couple of things, 
1. The functionality is not working as expected, though the filters are working fine for `in progress`,  `closed`, `new`, `sent for approval`, and `approved` status, for the rest of the items it's not filtering. Also as you are filtering based on the status, other than the above mentioned are tags, which are applied based on certain conditions. So you may need to check relationships how these tags are added and filter them accordingly. So far, the work looks good, please fix the other things and we should be good to go
2. Fix the CI checks as well

P4NK4J (changes requested):
Left a comment, please check.

P4NK4J:
We are making changes around the requirements, and this would exclude most of the changes from this PR, hence closing it as of now.

Issue #1773, opened 2022-07-08
Sort By Filter in HR applications listing


P4NK4J:
**Describe the task**
Currently, in the job applications listing page under the HR module, the applications are listed as `new application` first. We would like to have a `Sort By` dropdown, next to the University dropdown. Upon clicking this dropdown, there would be two, options
```
Date
Name
```
Selecting any one of them should refresh the page and the applications should be sorted by the selected parameter.
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/35001775/177937653-1a183bce-0153-4dce-bc62-3c7ab5e9fb2a.png">

deepankar-sajwan:
The task is complete please refer screenshot
 
![image](https://user-images.githubusercontent.com/96309135/181696161-8a2282c6-3348-4e1e-9c90-8f65be22eb1a.png)

![image](https://user-images.githubusercontent.com/96309135/181696234-b1f9a9a5-d8e8-4985-9b44-dd54a834c536.png)

![image](https://user-images.githubusercontent.com/96309135/181696313-b117ab1b-ccb2-4a17-b37f-11c5480a9c89.png)


![image](https://user-images.githubusercontent.com/96309135/181696425-1e43b1aa-de8a-4ddd-b35c-0f4e27dde19c.png)

PR #2048, opened 2022-07-28
Fix/1773/sort by filter in hr applications listing

aquibmoin786:
#1773

-I add Sort by filter in hr applications listings.  There are two option one is name and another is date. If I select name then we get name_wise ascending order if select date then we get date_wise descending order


P4NK4J (approved):
Looks good.

PR #2037, opened 2022-07-27
Full salary breakdown

abhishekkhanduri:
#2026

### Description

- Complete breakdown of Salary
- Added fields in salary setting page to store health insurance amount and  employer Esi limit
- We can also add decimal values in setting page
- Health insurance amount will only be applicable incase where there will be no employee Esi or no employer Esi

### Checklist:

- [x] I have performed a self-review of my own code.

PR #2057, opened 2022-07-28
Fix replaced nan with zero

abhishekkhanduri:
#2037

### Description
Replaced NaN with zero when gross salary is not present.
### Checklist:

### DIFF
--- a/Modules/Salary/Resources/assets/js/components/SalaryBreakdown.vue
+++ b/Modules/Salary/Resources/assets/js/components/SalaryBreakdown.vue
@@ -254,6 +254,9 @@ export default {
 			return Math.min(Math.ceil(multiplier * percentage / 100) , Math.ceil(this.salaryConfigs.edli_charges_limit.fixed_amount * percentage / 100));
 		},
 		ctc() {
+			if (this.grossSalary === "") {
+				return 0;
+			}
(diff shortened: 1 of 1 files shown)


abhishekkhanduri:
image
Before Fix

abhishekkhanduri:
image
After fix

Modules/Salary/Resources/assets/js/components/SalaryBreakdown.vue
```diff
@@ -254,6 +254,9 @@ export default {
 			return Math.min(Math.ceil(multiplier * percentage / 100) , Math.ceil(this.salaryConfigs.edli_charges_limit.fixed_amount * percentage / 100));
 		},
 		ctc() {
+			if (this.grossSalary === "") {
```
rathorevaibhav, line 257: is `NaN` only showing at one place?

rathorevaibhav (approved):

Modules/Salary/Resources/assets/js/components/SalaryBreakdown.vue
abhishekkhanduri: It was showing NaN at three places, CTC, CTC Annual, CTC Aggreed and as I was using CTC in calculating other two values. Resolving CTC value worked for me in resolving other values as well.

PR #2069, opened 2022-07-30
New custom invoice template for a client

GauravGusain98:
### Description

- Added functionality to bill clients based on resources
- Added a billing engagement column that will store the percentage of the engagement per month
- Calculate the fixed budget projects based on resources
- Updated the UI for resource based invoices
- Added a custom invoice template
- Added the template mapping in config so that for a particular client the custom template should be picked
- Create new config variables for finance details
- Added ledger form that is dynamic and will store the details related to expense or debit
- Added scopes to fetch the ledger data based on the quarter number
- Added relationship b/w ledger account and client and project 
- Added functions to calculate the amount payable as per the ledger form
- Added the cast file that will decrypt the encrypted fields while converting the object to an array. This is required as the vue js was not decrypting the fields, so we need to handle it on PHP level.
- Added a helper function to covert amount to words

### Checklist:


Modules/Invoice/Config/config.php
```diff
@@ -137,6 +139,12 @@
                     ],
                 ],
             ]
+        ],
+        'invoice' => [
+            'clients' => [
+                'Tech4Dev' => 'custom-invoice-template-1'
```
P4NK4J, line 145: Can we pick it from ENV, it's a confidential information and the portal's code is open source ?

P4NK4J, line 145: Can we rename this to some specific template, not sure if 1, 2, 3 types would make more sense for readability.

Modules/Invoice/Resources/views/render/invoice-template.blade.php
```diff
@@ -302,27 +302,67 @@
...
+                    @include('invoice::render.resource-based-billing-template')
+                @else 
+                    @include('invoice::render.international-hourly-billing-template')
+                @endif
                 <div><br>
                     <table class="table" style="margin-left:33%; width:67%;">
                         <tr class="border-bottom" >
                             <td style="width: 135px;">Total</td>
-                            <td style="width: 94px;">{{ $billingLevel == 'client' ? $client->getClientLevelProjectsBillableHoursForInvoice($monthsToSubtract) : $project->getBillableHoursForMonth($monthsToSubtract) }}</td>
+                            <td style="width: 94px;">
```
P4NK4J, line 336: Is it possible to avoid inline style here ?

Modules/Invoice/Resources/views/render/invoice-template.blade.php
```diff
@@ -350,7 +404,21 @@
...
+                                        {{ $client->country->currency_symbol . $client->getTotalPayableAmountForTerm($monthsToSubtract, $projects) }}
+                                    @endif
+                                @else 
+                                    @if(optional($client->billingDetails)->service_rate_term == config('client.service-rate-terms.per_resource.slug'))
+                                        {{ $project->client->country->currency_symbol . $project->getResourceBillableAmount() }}
+                                    @else
+                                        {{ $project->client->country->currency_symbol . $project->getTotalPayableAmountForTerm($monthsToSubtract) }}
+                                    @endif     
+                                @endif
+                            </strong></td>
```
P4NK4J, line 421: It seems repetitive to me, can we make it more generic ?

P4NK4J (changes requested):
Overall looks good. Left minor suggestions.

Modules/Invoice/Resources/views/render/invoice-template.blade.php
GauravGusain98: Done

Modules/Invoice/Config/config.php
GauravGusain98: We have only one such case where invoice is custom. For now I think we can use this name only. If required we will set a format for this later.

GauravGusain98: done

P4NK4J (approved):
Looks good.

PR #2072, opened 2022-07-30
Permissions policies fixes

rathorevaibhav:
### Description
1. Fixes for several policies for permissions. Created new policies where they were missing. Mainly Employee, Finance & HR
2. Cleanups in some seeders and factories so that they produce more realistic data
3. Removed some redundant files & routes & migrations that are not being used.

### Checklist:


P4NK4J (approved):
Looks good.

PR #2077, opened 2022-07-31
Calculate billable amount of project based on resource

GauravGusain98:
### Description

- Update the billable amount and ledger amount calculation in the custom invoice 
- Update the ledger table in the custom invoice
- Added functionality to remove the the ledger details
- Calculate the billable amount for the client based on resource
- Added project billing details table to store the service rates of the project
- Fix the UI in the project team members page
- Show the start date and end date in the project team members details page
- Added Migration to make the started_on column to nullable as by default it is set to current timestamp. So whenever we were udapting the column the column gets updated to current timestamp

### Checklist:

### DIFF
--- a/Modules/Invoice/Services/InvoiceService.php
+++ b/Modules/Invoice/Services/InvoiceService.php
@@ -815,6 +815,15 @@ class InvoiceService implements InvoiceServiceContract
 
     public function storeLedgerAccountData(array $data)
     {
+        $project = Project::find($data['project_id'] ?? null);
+        $client = Client::find($data['client_id'] ?? null);
+
+        if ($project) {
+            $ledgerAccountsIdToDelete = LedgerAccount::where('project_id', $project->id)->pluck('id')->toArray();
+        } else {
+            $ledgerAccountsIdToDelete = LedgerAccount::where('client_id', $client->id)->pluck('id')->toArray();
+        }
+
         foreach ($data['ledger_account_data'] as $ledgerAccountData) {
             if ($ledgerAccountData['id'] == null) {
                 LedgerAccount::create($ledgerAccountData);
@@ -823,6 +832,10 @@ class InvoiceService implements InvoiceServiceContract
 
             $ledgerAccount = LedgerAccount::find($ledgerAccountData['id']);
             $ledgerAccount->update($ledgerAccountData);
+            $index = array_search($ledgerAccount->id, $ledgerAccountsIdToDelete);
+            unset($ledgerAccountsIdToDelete[$index]);
         }
+
+        LedgerAccount::destroy($ledgerAccountsIdToDelete);
     }
 }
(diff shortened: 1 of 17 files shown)


Modules/Invoice/Services/InvoiceService.php
```diff
@@ -815,6 +815,15 @@ public function getLedgerAccountData(array $data)
 
     public function storeLedgerAccountData(array $data)
     {
+        $project = Project::find($data['project_id'] ?? null);
+        $client = Client::find($data['client_id'] ?? null);
+
+        if ($project) {
+            $ledgerAccountsIdToDelete = LedgerAccount::where('project_id', $project->id)->pluck('id')->toArray();
+        } else {
+            $ledgerAccountsIdToDelete = LedgerAccount::where('client_id', $client->id)->pluck('id')->toArray();
```
rathorevaibhav, line 824: what if client is also null?

Modules/Project/Resources/views/edit.blade.php
```diff
@@ -16,31 +16,39 @@ class="text-theme-body text-decoration-none mb-2 mb-xl-4 d-inline-flex align-ite
         <div class="mt-2">
             <ul class="nav nav-pills mb-2" id="pills-tab" role="tablist">
                 <li class="nav-item" role="presentation">
-                    <a class="nav-link active" data-toggle="pill" data-target="#project-details" type="button"
+                    <a class="nav-link active" data-toggle="pill" data-target="#projectDetails" type="button"
                         role="tab" aria-selected="true">Project details</a>
                 </li>
                 <li class="nav-item" role="presentation">
-                    <a class="nav-link" data-toggle="pill" data-target="#project-team-members" type="button"
+                    <a class="nav-link" data-toggle="pill" data-target="#ProjectTeamMembers" type="button"
```
rathorevaibhav, line 23:
```suggestion
                    <a class="nav-link" data-toggle="pill" data-target="#projectTeamMembers" type="button"
```

Modules/Project/Resources/views/edit.blade.php
```diff
@@ -16,31 +16,39 @@ class="text-theme-body text-decoration-none mb-2 mb-xl-4 d-inline-flex align-ite
...
             </ul>
             @include('status', ['errors' => $errors->all()])
             <div class="tab-content">
-                <div class="tab-pane fade show active mb-5" id="project-details" role="tabpanel">
+                <div class="tab-pane fade show active mb-5" id="projectDetails" role="tabpanel">
                     @include('project::subviews.edit-project-details')
                 </div>
 
-                <div class="tab-pane fade mb-5" id="project-team-members" role="tabpanel">
+                <div class="tab-pane fade mb-5" id="ProjectTeamMembers" role="tabpanel">
```
rathorevaibhav, line 41:
```suggestion
                <div class="tab-pane fade mb-5" id="projectTeamMembers" role="tabpanel">
```

Modules/Invoice/Services/InvoiceService.php
GauravGusain98: Correct, in that case we need to return

rathorevaibhav (approved):

Issue #1966, opened 2022-07-22
Add start date and end date for fixed billing projects


rathorevaibhav:
**Describe the task**
We need to track the project duration so that we can determine the team allocation status. Right now, there are some projects which have zero actual effort but there are team members assigned to those projects.

These projects are a problem because they do not give a clear projection of the project status and for how long the project will be open.

**Expected behavior**
- [x] When creating a new project, if the user selects fixed budget project type, two fields should show up. When user selects any other option than fixed-budget, these fields will be hidden. 
- [x] The two new fields will be non-mandatory and will capture start and end date for the projects.
- [x] When editing a project, if the project already has fixed budget selected, these two fields should show up.
- [x] When editing a project, if the project is monthly billing, then these fields should NOT show up. When the project type is changed and selected to fixed-budget, then these two fields should show up again.
- [x] We already have `start_date` and `end_date` columns in the `projects` table. No need to create two new columns. We'll use them.
- [x] Upon creating and editing projects, these two fields should be saved in DB.
- [x] In the project view page (`/projects/{project-id}/show`), add these two fields as readable fields.

GauravGusain98:
working fine. Closing

PR #2092, opened 2022-08-02
1966/add start date end date for fixed billing projects

Abhinegi2:
Targets #1966 

### Description
1-  When creating a new project, if the user selects fixed budget project type, two fields should show up. When user selects 
any other option than fixed-budget, these fields will be hidden. 
2-  The two new fields will be non-mandatory and will capture start and end date for the projects.
3-  When editing a project, if the project already has fixed budget selected, these two fields should show up.
4-  When editing a project, if the project is monthly billing, then these fields should NOT show up. When the project type is 
changed and selected to fixed-budget, then these two fields should show up again.
5-  We already have `start_date` and `end_date` columns in the `projects` table. No need to create two new columns. We'll use them.
6- Upon creating and editing projects, these two fields should be saved in DB.
7-  In the project view page (`/projects/{project-id}/show`), add these two fields as readable fields.

### Checklist:


Modules/Project/Resources/views/subviews/edit-project-details.blade.php
```diff
@@ -1,3 +1,4 @@
+<div id="edit_project_details_form">
 <div class="card">
```
rathorevaibhav, line 2: please add indentation as everything that's a child of `edit_project_details_form` should be one tab extra indented

Modules/Project/Services/ProjectService.php
```diff
@@ -172,8 +172,8 @@ private function updateProjectDetails($data, $project)
             'type' => $data['project_type'],
             'total_estimated_hours' => $data['total_estimated_hours'] ?? null,
             'monthly_estimated_hours' => $data['monthly_estimated_hours'] ?? null,
-            'start_date' => date('Y-m-d'),
-            'end_date' => date('Y-m-d'),
+            'start_date' => $data['start_date'],
+            'end_date' => $data['end_date'],
```
rathorevaibhav, line 176: are these getting passed from the frontend for hourly billing projects as well? Please check. If they're not being passed, then this will throw an error. So you'll have to check

rathorevaibhav (changes requested):

rathorevaibhav (approved):

Issue #2003, opened 2022-07-25
Button alignment fixes on HR job application listing page


P4NK4J:
**Describe the task**
The add new application button and search button are a bit off in vertical alignment, we would want them both to be on the same line. 
<img width="1148" alt="image" src="https://user-images.githubusercontent.com/35001775/180834398-33aecf5d-e0a2-422e-9a83-30af7e68fc4f.png">

GauravGusain98:
Looks fine on UAT closing

PR #2096, opened 2022-08-02
Button Allignment Fixed

jaskaransingh304:
Targets #2003 

### Description
<Changed alignment of search button , so that it  can align with Add new Application Button. 

### Checklist:


P4NK4J (approved):
Looks good.

PR #2114, opened 2022-08-03
Updated no show time limit to 12hrs

P4NK4J:
In HR rounds evaluation, if a round result is not updated for 2hrs, we used to mark the application `no-show`. We need to increase the time limit to 12hrs


P4NK4J:
Need to deploy this on. urgent basis, hence closing this PR and making another one from release

Issue #2112, opened 2022-08-03
Show Total Estimated Hours for fixed billing projects


Abhinegi2:
**Describe the task**
We need to trace the  Total Estimated Hours for fixed budget billing.

**Expected behavior**
- [x] When creating a new project, if the user selects fixed budget project type,  ` Total Estimated Hours ` should show up. When user selects any other option than fixed-budget, this fields will be hidden. 
- [x] The   ` Total Estimated Hours `  field  will be non-mandatory .
- [x] When editing a project, if the project already has fixed budget selected, this  field should show up.
- [x] When editing a project, if the project is monthly billing, then this field  should NOT show up. When the project type is changed and selected to fixed-budget, then this field should show up again.
- [x] We already have ` Total Estimated Hours `  column in the `projects` table. No need to create  new column. We'll use them.
- [x] Upon creating and editing projects, this field should be saved in DB.
- [x] In the project view page (`/projects/{project-id}/show`), add this field as readable fields.

rathorevaibhav:
This is done. Closing.

PR #2115, opened 2022-08-03
show total estimated hours for fixed billing projects

Abhinegi2:
Targets #2112 

### Description
1- When creating a new project, if the user selects fixed budget project type, Total Estimated Hours should show up. When user selects any other option than fixed-budget, this fields will be hidden.
2- The Total Estimated Hours field will be non-mandatory .
3- When editing a project, if the project already has fixed budget selected, this field should show up.
4- When editing a project, if the project is monthly billing, then this field should NOT show up. When the project type is changed and selected to fixed-budget, then this field should show up again.
5- We already have Total Estimated Hours column in the projects table. No need to create new column. We'll use them.
6- Upon creating and editing projects, this field should be saved in DB.
7- In the project view page (/projects/{project-id}/show), add this field as readable fields

### Checklist:


gautam071 (approved):
Looks good

Issue #1871, opened 2022-07-15
The filter for selecting start date and End date for displaying FTE should be dynamic


deepankar-sajwan:
**Describe the task**
Currently the filter selecting start date and End date for displaying the FTE is static, as the start date is set as 1st day of  current month, and end date is set as last day of current month. Therefore only FTE of current month is dipalyed. Hence there is no methord to check FTE of past month.

Refer screenshot for refernce

![image](https://user-images.githubusercontent.com/96309135/179198864-f3957a12-50d2-4a4b-8702-fc214b21eade.png)

Procedure to reach FTE details page
- Go to CRM and click o Projects
- Click on the any project under Project Name
- Click on FTE button

**Expected behavior**
The filter for selecting start date and End date should be dynamic. On selecting start date and end date for previous month, the FTE for past month should be dispayed

Ayush8923:
This issue PR is merged. Closing it.

PR #2140, opened 2022-08-05
The filter for selecting start date and End date for displaying FTE should be dynamic

Himanshunegi0332:
#1871 
### Description
Currently the filter selecting start date and End date for displaying the FTE is static, as the start date is set as 1st day of current month, and end date is set as last day of current month. Therefore only FTE of current month is displayed. Hence there is no method to check FTE of past month. 
Made it dynamic.

### Checklist:

### DIFF
--- a/Modules/EffortTracking/Resources/views/show.blade.php
+++ b/Modules/EffortTracking/Resources/views/show.blade.php
@@ -1,7 +1,7 @@
 @extends('efforttracking::layouts.master')
 @section('content')
 <div class="project-effort-tracking-container container py-10">
-    <a href="{{ route('project.index') }}" class="text-theme-body text-decoration-none mb-2 mb-xl-4 d-flex align-items-center">
+    <a href="{{ route('project.index') }}" class="text-theme-body text-decoration-none mb-2 mb-xl-4 align-items-center">
         <span class="mr-1 d-inline-flex w-8 h-8 w-xl-12 h-xl-12">
             {!! file_get_contents(public_path('icons/prev-icon.svg')) !!}
         </span>
@@ -36,19 +36,9 @@
             <div class="d-flex">
                 <div class='form-group mr-4 ml-1 mt-1 w-168'>
                     <select class="form-control bg-light" name="month" onchange="document.getElementById('FilterForm').submit();">
-                        <option {{ request()->input('month') == '' ? "selected=selected" : '' }} value="{{$currentMonth}}">{{$currentMonth}}</option>
-                        <option {{ request()->input('month') == '01' ? "selected=selected" : '' }} value="January">January</option>
-                        <option {{ request()->input('month') == '02' ? "selected=selected" : '' }} value="February">February</option>
-                        <option {{ request()->input('month') == '03' ? "selected=selected" : '' }} value="March">March</option>
-                        <option {{ request()->input('month') == '04' ? "selected=selected" : '' }} value="April">April</option>
-                        <option {{ request()->input('month') == '05' ? "selected=selected" : '' }} value="May">May</option>
-                        <option {{ request()->input('month') == '06' ? "selected=selected" : '' }} value="June">June</option>
-                        <option {{ request()->input('month') == '07' ? "selected=selected" : '' }} value="July">July</option>
-                        <option {{ request()->input('month') == '08' ? "selected=selected" : '' }} value="August">August</option>
-                        <option {{ request()->input('month') == '09' ? "selected=selected" : '' }} value="September">September</option>
-                        <option {{ request()->input('month') == '10' ? "selected=selected" : '' }} value="October">October</option>
-                        <option {{ request()->input('month') == '11' ? "selected=selected" : '' }} value="November">November</option>
-                        <option {{ request()->input('month') == '12' ? "selected=selected" : '' }} value="December">December</option>
+                        @foreach (config('constants.months') as $months => $month)
+                        <option value="{{ $month }}" {{ $currentMonth == $month ? 'selected' : '' }}>{{ $month }}</option>
+                        @endforeach
                     </select>
                 </div>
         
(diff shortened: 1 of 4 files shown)


Modules/EffortTracking/Resources/views/show.blade.php
```diff
@@ -1,56 +1,83 @@
...
-            </span>
-            <span>All Projects</span>
-        </a>
+<div class="project-effort-tracking-container container py-10">
+    <a href="{{ route('project.index') }}" class="text-theme-body text-decoration-none mb-2 mb-xl-4 d-flex align-items-center">
+        <span class="mr-1 d-inline-flex w-8 h-8 w-xl-12 h-xl-12">
+            {!! file_get_contents(public_path('icons/prev-icon.svg')) !!}
+        </span>
+        <span>All Projects</span>
+    </a>
```
P4NK4J, line 9: d-flex would make the whole container full width and will lead it to be clickable, we need only space up to All projects to be clickable. The UI will show `All projects` on a small space, while the remaining space will be clickable and take users to `All projects` page. Please test it once in UI

Modules/EffortTracking/Resources/views/show.blade.php
```diff
@@ -1,56 +1,83 @@
...
+                        <option {{ request()->input('month') == '05' ? "selected=selected" : '' }} value="May">May</option>
+                        <option {{ request()->input('month') == '06' ? "selected=selected" : '' }} value="June">June</option>
+                        <option {{ request()->input('month') == '07' ? "selected=selected" : '' }} value="July">July</option>
+                        <option {{ request()->input('month') == '08' ? "selected=selected" : '' }} value="August">August</option>
+                        <option {{ request()->input('month') == '09' ? "selected=selected" : '' }} value="September">September</option>
+                        <option {{ request()->input('month') == '10' ? "selected=selected" : '' }} value="October">October</option>
+                        <option {{ request()->input('month') == '11' ? "selected=selected" : '' }} value="November">November</option>
+                        <option {{ request()->input('month') == '12' ? "selected=selected" : '' }} value="December">December</option>
+                    </select>
```
P4NK4J, line 52: Can we fetch it from Config ? rather than hardcoding here ? Or maybe look out for some better way to pass an array of months in the controller itself.

Modules/EffortTracking/Services/EffortTrackingService.php
```diff
@@ -14,19 +14,27 @@
...
+        $Month = intval(date('m', strtotime($currentMonth)));
+        $thisMonth = intval(Carbon::now()->format('m'));
+        $monthsDifference = ($thisMonth - $Month);
+        $totalYears = (Carbon::now()->format('Y') - $currentYear);
+        $totalMonths = $monthsDifference + ($totalYears * 12);
+        $startDate = $project->client->getMonthStartDateAttribute($totalMonths);
+        $endDate = $project->client->getMonthEndDateAttribute($totalMonths);
         $totalWorkingDays = count($this->getWorkingDays($startDate, $endDate));
+        $totalEffort = $project->getCurrentHoursForMonthAttribute($startDate, $endDate);
+        $currentDate = now(config('constants.timezone.indian'));
```
P4NK4J, line 37: Can we fetch these variables in different function ? Let's say create a function `getFilterParameters` and return all these from there. This currently looks like a lot of things for `show()`

P4NK4J (changes requested):
Left some comments, overall looks good.

Modules/EffortTracking/Services/EffortTrackingService.php
```diff
@@ -14,18 +14,23 @@
...
         }
-        $totalEffort = $project->current_hours_for_month;
         $workingDays = $this->getWorkingDays($project->client->month_start_date, $currentDate);
-        $startDate = $project->client->month_start_date;
-        $endDate = $project->client->month_end_date;
+        $currentDate = now(config('constants.timezone.indian'));
+        $currentMonth = $data['month'] ?? Carbon::now()->format('F');
+        $currentYear = $data['year'] ?? Carbon::now()->format('Y');
+        $totalMonths = $this->getFilterParameter($currentMonth, $currentYear);
+        $startDate = $project->client->getMonthStartDateAttribute($totalMonths);
```
P4NK4J, line 30: Why can't we use the get attribute like `$project->client->month_start_date` here and other places?

Modules/EffortTracking/Services/EffortTrackingService.php
```diff
@@ -38,10 +43,24 @@ public function show($project)
...
+    {
+        $Month = intval(date('m', strtotime($currentMonth)));
+        $thisMonth = intval(Carbon::now()->format('m'));
+        $monthsDifference = ($thisMonth - $Month);
+        $totalYears = (Carbon::now()->format('Y') - $currentYear);
+        $totalMonths = $monthsDifference + ($totalYears * 12);
+
+        return $totalMonths;
+    }
+
```
P4NK4J, line 63: If its returning totalMonths can we rename this function to some specific? something like `getTotalMonthsFilterParameter` ?

P4NK4J (changes requested):
Left some comments, overall looks good.

Modules/EffortTracking/Services/EffortTrackingService.php
Himanshunegi0332: we are using $project->client->getMonthStartDateAttribute($totalMonths) because this function is made on project.php and by giving $totalMonths to this attributes helps us in getting the previous months start date and FTE also whereas by using $project->client->month_start_date  we are getting only current months start date.

Modules/EffortTracking/Services/EffortTrackingService.php
Himanshunegi0332: ok sir.

P4NK4J (approved):
Looks good.

Issue #2117, opened 2022-08-03
Application listing sorting


P4NK4J:
**Describe the task**
Under HR Recruitment, currently, applications other than "OPEN" are sorted by date, with older applications first. We want these applications to be sorted by date.
Whenever tabs other than "OPEN" are clicked, a GET Parameter is passed in the backend and filters the applications based on the GET Parameter query. The applications listed in other tabs are ordered by date, where the older applications appear on the top. We want them to be in the "Latest first" order.
<img width="1135" alt="image" src="https://user-images.githubusercontent.com/35001775/182592485-0e44b0af-de5c-48f2-9b26-a115168e6b3a.png">

Ayush8923:
This issue PR is merged. Closing it.

PR #2148, opened 2022-08-07
Sorting list based on updated at parameter in HR application tab

shardulgusain:
#2117 

### Description
1.Updated the filtered result listing based on "updated at" parameter.

### Checklist:


P4NK4J (approved):
Looks good.

Issue #2164, opened 2022-08-10
Create resources_has_variable_rates column in the project_billing_detail table


GauravGusain98:
**Describe the task**
Create a column with name `resources_has_variable_rates` in the`project_billing_detail` table. 

**Expected behavior**
We need to create a migration file that will add a column in the table
The column should be of type boolean and the default should be false

**References**
You can read about the migration form [here](https://laravel.com/docs/9.x/migrations#creating-columns)

PR #2165, opened 2022-08-10
created a column

Himanshunegi0332:
Targets #2164 

### Description
Created a resources has variable rates column in the project_billing_detail table

### Checklist:


P4NK4J (commented):
@Himanshunegi0332 I think this column is going to be either true or false, based on its name. Can you confirm, and update the data type as well ?

GauravGusain98 (approved):
Looks fine

Issue #2174, opened 2022-08-11
CC and BCC field available in the email body does not have email validation


gautam071:
**Describe the task**
When pending invoice email or payment confirmation mail is sent to client from portal, then the email body used was sending email contains **CC**  and **BCC** field do not have email validation. If we mention the email address as **abcdgmail.com** in BCC and CC fields, the mail is sent without any error displayed to the user.

Procedure to check pending invoice  reminder mail 
1. Go to Finance
2. Click on Invoices
3. Click on Add Invoice
4. Fill the form input sent date as 10th June 2022 and due date as 18th June 2022
5. Click on the save button
6. Change status as Pending on Invoice page
7. Click on the Reminder button available under the Email column
8. Fill in details in the form and check the error

Refer screenshot
![179501701-78589cce-6dfe-49eb-9e7b-54625e6c3b2a](https://user-images.githubusercontent.com/73119570/184133815-f799e660-6afa-4066-9226-a2b08175b4b4.png)

deepakbarmola:
Closing the issue as linked PRs are approved and merged.

PR #2179, opened 2022-08-11
Added email validation for CC and BCC field available in the email body

Sachinbisht27:
Targets #2174 

### Description
Added email validation for CC and BCC field available in the invoice reminder email body.

### Checklist:

### DIFF
--- a/Modules/Invoice/Resources/views/modals/invoice-reminder.blade.php
+++ b/Modules/Invoice/Resources/views/modals/invoice-reminder.blade.php
@@ -34,7 +34,7 @@
                                     <i class="fa fa-question-circle"></i>
                                 </span>
                             </label>
-                            <input type="text" name="cc" id="cc" class="form-control" value="{{ config('invoice.mail.send-invoice.email') }}">
+                            <input type="email" name="cc" id="cc" class="form-control" value="{{ config('invoice.mail.send-invoice.email') }}">
                         </div>
                         <div class="form-group col-md-12">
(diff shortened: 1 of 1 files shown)


Modules/Invoice/Resources/views/modals/invoice-reminder.blade.php
```diff
@@ -43,7 +43,7 @@ class="form-control" required>
                                     <i class="fa fa-question-circle"></i>
                                 </span>
                             </label>
-                            <input type="text" name="bcc" id="bcc" class="form-control" value="">
+                            <input type="email" name="bcc" id="bcc" class="form-control" value="{{ config('invoice.mail.send-invoice.email') }}">
```
gautam071, line 46: Already via CC, the mail is sent to the same email we don't need this,
we can remove this line 
```suggestion
                            <input type="email" name="bcc" id="bcc" class="form-control" value="">
```

gautam071 (commented):
One suggestion

Modules/Invoice/Resources/views/modals/invoice-reminder.blade.php
Sachinbisht27: Yes, sure getting the point.

gautam071 (approved):
Looks good

gautam071 (changes requested):
looks good!

gautam071 (approved):

Issue #2169, opened 2022-08-11
The project details are not getting stored


GauravGusain98:
**Describe the bug**
When we save the project details in the project edit page and the project billing level field is not selected, in that case success message comes up while saving the details but we actually get error from the backend that billing level is required. 

Step 1: Donot select any option in billing level field and also update any other field to check if the data is getting saved or not
<img width="1216" alt="image" src="https://user-images.githubusercontent.com/54376313/184074754-e90f6e1d-7f70-47bb-bef9-001fa119ac0c.png">
 
Step 2: Save the form
<img width="1302" alt="image" src="https://user-images.githubusercontent.com/54376313/184074864-cc14323e-db93-42c0-88e4-1a006d7af0b4.png">

Step 3: Check the error in network tab
<img width="1415" alt="image" src="https://user-images.githubusercontent.com/54376313/184075035-6a085743-7f65-46fc-bf1b-f88fe6004f32.png">

**Expected behavior**
If there is any error coming from backend, we need to show error messages and the success message should not come 

**Additional context**

Ayush8923:
This issue PR is merged. Closing it.

PR #2187, opened 2022-08-12
error message shown when there is an error come from backend

tarunnjoshi:
Targets #2169

### Description
error message shown when there is an error come from backend

### Checklist:

### DIFF
--- a/Modules/Project/Resources/views/edit.blade.php
+++ b/Modules/Project/Resources/views/edit.blade.php
@@ -117,7 +117,8 @@
                             }
                             $('#edit-project-errors').removeClass('d-none');
                             $('.save-btn').attr('disabled', false);
-                            this.$toast.success('Project details updated!');
+                            if(error)
+                            this.$toast.error('error!');
                         })
                 },
 


Modules/Project/Resources/views/edit.blade.php
```diff
@@ -117,7 +117,8 @@ class="text-theme-body text-decoration-none mb-2 mb-xl-4 d-inline-flex align-ite
                             }
                             $('#edit-project-errors').removeClass('d-none');
                             $('.save-btn').attr('disabled', false);
-                            this.$toast.success('Project details updated!');
+                            if(error)
+                            this.$toast.error('error!');
```
gautam071, line 121:
```suggestion
                            if(error){
                                this.$toast.error('Error while saving the project!');
                            }
```

gautam071: Can we also show where the error is happening?

gautam071 (changes requested):
Left few suggestions

Modules/Project/Resources/views/edit.blade.php
tarunnjoshi: Screenshot (5)
The error is happening in network section and it will not save the data

gautam071: The error is a backend validation it will be passed to the front end!

gautam071 (approved):
Looks good

PR #2202, opened 2022-08-13
Show FTE chart only for current month

GauravGusain98:
### Description

- Added check to show the chart only for current month

### Checklist:


gautam071 (approved):
Looks good!

PR #2208, opened 2022-08-15
Ability to recurring expenses in the portal 

pankaj-ag:
This is our first cut to move in the direction of organization expenses that's why I am not linking to any issue for now. 
I have made the following changes in the PR: 
1. Setup CURD Ops for recurring expenses (Subscriptions)
2. Added a scaffolding to link the recurring expenses to all the expenses
3. A migration and modal for recurring expenses
4. Remove some unwanted and commented code from the old modules. 

### Checklist:

### DIFF
--- a/Modules/Expense/Entities/RecurringExpense.php
+++ b/Modules/Expense/Entities/RecurringExpense.php
@@ -28,16 +28,20 @@ class RecurringExpense extends Model
         $initialDueDate = $this->initial_due_date;
         $frequency = $this->frequency;
 
+        if(now()->lt($initialDueDate)) {
+            return $initialDueDate;
+        }
+
         if($frequency == 'monthly' && now()->gt($initialDueDate)) {
            $month = date('m');
-           $initialDueDate = $initialDueDate->format("Y-$month-d");
+           return Carbon::parse($initialDueDate->format("Y-$month-d"));
         } 
         
         if($frequency == 'yearly') {
             $year = date('Y');
-            $initialDueDate = $initialDueDate->format("$year-m-d");
+            return Carbon::parse($initialDueDate->format("$year-m-d"));
         }
 
-        return Carbon::parse($initialDueDate);
+        return $initialDueDate;
     }
 }
--- a/Modules/Expense/Entities/RecurringExpense.php
+++ b/Modules/Expense/Entities/RecurringExpense.php
@@ -33,7 +33,7 @@ class RecurringExpense extends Model
             return $initialDueDate;
         }
 
-        if ($frequency == 'monthly' && now()->gt($initialDueDate)) {
+        if ($frequency == 'monthly') {
             $month = date('m');
 
             return Carbon::parse($initialDueDate->format("Y-$month-d"));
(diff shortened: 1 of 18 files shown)


Modules/Expense/Entities/RecurringExpense.php
```diff
@@ -0,0 +1,50 @@
...
+            $month = date('m');
+
+            return Carbon::parse($initialDueDate->format("Y-$month-d"));
+        }
+
+        if ($frequency == 'yearly') {
+            $year = date('Y');
+
+            return Carbon::parse($initialDueDate->format("$year-m-d"));
+        }
```
P4NK4J, line 46: @pankaj-ag Do you think using switch case would be helpful here, we may have recurring expenses quarterly, fortnightly or weekly as well ?

Modules/Expense/Http/Controllers/RecurringExpenseController.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public function create()
+    {
+        return view('expense::recurring.create', ['countries' => Country::all()]);
+    }
+
+    public function store(Request $request)
+    {
+        // ToDo:: we need to add validations here.
```
P4NK4J, line 33: Thanks for adding the ToDos

Modules/Expense/Resources/views/recurring/edit.blade.php
```diff
@@ -0,0 +1,101 @@
...
+            <h4 class="mb-1 pb-1 fz-28">Edit {{ $recurringExpense->name }}</h4>
+            <span>
+                <form action={{ route('expense.recurring.destroy', $recurringExpense) }} method="POST"
+                    onsubmit="return confirm('Are you sure?')">
+                    @csrf
+                    @method('delete')
+                    <button type="submit" class="btn btn-danger text-white">Remove</button>
+                </form>
+
+
```
P4NK4J, line 16: Let's remove these spaces

P4NK4J (changes requested):
Looks good. Left some comments. Also, if we can remove the extra spaces from blade files, that would be great

Modules/Expense/Entities/RecurringExpense.php
pankaj-ag: @P4NK4J I think we will use a switch case when we have more than 3 conditions. I don't see a possibility of that right now.

Modules/Expense/Resources/views/recurring/edit.blade.php
pankaj-ag: Done and sorry about that, Need to reinstall my IDE format extension.

Modules/Expense/Entities/RecurringExpense.php
P4NK4J: Yeah that makes sense, I was only thinking in terms of future possibilities, but if we are only having monthly expenses as of now, I am fine with that.

P4NK4J (approved):
Looks good.

Issue #2170, opened 2022-08-11
Book borrowers can't be greater than the number of copies available


mdshamoon:
**Describe the bug**
A book having two copies has three borrowers. This can't be possible in real life.
We need to add a restriction that does not allow borrowing a book if the number of copies and borrowers are same.

**To Reproduce**
1. Go to books page
2. Try to borrow a book with one copy that is already borrowed by someone. 
3. The system allows you to do that

**Expected behavior**
It should not allow the user to borrow the books if the number of copies are finished. It should show an error saying `Sorry ! No more copies of this book available right now.`.

**Screenshots**

A book having two copies and three borrowers
<img width="416" alt="image" src="https://user-images.githubusercontent.com/32592458/184076812-1b517a77-3dfa-4905-9a94-53e68a663f55.png">

Ayush8923:
This issue PR is merged. Closing it.

Ayush8923:
This issue PR is merged. Closing it.

PR #2211, opened 2022-08-15
Bugfix/borrowers can't be more

Himanshunegi0332:
#2170 

### Description
A book having two copies has three borrowers. This can't be possible in real life.
We need to add a restriction that does not allow borrowing a book if the number of copies and borrowers are same.
Fixed this issue

### Checklist:

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -961,7 +961,6 @@ if (document.getElementById("show_book_info")) {
 				if (this.borrowers.length<this.book.number_of_copies) {
 				let response = await axios.get(this.borrowBookRoute);
 				this.isBorrowed = true;
-				this.borrowers = response.data.borrowers;
 				}
 				else {
 				alert("Sorry ! No more copies of this book available right now.");
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -958,9 +958,14 @@ if (document.getElementById("show_book_info")) {
 			},
 
 			borrowTheBook: async function() {
+				if (this.borrowers.length<this.book.number_of_copies) {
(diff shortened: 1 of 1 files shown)


resources/js/app.js
```diff
@@ -958,9 +958,14 @@ if (document.getElementById("show_book_info")) {
 			},
 
 			borrowTheBook: async function() {
-				let response = await axios.get(this.borrowBookRoute);
-				this.isBorrowed = true;
-				this.borrowers = response.data.borrowers;
+				if (this.borrowers.length<this.book.number_of_copies) {
```
gautam071, line 961:
```suggestion
				if (this.borrowers.length < this.book.number_of_copies) {
```

resources/js/app.js
```diff
@@ -958,9 +958,14 @@ if (document.getElementById("show_book_info")) {
...
+					let response = await axios.get(this.borrowBookRoute);
```
gautam071, line 962: what does this do?

gautam071 (changes requested):
Left some questions?

resources/js/app.js
Himanshunegi0332: Here, if the numbers of copies are greater than the borrowers length then it will run this line  await axios.get(this.borrowBookRoute);  ,  in this await function is used to wait for a promise and axios.get is used to get data from the this.borrowBookRoute .

resources/js/app.js
Himanshunegi0332: ok sir.

gautam071 (approved):
Approved this PR

Issue #2214, opened 2022-08-16
No Contract label UI change


gautam071:
**Describe the task**
The `No Contract` button on the `CRM->Project` page doesn't look good. We need to make the button UI friendly to the user. 
This is how it looks
<img width="1440" alt="Screenshot 2022-08-16 at 12 53 42 PM" src="https://user-images.githubusercontent.com/73119570/184822257-9c7d4960-02b1-4b71-bfbd-5545af56dc26.png">
 

**Expected behaviour**
Can we change the label to something like in the below screenshot
<img width="1440" alt="Screenshot 2022-08-16 at 12 54 51 PM" src="https://user-images.githubusercontent.com/73119570/184822425-afc5d57c-7b32-4598-8074-0c79888ed9e8.png">

deepankar-sajwan:
Closing as No contract label is updated

PR #2219, opened 2022-08-16
updated no contract label ui

Jkishan544:
Targets #2214 

### Checklist:

### DIFF
--- a/Modules/Project/Resources/views/index.blade.php
+++ b/Modules/Project/Resources/views/index.blade.php
@@ -85,7 +85,7 @@
                                 </td>
                                 <td>
                                     @if(empty($project->projectContracts->first()->contract_file_path))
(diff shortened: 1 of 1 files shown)


Modules/Project/Resources/views/index.blade.php
```diff
@@ -85,7 +85,7 @@
                                 </td>
                                 <td>
                                     @if(empty($project->projectContracts->first()->contract_file_path))
-                                        <span class="badge badge-light border border-dark rounded-0">No Contract</span>
+                                        <span class="btn btn-danger font-weight-bold rounded-pill ">No Contract</span>
```
gautam071, line 88: Can we change it to the label from the button? 
Beacuse the button is not needed here

gautam071 (changes requested):

PR #2218, opened 2022-08-16
making the delete button

gulshan81:
Targets #2190

### Description
making the delete button

### Checklist:
- [x] I have performed a self-review of my own code.

PR #2229, opened 2022-08-17
Creating the delete button

gulshan81:
Targets #2218

### Description
created the delete button and it's functionality in prospect with the help of soft delete. 

### Checklist:

### DIFF
--- a/Modules/Prospect/Entities/Prospect.php
+++ b/Modules/Prospect/Entities/Prospect.php
@@ -2,18 +2,20 @@
 
 namespace Modules\Prospect\Entities;
 
-use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Modules\User\Entities\User;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Modules\ModuleChecklist\Entities\NDAMeta;
 use Modules\Communication\Entities\CalendarMeeting;
 use Modules\ModuleChecklist\Entities\ModuleChecklist;
 
 class Prospect extends Model
 {
-    use HasFactory;
+    use HasFactory, SoftDeletes;
 
     protected $table = 'prospects';
+    protected $dates = ['deleted_at'];
     protected $primarykey = 'prospect_id';
     protected $fillable = ['created_by', 'status', 'assign_to', 'name', 'coming_from', 'coming_from_id', 'brief_info'];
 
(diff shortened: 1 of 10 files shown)


gulshan81:
Please delete the previous pr named (making the delete button #2218 ) because I created a new pr for that issue. And prefer this pr for that issue.

Modules/Prospect/Entities/Prospect.php
```diff
@@ -4,12 +4,19 @@
...
 use Modules\Communication\Entities\CalendarMeeting;
 use Modules\ModuleChecklist\Entities\ModuleChecklist;
 
 class Prospect extends Model
 {
+    use HasFactory, SoftDeletes;
+
+    protected $table = 'prospects';
+    protected $dates = ['deleted_at'];
+    protected $primarykey = 'prospect_id';
```
gautam071, line 19: What was the reason for adding this $primarykey?

Modules/Prospect/Http/Controllers/ProspectController.php
```diff
@@ -88,4 +88,34 @@ public function openDocument(Request $request, $documentID)
...
+
+    /**
+     * restore all prospect.
+     */
+    public function restoreAll()
+    {
+        Prospect::onlyTrashed()->restore();
+
+        return redirect()->back();
+    }
```
gautam071, line 120: What is the purpose of adding restore?

gautam071 (changes requested):
few questions?

Modules/Prospect/Entities/Prospect.php
gulshan81: because  this primary key is worked as a parent row

Modules/Prospect/Http/Controllers/ProspectController.php
gulshan81: Because
 1. destroy() method will soft delete a specific post from the list.
2. restore() method will restore specific posts from the list
3. restore all() method will restore all soft deleted posts.

Modules/Prospect/Entities/Prospect.php
gulshan81: the reason is that I supposed that when the delete function executes in the prospects table then it searches the prospect id in the module file where we describe the soft delete. But it works properly without it so I understand that the soft delete already has that property.

Modules/Prospect/Http/Controllers/ProspectController.php
gulshan81: sorry, I did not understand you at first but the purpose of this restore function is to restore the prospects in the database but there is no need for this so I removing the functionality from my code.

gautam071 (approved):
Looks good!

PR #2248, opened 2022-08-19
Added the fix for the segment ID not defined in the segment page

gautam071:
### Description
- Added the fix for the not defined `segment.id` on the `HR->Segment` Page.

### Checklist:


P4NK4J (approved):
Looks good.

PR #2231, opened 2022-08-17
added search by category functionality

parasnegi2002:
Targets #{2185}

### Description
added the functionality "sort by category "

Issue #2185, opened 2022-08-12
Sort by book categories 


gautam071:
**Describe the task**
Right now if a user wants to see the books by category is not possible. 
<img width="1440" alt="Screenshot 2022-08-12 at 1 29 48 PM" src="https://user-images.githubusercontent.com/73119570/184311962-b544afa1-e91d-48da-94ae-7f360c4ed43a.png">

**Expected behaviour**
- We can add a dropdown field with all the categories. Show the books from selected categories.

deepankar-sajwan:
closing as task is implemented 
<img width="885" alt="image" src="https://github.com/ColoredCow/portal/assets/96309135/e8a46a68-abb3-46c5-b429-6891e5512909">

PR #2263, opened 2022-08-23
Removed the functionalty for the sort by category on books 

gautam071:
…causing the error on the UAT

Targets #2185 

### Description
- Removed the code added in the sort by category in this PR #2231 

### Checklist:

### DIFF
--- a/app/Http/Controllers/KnowledgeCafe/Library/BookController.php
+++ b/app/Http/Controllers/KnowledgeCafe/Library/BookController.php
@@ -2,14 +2,14 @@
 
 namespace App\Http\Controllers\KnowledgeCafe\Library;
 
+use Carbon\Carbon;
+use Illuminate\Http\Request;
+use App\Services\BookServices;
 use App\Http\Controllers\Controller;
-use App\Http\Requests\KnowledgeCafe\Library\BookRequest;
 use App\Models\KnowledgeCafe\Library\Book;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/KnowledgeCafe/Library/BookController.php
```diff
@@ -26,9 +26,8 @@ public function __construct()
     public function index(Request $request)
     {
         $this->authorize('list', Book::class);
-        $searchCategory = $request->category_name ?? false;
-        $searchString = $request->search ?? false;
-        $books = $searchCategory ? Book::getByCategoryName($searchCategory) : Book::getList($searchString);
+        $searchString = (request()->has('search')) ? request()->input('search') : false;
+        $books = Book::getList($searchString);
```
P4NK4J, line 30: Seems like this was a search query, based on book name and category both and with these changes we have removed the category part ? If that's true, why we are removing it ? If you are doing this to revert the changes for the release, I am fine by that, but we need to make sure this is to be added in future

P4NK4J (approved):
Approved with a question

app/Http/Controllers/KnowledgeCafe/Library/BookController.php
gautam071: I am doing this to revert the changes for the release. yea will make sure we will add in the future

Issue #2220, opened 2022-08-16
Send email check on bulk import


P4NK4J:
**Describe the task**
When we import an excel file in `add new application`, the system currently shoots out a verification email. We would like to have a `send email` check, and a `view` button (an eye icon) to select or unselect the verification email trigger. 

The email Under the modal of import excel, we need similar functionality. Clicking on the `view` icon, the verification email should appear in the editor. The user should have the functionality to edit the email. Once the import form is submitted, the backend should check whether or not to trigger the email based on this `send email` check. 

You can find the verification email template in the settings section

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/35001775/184864057-79a593f8-aec9-4fae-9958-b305453308bf.png">
<img width="800" alt="image" src="https://user-images.githubusercontent.com/35001775/184863949-a33368b9-899e-40c9-8004-e352d011366e.png">

Ayush8923:
This issue PR is merged. Closing it.

PR #2271, opened 2022-08-24
send email check on bulk import

LAKSHYA0501:
Targets #2220 

### Description
1. Added  `send email` check and view button to select or unselect the verification email trigger .
2. Added email template to the import excel modal .
3. on import, the backend should check whether or not to trigger the email based on this `send email` check

### Checklist:

### DIFF
--- a/Modules/HR/Http/Controllers/Recruitment/ApplicantController.php
+++ b/Modules/HR/Http/Controllers/Recruitment/ApplicantController.php
@@ -37,7 +37,6 @@ class ApplicantController extends Controller
     {
         $hrJobs = Job::whereIn('type', ['job', 'internship'])->orderBy('title')->get();
 
-        $this->authorize('view', Setting::class);
         $verifyMail = Setting::where('module', 'hr')->get()->keyBy('setting_key');
 
         return view('hr.application.create', ['hrJobs' => $hrJobs], ['verifyMail' => $verifyMail]);
--- a/Modules/HR/Http/Controllers/Recruitment/ApplicantController.php
+++ b/Modules/HR/Http/Controllers/Recruitment/ApplicantController.php
@@ -3,17 +3,18 @@
 namespace Modules\HR\Http\Controllers\Recruitment;
 
 use App\Imports\ApplicationImport;
+use App\Models\Setting;
 use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Controller;
 use Maatwebsite\Excel\Facades\Excel;
 use Modules\HR\Contracts\ApplicationServiceContract;
 use Modules\HR\Entities\Applicant;
-use Modules\HR\Entities\Job;
-use Modules\HR\Http\Requests\Recruitment\ApplicantRequest;
 use Modules\HR\Entities\Application;
-use Modules\User\Entities\User;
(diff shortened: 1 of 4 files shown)


Modules/HR/Http/Controllers/Recruitment/ApplicantController.php
```diff
@@ -36,7 +37,10 @@ public function create()
     {
         $hrJobs = Job::whereIn('type', ['job', 'internship'])->orderBy('title')->get();
 
-        return view('hr.application.create', ['hrJobs' => $hrJobs]);
+        $this->authorize('view', Setting::class);
```
P4NK4J, line 40: Why we are using this here?

Modules/HR/Listeners/Recruitment/ApplicantEmailVerification.php
```diff
@@ -29,21 +28,24 @@ public function handle(ApplicationCreated $event)
...
-        $body = Setting::where('module', 'hr')->where('setting_key', 'applicant_verification_body')->first();
+        $subject = request()->subject;
+        $body = request()->body;
 
-        $body->setting_value = str_replace(config('constants.hr.template-variables.applicant-name'), $application->applicant->name, $body->setting_value);
+        $body = str_replace(config('constants.hr.template-variables.applicant-name'), $application->applicant->name, $body);
         $encryptedEmail = encrypt($application->applicant->email);
         $verification_link = route('applicant.email.verification', [$encryptedEmail, $application->id]);
-        $body->setting_value = str_replace(config('constants.hr.template-variables.verification-link'), $verification_link, $body->setting_value);
+        $body = str_replace(config('constants.hr.template-variables.verification-link'), $verification_link, $body);
```
P4NK4J, line 39: This is breaking the current email verification functionality. You have disabled the email verification auto email, while you need to pass a flag and check if we are importing through excel, then only perform the actions you have written

P4NK4J (changes requested):
@LAKSHYA0501 The other functionality of sending verification email when an application is applied through website. Please take care of both the functionalities should be working

Modules/HR/Http/Controllers/Recruitment/ApplicantController.php
LAKSHYA0501: Ok so from the place i use it there it was use for authorization for every user, so i thought it would be needed here as well. That's why i used it, can you confirm whether it is needed or not if not i can remove it.

Modules/HR/Listeners/Recruitment/ApplicantEmailVerification.php
LAKSHYA0501: ok i am on it, fixing the particular.

Modules/HR/Http/Controllers/Recruitment/ApplicantController.php
P4NK4J: We already have the resources authorized in constructor for places needed. We should be good to remove this.

P4NK4J (commented):
Responsed to your comment

Modules/HR/Http/Controllers/Recruitment/ApplicantController.php
LAKSHYA0501: ok done

P4NK4J (approved):
Looks good.

Issue #2322, opened 2022-08-31
Total Calculation of the hours booked and expected hours in the project page. 


gautam071:
**Is your feature request related to a problem? Please describe.**
- There is no way to see the total hours booked and expected hours on a project page  
<img width="1440" alt="Screenshot 2022-08-31 at 11 54 16 AM" src="https://user-images.githubusercontent.com/73119570/187612705-d8314dc4-77e5-443e-9970-2f530db92d7b.png">

**Describe the solution you'd like**
- We need to show a `total` row to show all the total effort of the team member.

PR #2346, opened 2022-09-03
Total Calculation of the hours booked and expected hours in the proje…

gulshan81:
…ct page

Targets #2322 

### Description
"Total Calculation of the hours booked and expected hours in the project page"

### Checklist:


Modules/Project/Resources/views/show.blade.php
```diff
@@ -134,6 +134,11 @@
                                             <div class="fz-lg-28 text-center mt-4">No member in the project</div>
                                         @else
                                             <tbody>
+                                                @php
+                                                    $total = 0;
+                                                    $var = 0;
+                                                    $velocity = 0;
```
gautam071, line 140: What is the use of adding velocity it was not in the requirement?

gautam071 (changes requested):

Modules/Project/Resources/views/show.blade.php
gulshan81: I removed the velocity term.

Issue #2365, opened 2022-09-05
On filter for the project FTE page


gautam071:
**Is your feature request related to a problem? Please describe.**
- When you select the filter on the FTE page for the project it shows no data available
<img width="1141" alt="Screenshot 2022-09-05 at 1 19 45 PM" src="https://user-images.githubusercontent.com/73119570/188399408-0cd9d7b9-8678-49f2-a036-9fedfe2bd010.png">

**Describe the solution you'd like**

- [ ] Instead of showing no data create a table like this and calculate the effort spent by each individual team member shown there on the `No Data Available` page
        <img width="598" alt="Screenshot 2022-09-05 at 1 19 33 PM" src="https://user-images.githubusercontent.com/73119570/188399624-7dc28f02-b42e-403f-a775-cb07f91649b2.png">

**Describe alternatives you've considered**

**Additional context**

Ayush8923:
This issue PR is merged. Closing it.

PR #2367, opened 2022-09-05
On filtering months in FTE page table is available

Sachinbisht27:
Targets #2365 

### Description
On filtering months in FTE page for Projects table is available for all months.

### Checklist:

### DIFF
--- a/Modules/EffortTracking/Resources/views/show.blade.php
+++ b/Modules/EffortTracking/Resources/views/show.blade.php
@@ -67,7 +67,6 @@
         @endif
         </div>
     </div>
-    @if($totalMonths === 0 || $totalMonths != 0)
     <div class="project-resource-effort-tracking-container container mt-4 pb-10">
         <div class="card">
             <div class="card-header">
@@ -80,8 +79,8 @@
                         <strong>Last refreshed at:</strong>{{ (Carbon\Carbon::parse($project->last_updated_at)->setTimezone('Asia/Kolkata')->format(' Y-M-d , D h:i:s A')) }}
                         </div > 
                         <div class="fz-14 float-right">
-                        <strong>Timeline:</strong>{{ (Carbon\Carbon::parse($project->client->month_start_date)->format('dS M')) }}                       
-                        -{{ (Carbon\Carbon::parse($project->client->month_end_date)->format('dS M')) }}              
+                        <strong>Timeline:</strong>{{ $startDate->format('dS M') }}
+                        -{{ $endDate->format('dS M') }}
                         </div>
                     </h4>
                 </div>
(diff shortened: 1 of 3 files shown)


gautam071:
@Sachinbisht27 Displaying the table is part of the requirement 
but the table needs to calculate all the details according to the filtered month

gautam071:
Please make the changes accordingly

Modules/EffortTracking/Resources/views/show.blade.php
```diff
@@ -67,7 +67,6 @@
         @endif
         </div>
     </div>
-    @if($totalMonths === 0)
```
gautam071, line 70: why did you remove this code?

Modules/EffortTracking/Resources/views/show.blade.php
```diff
@@ -118,23 +117,22 @@
...
+                                <td>{{ $teamMember->getCurrentExpectedEffortAttribute($startDate, $endDate) }}</td>
+                                <td>{{ $teamMember->getExpectedEffortTillTodayAttribute($startDate, $endDate) }}</td>
+                                <td>{{ $teamMember->getExpectedEffortTillTodayAttribute($startDate, $endDate) - $teamMember->getCurrentActualEffortAttribute($startDate, $endDate) }}
                                 </td>
-                                <td class="{{ $teamMember->velocity >= 1 ? 'text-success' : 'text-danger' }}">
-                                    {{ $teamMember->velocity }}</td>
-                                <td>{{ $teamMember->fte }}</td>
+                                <td class="{{ $teamMember->getVelocityAttribute($startDate, $endDate) >= 1 ? 'text-success' : 'text-danger' }}">
+                                    {{ $teamMember->getVelocityAttribute($startDate, $endDate) }}</td>
+                                <td>{{ $teamMember->getFteAttribute($startDate, $endDate) }}</td>
```
gautam071, line 128: Looks like the function is called in multiple places can we assign a variable for the function and call it

gautam071 (changes requested):
Left some comments

Modules/EffortTracking/Resources/views/show.blade.php
Sachinbisht27: if  i am putting @if($totalMonths === 0) there the table is only working for current month and if i am putting @if($totalMonths === 0 || $totalMonths != 0) then the table working for both current and previous months also.
But, i thought that if we are putting total month ($totalMonths === 0 || $totalMonths != 0) what is the need of this $totalMonths now.
that's why i removed it...
Should i have to add the condition for ==0 and != 0?

Modules/EffortTracking/Resources/views/show.blade.php
Sachinbisht27: *send

Sachinbisht27: @gautam071 after doing suggested changes i am not getting the data as expected.

Sachinbisht27:
@gautam071 on adding variables over attribute functions in show.blade not getting correct values in table.

Modules/EffortTracking/Resources/views/show.blade.php
gautam071: @Sachinbisht27 
`@if($totalMonths === 0 || $totalMonths != 0)`

it means `$totalMonths` if it is zero and not zero it should run right. 
Then what is the use of the check?

gautam071 (changes requested):

Modules/EffortTracking/Resources/views/show.blade.php
Sachinbisht27: yes, we have to run the table for both $totalMonths zero and not zero.  i thought that if we are putting ($totalMonths === 0 || $totalMonths != 0) what is the need of this. that's why i previously removed it.

gautam071 (approved):
Looks good to me

Issue #2359, opened 2022-09-05
Team interaction round applications shouldn't be marked for followup


P4NK4J:
**Describe the task**
Currently, the applications sent into Team Interaction Round, are marked as `Need Followup` in the front end after 2 days. You will find it in the codebase, we have a scheduler that marks the applications for `need follow up`. We would like to avoid this when moving an application into Team Interaction Round.

Steps to replicate:
1. Send any application into Team Interaction Round
2. Adjust the dates of your system for 3-4 days later
3. Run the `need follow up` scheduler using the artisan command, you will find this in the codebase

Ayush8923:
This issue is completed. Closing it.

PR #2373, opened 2022-09-05
Features/team interaction round application should not be marked up for followup

gulshan81:
Targets #2359 

### Description
  team interaction round application should not be marked up for followup

### Checklist:

### DIFF
--- a/Modules/HR/Console/Recruitment/MarkApplicationForFollowUp.php
+++ b/Modules/HR/Console/Recruitment/MarkApplicationForFollowUp.php
@@ -57,7 +57,7 @@ class MarkApplicationForFollowUp extends Command
             $previousRoundConductedOn = $applicationRound->getPreviousApplicationRound()->conducted_date;
 
             // check if the previous round has been conducted more than 3 days ago
-            if ($previousRoundConductedOn->diffInDays(now()) > 3 && $applicationRound->round->name != 'Trial Program') {
+            if ($previousRoundConductedOn->diffInDays(now()) > 3 && $applicationRound->round->name != 'Trial Program'  && $applicationRound->round->name != 'Team Interaction Round') {
                 $application->tag('need-follow-up');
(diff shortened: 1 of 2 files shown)


Modules/HR/Console/Recruitment/MarkApplicationForFollowUp.php
```diff
@@ -57,7 +57,7 @@ public function handleAwaitingCalendarConfirmationApplications()
             $previousRoundConductedOn = $applicationRound->getPreviousApplicationRound()->conducted_date;
 
             // check if the previous round has been conducted more than 3 days ago
-            if ($previousRoundConductedOn->diffInDays(now()) > 3 && $applicationRound->round->name != 'Trial Program') {
+            if ($previousRoundConductedOn->diffInDays(now()) > 3 && $applicationRound->round->name != 'Trial Program' && $applicationRound->round->name != 'Team Interaction Round') {
```
P4NK4J, line 60: Instead of round names, and different conditions, let's put the rounds in array and check if the `$application->round->name` is present in the rounds array or not.

gulshan81: it means we should not use the if condition.  we should use the array for the application round and check the round name in array with the help of dd.?

P4NK4J: What I meant was

```
if ($previousRoundConductedOn->diffInDays(now()) > 3 && in_array($applicationRound->round->name, ['Trial Program', 'Team Interaction Round']))
```

gulshan81: sorry, I see your msg right now that's why I reply you right now.

gulshan81: and late too.

P4NK4J (approved):
Looks good.

PR #2403, opened 2022-09-10
Condition for Users Table seeder should not run on production

brajesh-lab:
Targets #{2399}

### Description
check condition for Users table seeder should not run on production
### Checklist:


P4NK4J (approved):
Looks good.

Issue #2429, opened 2022-09-13
Team members not getting updated


tarunnjoshi:
**Describe the task**
First, you need to take access to the effort sheet by the use of Google Sheets API from there you can get access to the sheet and then configure it in your env, then use the link in the effort sheet which is present in projects. 

In the google sheet, it does not take the value according to the user name it takes the value according to its sequence same as the team members So we need to take the value according to the team member name, not by its sequence.  so next time when someone updates the team members booked hours by their user name irrespective of their location present on the FTE page it will save.

As shown below :

![image](https://user-images.githubusercontent.com/109790464/189925530-e7e3818a-5ac2-4b9e-baa0-1709fc14b769.png)

![image](https://user-images.githubusercontent.com/109790464/189926326-a2c943e1-3f17-443c-984e-7ed8d20677b9.png)    

Example:
In the first image team member, Abhishek Sharma is present but when we change the hours booked for Abhishek Sharma in image second which is present in the sheet it will not change because the sheet takes the starting four values not the fifth according to the sequence.
so we need to change it and take value from the user name not by sequence.

Ayush8923:
This issue PR is merged. Closing it.

PR #2435, opened 2022-09-14
team members are updated according to their user name

tarunnjoshi:
Targets #2429

### Description
Team Members are updated according to their user names.

### Checklist:
- [x] I have performed a self-review of my code.

### DIFF
--- a/Modules/EffortTracking/Services/EffortTrackingService.php
+++ b/Modules/EffortTracking/Services/EffortTrackingService.php
@@ -203,7 +203,6 @@ class EffortTrackingService
                 $projectMembersCount++;
             }
 
-
             try {
                 while (true) {
                     $range = 'C1:' . ++$lastColumn . '1';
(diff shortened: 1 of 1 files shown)


gautam071:
@GauravGusain98 Can you review this PR once

Modules/EffortTracking/Services/EffortTrackingService.php
```diff
@@ -191,6 +191,18 @@ public function getEffortForProject($project)
...
+            $range = config('efforttracking.default_start_column_in_effort_sheet') . '2:' . config('efforttracking.default_start_column_in_effort_sheet');
+            $sheet = $sheets->spreadsheet($sheetId)
+                ->range($range)
+                ->get();
+
+            foreach ($sheet as $rows) {
+                if (count($rows) == 0) {
+                    break;
+                }
+                $projectMembersCount++;
```
GauravGusain98, line 203: According to current logic the teamember will already have the value and this for loop will do an increment over that value. The initial value should be 0 of `projectMembersCount` at line 189. Then we will get the exact count.

GauravGusain98 (changes requested):
Please check the comment

Modules/EffortTracking/Services/EffortTrackingService.php
tarunnjoshi: Done with the changes.

PR #2440, opened 2022-09-14
Restriction follow up email attempts to 2

brajesh-lab:
Targets #{2411}

### Description
Restrict need follow up email attempts to 2 

### Checklist:


P4NK4J (changes requested):
@brajesh-lab This button you have put a condition, is to manually add follow up comments. There is an scheduled event, that triggers automated email for applicants who need followup. We already have that implemented for job applications, need to implement that for internship ones as well. Please go through the codebase and see where we tag applications for followup, there we have an email trigger added, we need to limit that.
Please revert these changes

brajesh-lab:
all are working good

Issue #2448, opened 2022-09-15
Change the text for the search input feild


gautam071:
**Is your feature request related to a problem? Please describe.**
- The placeholder of the text in the books for the search input field is not working as expected
<img width="1440" alt="Screenshot 2022-09-15 at 11 38 34 AM" src="https://user-images.githubusercontent.com/73119570/190327505-ee542df7-b678-49ce-8488-6363871f30bb.png">

**Describe the solution you'd like**
Change the text from `Search all Books` to  `Enter the book you want to search`

Ayush8923:
This issue PR is merged. Closing it.

PR #2456, opened 2022-09-15
I change the text from Search all Books to Enter the book name

Amanraj1315:
Targets #2448 

Description:
I change the text from Search all Books to Enter the book name


gautam071 (approved):
Looks good

Issue #2404, opened 2022-09-12
Inactive project should not come in the list of detail projects


pankaj-ag:
- Currently, we show the inactive and all the projects in the project details screen. 
- We should show only active projects.
<img width="1074" alt="image" src="https://user-images.githubusercontent.com/10670690/189583557-a4e142b9-6b8b-4f0d-b53b-7adeb7d954d2.png">

PR #2457, opened 2022-09-15
 only active projects are shown

ditya0504:
Targets #2404  

### Description
showing the only active projects.

### Checklist:


Shivam-Samant (approved):
Looks good

PR #2458, opened 2022-09-15
RevenueProceed Policies

pankaj-ag:
Currently, we are authorizing the requests without adding the policies. 

I have removed that restriction and created a task [here](https://github.com/ColoredCow/portal/issues/2460) to add the policies and enable this change.


GauravGusain98 (approved):
Approving. Please resolve the CI check before merging.

Issue #2442, opened 2022-09-14
Enhancements in Revenue Proceeds 


pankaj-ag:
- [x] - We need to make the category and currency dynamic. Currently, it's hardcoded in the blade files. Check to create an invoice form for an example.

- [x] - Let's add two more columns in the revenue proceed table and the form. 
   a: Year. - A dropdown will show where the years will start from 2017 to the current year
   b: month  - the name of the months will be shown. 

- [x] - Make year and month dynamic currently it's hardcoded. 

These two fields will help us to show the data in reports.

PR #2465, opened 2022-09-15
made year and month dynamic

Abhishek25081999:
Targets #2442

### Description
Made year and Month dynamic

### Checklist:

### DIFF
--- a/Modules/Revenue/Resources/views/index.blade.php
+++ b/Modules/Revenue/Resources/views/index.blade.php
@@ -25,7 +25,7 @@
             @foreach ($revenueProceeds as $revenueProceed)
                 <tr>
                     <td>{{ $revenueProceed->name }}</td>
-                    <td>{{Str::title(str_replace('-', ' ', $revenueProceed->category))}}</td>
+                    <td>{{Str::title(str_replace('_', ' ', $revenueProceed->category))}}</td>
                     <td>{{ $revenueProceed->currency }}</td>
                     <td>{{ $revenueProceed->amount }}</td>
                     <td>{{ $revenueProceed->notes }}</td>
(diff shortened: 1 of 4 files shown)


Modules/Revenue/Resources/views/index.blade.php
```diff
@@ -25,7 +25,7 @@
             @foreach ($revenueProceeds as $revenueProceed)
                 <tr>
                     <td>{{ $revenueProceed->name }}</td>
-                    <td>{{Str::title(str_replace('-', ' ', $revenueProceed->category))}}</td>
+                    <td>{{Str::title(str_replace('_', ' ', $revenueProceed->category))}}</td>
```
pankaj-ag, line 28: Can we use Str::title or something like that?

pankaj-ag (changes requested):
Also when we create revenue can we set the month, year and receive the date as per the current day?

Modules/Revenue/Resources/views/index.blade.php
Abhishek25081999: ok I'll try.

pankaj-ag: Yes @Abhishek25081999  you can use something from this list. 
https://laravel.com/docs/9.x/helpers#strings-method-list

Abhishek25081999: I think we can. I took reference from other files and its being used there already.

pankaj-ag, line 28:
```suggestion
                    <td>{{ Str::title($revenueProceed->category) }}</td>
```

pankaj-ag (approved):
Last feedback then we are good to merge.

PR #2488, opened 2022-09-19
Increase the max value

rohit123-rawat:
### Description
Increase the max value of Maximum interviews in a day in user-settings


P4NK4J (approved):
Looks good.

Issue #2521, opened 2022-09-22
Calculating FTE with projects which are not AMC


dhruvshrinet1:
Calculating the FTE of user for only those projects which have no AMC.

deepakbarmola:
PR is merged after approval hence closing the issue.

PR #2522, opened 2022-09-22
Calculating fte for projects which are not AMC

dhruvshrinet1:
Targets #{2521}
Issue: https://github.com/ColoredCow/portal/issues/2521

Calculating the fte of user/employee for only those projects which have not AMC

### DIFF
--- a/Modules/User/Entities/User.php
+++ b/Modules/User/Entities/User.php
@@ -180,7 +180,7 @@ class User extends Authenticatable
         $fte = 0;
        
         foreach ($this->projectTeamMembers()->with('project')->get() as $projectTeamMembers) {
-            if (!$projectTeamMembers->project->is_amc) {
+            if (! $projectTeamMembers->project->is_amc) {
                 $fte += $projectTeamMembers->fte;
             }
         }
--- a/Modules/User/Entities/User.php
+++ b/Modules/User/Entities/User.php
@@ -170,7 +170,7 @@ class User extends Authenticatable
     {
         $fte = 0;
 
-        $fteAmc=0;    
+        $fteAmc=0;
         foreach ($this->projectTeamMembers()->with('project')->get() as $projectTeamMember) {
             if (! $projectTeamMember->project->is_amc) {
                 $fte += $projectTeamMember->fte;
--- a/Modules/User/Entities/User.php
+++ b/Modules/User/Entities/User.php
@@ -177,89 +177,17 @@ class User extends Authenticatable
     }
     public function getMainProjectsFteAttribute()
     {
-
         $fte = 0;
+        foreach ($this->projectTeamMembers as $projectTeamMembers) {
 
-        foreach ($this->projectTeamMembers as $projectTeamMember) {
-
-            if(!$projectTeamMember->project->isAMC(0))
-            {
-            $fte += $projectTeamMember->fte;
+            if($projectTeamMembers->project->isAMC(0))
+            {                
+                $fte += $projectTeamMembers->fte;
             }
         }
-
-        return $fte;
-
-
-        // $projects = Project::where('is_amc', '=', 0)->get();
-        // foreach($projects as $project) {
-        //     dd($project);
-        // }
-        // dd($project);
-        // // $d = ProjectTeamMember::find(1);
-        // // echo $d;
-        // // foreach ($comments as $c)
-        // // {
-        // //      echo $c;
-        // // // }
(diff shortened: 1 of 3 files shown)


Modules/User/Entities/User.php
```diff
@@ -176,6 +175,17 @@ public function getFteAttribute()
 
         return $fte;
     }
+    public function getMainProjectsFteAttribute()
+    {
+        $fte = 0;
+        foreach ($this->projectTeamMembers as $projectTeamMembers) {
```
pankaj-ag, line 181:
```suggestion
        foreach ($this->projectTeamMembers()->with('project')->get() as $projectTeamMembers) {
```

pankaj-ag (changes requested):

Modules/User/Entities/User.php
```diff
@@ -146,7 +146,6 @@ public function activeProjectTeamMembers()
     {
         return $this->hasMany(ProjectTeamMember::class, 'team_member_id')->where('ended_on', null);
     }
-
```
gautam071, line 149: Add the space back

gautam071 (commented):
Looks good one suggestion

Modules/User/Entities/User.php
```diff
@@ -169,13 +169,34 @@ public function getMonthTotalEffortAttribute()
...
         foreach ($this->projectTeamMembers as $projectTeamMember) {
             $fte += $projectTeamMember->fte;
         }
 
         return $fte;
     }
+    public function getMainProjectsFteAttribute()
+    {
+        $fte = 0;
+        foreach ($this->projectTeamMembers()->with('project')->get() as $projectTeamMembers) {
```
pankaj-ag, line 181:
```suggestion
        foreach ($this->projectTeamMembers()->with('project')->get() as $projectTeamMember) {
```

Modules/User/Entities/User.php
```diff
@@ -169,13 +169,34 @@ public function getMonthTotalEffortAttribute()
...
+            if (! $projectTeamMembers->project->is_amc) {
+                $fte += $projectTeamMembers->fte;
+            }
+        }
+
+        return $fte;
+    }
+    public function getamcProjectsFteAttribute()
```
pankaj-ag, line 189:
```suggestion
    public function getAmcProjectsFteAttribute()
```

Modules/User/Entities/User.php
```diff
@@ -169,13 +169,34 @@ public function getMonthTotalEffortAttribute()
...
+    {
+        $fte = 0;
+        foreach ($this->projectTeamMembers()->with('project')->get() as $projectTeamMembers) {
+            if ($projectTeamMembers->project->is_amc) {
+                $fte += $projectTeamMembers->fte;
+            }
+        }
+
+        return $fte;
+    }
```
pankaj-ag, line 199: Since we are running a loop twice I think let's create a single function that gave a fet of both types of projects in one go. 

with name getFtesAttribute which will return [ 'main' => 10, 'amc' => 2 ] kind of thing.

resources/views/hr/employees/show.blade.php
```diff
@@ -15,7 +15,8 @@
                 @endif
             </div>
             <hr class='bg-dark mx-4 pb-0.5'>
-            <div class="font-weight-bold fz-24 pl-5 mt-5 mb-3 d-flex justify-content-inline">{{__('Current FTE: ')}}<div class=" ml-1 {{ $employee->user ? ($employee->user->fte > 1 ? 'text-success' : 'text-danger') : 'text-secondary'}} font-weight-bold">{{ $employee->user ? $employee->user->fte :'NA' }}</div></div>
+            <div class="font-weight-bold fz-24 pl-5 mt-5 mb-3 d-flex justify-content-inline">{{__('Current FTE: ')}}<div class=" ml-1 {{ $employee->user ? ($employee->user->MainProjectsFte > 1 ? 'text-success' : 'text-danger') : 'text-secondary'}} font-weight-bold">{{ $employee->user ? $employee->user->MainProjectsFte  :'NA' }}</div></div>
```
pankaj-ag, line 18: Let's move this function in a single if block with the employer check. We don't need to repeat the full code again.

pankaj-ag (changes requested):

Modules/User/Entities/User.php
dhruvshrinet1: Updated as per the feedback

pankaj-ag (approved):

Issue #2487, opened 2022-09-19
Use CurrencyAvgRates table data to calculate the profit and loss amount. 


pankaj-ag:
As we have added a table in the system now we need to use that for the calculations. 

1. If you go to Finance->Reports->ProfitAndLoss report you will see all the revenue data. 
2. All this data or amount is converted into INR. 
3. To convert the amount into INR we are taking the current daily exchange rates. 
4. We need to change that and get the rates for that month from the CurrencyAvgRate table. If there is no entry for that month then we can use the current rates by default. 
5. We need to change this in function **getParticularAmountForExport** in file **Modules/Report/Services/Finance/RevenueReportService.php**

PR #2527, opened 2022-09-23
CurrencyAvgRate table data to calculate p&l amount

Himanshunegi0332:
Targets #2487 

### Description
Created a CurrencyAvgRate table data to calculate p&l amount If there is no entry for that month then we can use the current rates by default.

### Checklist:

### DIFF
--- a/Modules/Report/Services/Finance/RevenueReportService.php
+++ b/Modules/Report/Services/Finance/RevenueReportService.php
@@ -38,9 +38,27 @@ class RevenueReportService
         return $particular;
     }
 
+    public function getAvgCurrencyRates()
+    {
+        $totalRate = 0;
+        $exchangeRate = [];
+        $currency_chages = [];
+        $exchangeRates = CurrencyAvgRate::select('avg_rate', 'captured_for')->groupBy('captured_for')->get()->toArray();
+        $exchangeMonth = null;
+        foreach ($exchangeRates as $exchangeRate) {
+            $exchangeMonth = (date('m-y', strtotime($exchangeRate['captured_for'])));
+            $currency_chages[] = [
+            $exchangeMonth => $exchangeRate['avg_rate']
+            ];
+        }
+        $totalRate = $exchangeRate['avg_rate'];
+
+        return $totalRate;
+    }
+
     private function getParticularAmountForDomestic(array $particular, Object $startDate, Object $endDate): array
     {
-        $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'indian');
+        $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'non-indian');
         $totalAmount = 0;
         $results = [];
 
@@ -57,29 +75,24 @@ class RevenueReportService
 
     private function getParticularAmountForExport(array $particular, Object $startDate, Object $endDate): array
     {
-        $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'non-indian');
+        $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'indian');
         $totalAmount = 0;
         $results = [];
-        $currency_chages = [];
+        $exchangeMonth = null;
+        $exchangeRates = CurrencyAvgRate::select('captured_for')->groupBy('captured_for')->get()->toArray();
+        foreach ($exchangeRates as $exchangeRate) {
+        $exchangeMonth = (date('m-y', strtotime($exchangeRate['captured_for'])));
+        }
 
         // ToDo:: We need to change this logic and get the exchange rate for every month.
-        $exchangeRate = [];
-        $exchangeRates = CurrencyAvgRate::select('avg_rate', 'captured_for')->groupBy('captured_for')->get()->toArray();
-        $exchangeMonth = null;
-        foreach ($exchangeRates as $exchangeRate) {
-            $exchangeMonth = (date('m-y', strtotime($exchangeRate['captured_for'])));
-            $currency_chages[] = [
-                $exchangeMonth => $exchangeRate['avg_rate']
-            ];
-        }
         foreach ($invoices as $invoice) {
             $dateKey = $invoice->sent_on->format('m-y');
             if ($exchangeMonth == $dateKey) {
-                $exchangeDollor = $exchangeRate['avg_rate'];
+                $exchangeDollor = $this->getAvgCurrencyRates();
             } else {
                 $exchangeDollor = app(CurrencyService::class)->getCurrentRatesInINR();
             }
-            $amount = ($invoice->amount) * intval($exchangeDollor);
+            $amount = ($invoice->amount) * ($exchangeDollor);
             $totalAmount += $amount;
             $results[$dateKey] = ($results[$dateKey] ?? 0) + $amount;
         }
--- a/Modules/Report/Services/Finance/RevenueReportService.php
+++ b/Modules/Report/Services/Finance/RevenueReportService.php
@@ -75,8 +75,7 @@ class RevenueReportService
             $exchangeMonth = (date('m-y', strtotime($exchangeRate['captured_for'])));
             if ($exchangeMonth == $dateKey) {
                 $exchangeDollor = $exchangeRate['avg_rate'];
-            }
-            else {
+            } else {
                 $exchangeDollor = app(CurrencyService::class)->getCurrentRatesInINR();
             }
             $amount = ($invoice->amount) * $exchangeDollor;
--- a/Modules/Report/Services/Finance/RevenueReportService.php
+++ b/Modules/Report/Services/Finance/RevenueReportService.php
@@ -63,7 +63,7 @@ class RevenueReportService
         // ToDo:: We need to change this logic and get the exchange rate for every month.
         $exchangeRates = CurrencyAvgRate::select('avg_rate', 'captured_for')->groupBy('captured_for')->get()->toArray();
         foreach($exchangeRates as $exchangeRate) {
-            if($exchangeRate['avg_rate']){
+            if($exchangeRate['avg_rate']) {
                 $exchangeDollor =  $exchangeRate['avg_rate'];
             } else {
                 $exchangeDollor = app(CurrencyService::class)->getCurrentRatesInINR();
--- a/Modules/Report/Services/Finance/RevenueReportService.php
+++ b/Modules/Report/Services/Finance/RevenueReportService.php
@@ -59,6 +59,7 @@ class RevenueReportService
         $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'non-indian');
         $totalAmount = 0;
         $results = [];
+        $exchangeDollor = 0;
 
         // ToDo:: We need to change this logic and get the exchange rate for every month.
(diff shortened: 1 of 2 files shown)


Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -59,12 +60,19 @@ private function getParticularAmountForExport(array $particular, Object $startDa
...
-        $exchangeRates = app(CurrencyService::class)->getCurrentRatesInINR();
-
+        $exchangeRates = CurrencyAvgRate::select('avg_rate', 'captured_for')->groupBy('captured_for')->get()->toArray();
+        foreach ($exchangeRates as $exchangeRate) {
+            if ($exchangeRate['avg_rate']) {
+                $exchangeDollor = $exchangeRate['avg_rate'];
+            } else {
+                $exchangeDollor = app(CurrencyService::class)->getCurrentRatesInINR();
+            }
+        }
```
pankaj-ag, line 73: We don't need this loop or something. 

We just need a value where 
captured_for month is the same as invoice->sent->date month, 
currency  = invoice->currency

pankaj-ag (changes requested):

Modules/Report/Services/Finance/RevenueReportService.php
Himanshunegi0332: > We don't need this loop or something.
> 
> We just need a value where captured_for month is the same as invoice->sent->date month, currency = invoice->currency

Can you explain it little bit.

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -80,49 +88,26 @@ private function getParticularAmountForExport(array $particular, Object $startDa
 
     private function getParticularAmountForCommissionReceived(array $particular, Object $startDate, Object $endDate): array
     {
-        return $this->getAmountsForRevenueProceeds(Str::snake($particular['name']), $startDate, $endDate);
```
pankaj-ag, line 83: @Himanshunegi0332 why w reverted those changes ?

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -80,49 +88,26 @@ private function getParticularAmountForExport(array $particular, Object $startDa
...
-            $month = sprintf('%02d', $revenue->month);
-            $dateKey = $month . '-' . $year;
-            $totalAmount += $amount;
-            $results[$dateKey] = ($results[$dateKey] ?? 0) + $amount;
-        }
-        $results['total'] = $totalAmount;
-
-        return $results;
+        return ['total' => 0];
     }
```
pankaj-ag, line 112: Why we removed this code? Please add it back.

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -59,13 +59,21 @@ private function getParticularAmountForExport(array $particular, Object $startDa
...
+        $exchangeDollor = 0;
 
         // ToDo:: We need to change this logic and get the exchange rate for every month.
-        $exchangeRates = app(CurrencyService::class)->getCurrentRatesInINR();
-
+        $exchangeRates = CurrencyAvgRate::select('avg_rate', 'captured_for')->groupBy('captured_for')->get()->toArray();
         foreach ($invoices as $invoice) {
-            $amount = $invoice->amount * $exchangeRates;
             $dateKey = $invoice->sent_on->format('m-y');
+            foreach ($exchangeRates as $exchangeRate) {
```
pankaj-ag, line 68: Let's not run this loop again and again for each invoice. 
Let's take this loop outside of the top for loop and store the values in a variable like currency_chages [
'01-22' => 74,
'02-22' => 0,
 
]

and then use that variable.

pankaj-ag (changes requested):

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -59,17 +60,29 @@ private function getParticularAmountForExport(array $particular, Object $startDa
...
+            $exchangeMonth = (date('m-y', strtotime($exchangeRate['captured_for'])));
+            $currency_chages[] = [
+                $exchangeMonth => $exchangeRate['avg_rate']
+            ];
+        }
         foreach ($invoices as $invoice) {
-            $amount = $invoice->amount * $exchangeRates;
             $dateKey = $invoice->sent_on->format('m-y');
+            if ($exchangeMonth == $dateKey) {
+                $exchangeDollor = $exchangeRate['avg_rate'];
```
pankaj-ag, line 78: I think this should be like this right ?

```suggestion
            if (isset($currency_chages[$dateKey])) {
                $exchangeDollor = $currency_chages[$dateKey];
```

pankaj-ag: Please test by checking if the invoice amount is calculating based on the other the currency value. 

Like if a April invoice amount is 100
and April avg rate is 70 

Then the invoice value will be 7000

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -59,17 +60,29 @@ private function getParticularAmountForExport(array $particular, Object $startDa
...
-
+        $exchangeRate = [];
+        $exchangeRates = CurrencyAvgRate::select('avg_rate', 'captured_for')->groupBy('captured_for')->get()->toArray();
+        $exchangeMonth = null;
+        foreach ($exchangeRates as $exchangeRate) {
+            $exchangeMonth = (date('m-y', strtotime($exchangeRate['captured_for'])));
+            $currency_chages[] = [
+                $exchangeMonth => $exchangeRate['avg_rate']
+            ];
+        }
```
pankaj-ag, line 74: Very good now let's move this to new function with name like getAvgCurrencyRates()

pankaj-ag (changes requested):

Modules/Report/Services/Finance/RevenueReportService.php
Himanshunegi0332: $exchangeDollor = $currency_chages[$dateKey];     by doing this we will get only the month and year not the avg_rate

Modules/Report/Services/Finance/RevenueReportService.php
Himanshunegi0332: ok done.

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -37,9 +38,27 @@ public function getParticularReport(String $particularSlug, array $particular, i
...
+        $totalRate = $exchangeRate['avg_rate'];
+
+        return $totalRate;
+    }
+
     private function getParticularAmountForDomestic(array $particular, Object $startDate, Object $endDate): array
     {
-        $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'indian');
+        $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'non-indian');
```
pankaj-ag, line 61:
```suggestion
        $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'indian');
```

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -59,17 +78,24 @@ private function getParticularAmountForExport(array $particular, Object $startDa
...
+        }
 
         // ToDo:: We need to change this logic and get the exchange rate for every month.
-        $exchangeRates = app(CurrencyService::class)->getCurrentRatesInINR();
-
         foreach ($invoices as $invoice) {
-            $amount = $invoice->amount * $exchangeRates;
             $dateKey = $invoice->sent_on->format('m-y');
+            if ($exchangeMonth == $dateKey) {
+                $exchangeDollor = $this->getAvgCurrencyRates();
```
pankaj-ag, line 91: @Himanshunegi0332 please try to understand the code first and then write :).

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -37,9 +38,27 @@ public function getParticularReport(String $particularSlug, array $particular, i
         return $particular;
     }
 
+    public function getAvgCurrencyRates()
```
pankaj-ag, line 41: OKay now let's send the startDate and endDate in this function and get the currency average between those dates only.

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -59,17 +78,24 @@ private function getParticularAmountForExport(array $particular, Object $startDa
         $invoices = $this->invoiceService->getInvoicesBetweenDates($startDate, $endDate, 'non-indian');
         $totalAmount = 0;
         $results = [];
+        $exchangeMonth = null;
+        $exchangeRates = CurrencyAvgRate::select('captured_for')->groupBy('captured_for')->get()->toArray();
```
pankaj-ag, line 82:
```suggestion
        $exchangeRates = $this->getAvgCurrencyRates($startDate, $endDate);
```

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -59,17 +78,24 @@ private function getParticularAmountForExport(array $particular, Object $startDa
...
+        foreach ($exchangeRates as $exchangeRate) {
+            $exchangeMonth = (date('m-y', strtotime($exchangeRate['captured_for'])));
+        }
```
pankaj-ag, line 85:
```suggestion
       
```

Modules/Report/Services/Finance/RevenueReportService.php
```diff
@@ -59,17 +78,24 @@ private function getParticularAmountForExport(array $particular, Object $startDa
...
 
         // ToDo:: We need to change this logic and get the exchange rate for every month.
-        $exchangeRates = app(CurrencyService::class)->getCurrentRatesInINR();
-
         foreach ($invoices as $invoice) {
-            $amount = $invoice->amount * $exchangeRates;
             $dateKey = $invoice->sent_on->format('m-y');
+            if ($exchangeMonth == $dateKey) {
```
pankaj-ag, line 90:
```suggestion
            if (isset($exchangeRates[$dateKey])) { 
```

Modules/Report/Services/Finance/RevenueReportService.php
pankaj-ag, line 91:
```suggestion
                $exchangeDollor = $exchangeRates[$dateKey];
```

pankaj-ag (changes requested):

Modules/Report/Services/Finance/RevenueReportService.php
pankaj-ag: So please why this is not working. 

Also check the exchangeRates array format. It should be
[
"month-Year" => "AvgRates"
]

Modules/Report/Services/Finance/RevenueReportService.php
pankaj-ag: Sorry it was my bad.

Modules/Report/Services/Finance/RevenueReportService.php
Himanshunegi0332: ok sir

pankaj-ag (approved):

pankaj-ag (approved):

PR #2541, opened 2022-09-26
Application handover

P4NK4J:
Currently, the user requesting for application handover is not assigned the application for evaluation. Updated the same in this PR

### DIFF
--- a/Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
+++ b/Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
@@ -310,14 +310,17 @@ abstract class ApplicationController extends Controller
         return redirect()->back()->with('status', 'Your request has successfully been sent');
     }
 
-    public function acceptHandoverRequest(Application $application)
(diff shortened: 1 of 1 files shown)


Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -310,14 +310,17 @@ public function applicationHandoverRequest(Application $application)
         return redirect()->back()->with('status', 'Your request has successfully been sent');
     }
 
-    public function acceptHandoverRequest(Application $application)
+    public function acceptHandoverRequest(Request $request, Application $application)
     {
+        $scheduledPersonId = $request->get('user');
```
gautam071, line 315: Does this return the user id?

gautam071 (approved):
Looks good
One question

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
P4NK4J: Yes, it returns userID

Issue #2534, opened 2022-09-24
The wishlisted books is not working as expected


gautam071:
**Describe the bug**
- When we select the whitelisted books all the books are displaying 
<img width="1437" alt="Screenshot 2022-09-24 at 12 28 15 PM" src="https://user-images.githubusercontent.com/73119570/192084738-f46e9f55-822d-4444-b6bf-4f8245e4622b.png">

**To Reproduce**
1. Go to `Knowlagecafe-> Books`
2. Click on `Whislisted`

**Expected behavior**
- Only the whislisted books should display

Ayush8923:
This issue PR is merged. Closing it.

PR #2547, opened 2022-09-27
Fixed wishlisted books functionality

Kirannegi02:
Targets #2534 

### Description
- Fixed the wish listed books functionality.

### Checklist:


app/Http/Controllers/KnowledgeCafe/Library/BookController.php
```diff
@@ -34,10 +34,11 @@ public function index(Request $request)
             $books = auth()->user()->booksInWishlist;
         } elseif (request()->has('borrowedBook')) {
             $books = auth()->user()->booksBorrower;
+        } elseif (request()->has('category_name')) {
```
P4NK4J, line 37: Two questions, 
1. Can we use switch cases instead of these multiple else if ?
2. Why snake case is used in request variable here while in above line its camelCase ? can we have consistent parameters casing ?

P4NK4J (changes requested):
@Kirannegi02 Left some comments, please check

app/Http/Controllers/KnowledgeCafe/Library/BookController.php
Kirannegi02: Yes, I will correct it

P4NK4J (approved):
Looks good.

Ayush8923 (approved):

PR #2548, opened 2022-09-27
Add more filters in finance report

pankaj-ag:
- Added financial year filter in the tax report
- Fix a bug in financial year report

### DIFF
--- a/Modules/Invoice/Resources/views/subviews/tax-report/filters.blade.php
+++ b/Modules/Invoice/Resources/views/subviews/tax-report/filters.blade.php
@@ -1,56 +1,100 @@
-<form action="{{ route('invoice.tax-report')  }}" id="invoiceFilterForm">
+<form action="{{ route('invoice.tax-report') }}" id="invoiceFilterForm">
     <div class="d-flex">
         <div class='form-group mr-4 w-168'>
             <select class="form-control bg-light" name="region"
                 onchange="document.getElementById('invoiceFilterForm').submit();">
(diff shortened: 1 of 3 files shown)


Modules/Invoice/Resources/views/subviews/tax-report/filters.blade.php
```diff
@@ -1,56 +1,100 @@
...
+                <option {{ request()->input('month') == '08' ? 'selected=selected' : '' }} value="08">August
+                </option>
+                <option {{ request()->input('month') == '09' ? 'selected=selected' : '' }} value="09">September
+                </option>
+                <option {{ request()->input('month') == '10' ? 'selected=selected' : '' }} value="10">October
+                </option>
+                <option {{ request()->input('month') == '11' ? 'selected=selected' : '' }} value="11">November
+                </option>
+                <option {{ request()->input('month') == '12' ? 'selected=selected' : '' }} value="12">December
+                </option>
```
GauravGusain98, line 81: I would recommend using loop adn config variable for this, but letting it go as its not causing any issue.

GauravGusain98 (approved):
Looks fine

Modules/Invoice/Resources/views/subviews/tax-report/filters.blade.php
pankaj-ag: Yes @gaurav there are lot if refacting we need to do. Since this is not related to this PR I did not make any changes.

PR #2559, opened 2022-09-29
Bugfix/fixing the velocity

AmishaBisht:
Targets #2256 

### Checklist:


Modules/Project/Resources/views/index.blade.php
```diff
@@ -94,58 +94,55 @@
...
-                                        @endphp
-                                        <a class="{{ $textColor }}"
-                                            href="{{ route('project.effort-tracking', $project) }}"><i
-                                                class="mr-0.5 fa fa-external-link-square"></i></a>
-                                        <span
-                                            class="{{ $textColor }} font-weight-bold">{{ $project->velocity . ' (' . $project->current_hours_for_month . ' Hrs.)' }}</span>
+                                    @endforeach 
+                                </td>
+                                <td>
+
```
gautam071, line 100: INDENTATION

Modules/Project/Resources/views/index.blade.php
```diff
@@ -66,7 +66,7 @@
                 </thead>
                 <tbody>
                     @can('projects.view')
-                        @forelse($clients as $client)
+                        @forelse($data['clients'] as $client)
```
gautam071, line 69: Why are you using the $data['clients'] instead of $clients

amisha296: I have made the changes in the service file now the data is coming under data.

Modules/Project/Resources/views/index.blade.php
amisha296: done!

Modules/Project/Resources/views/index.blade.php
gautam071: my question is why?

gautam071 (changes requested):

Modules/Project/Resources/views/index.blade.php
amisha296: Because data is coming from data no, I have stored the data inside the data

Issue #2535, opened 2022-09-24
The borrowed books is not working as expected


gautam071:
**Describe the bug**
- When we select the borrowed books all the books are displaying 
<img width="1440" alt="Screenshot 2022-09-24 at 12 34 51 PM" src="https://user-images.githubusercontent.com/73119570/192084894-a3cd0c93-b7c8-422a-956f-c7986bcd1e72.png">

**To Reproduce**
1. Go to `Knowlagecafe-> Books`
2. Click on `Borrowed Books`

**Expected behavior**
- Only the Borrowed books should display

Ayush8923:
This issue PR is merged. Closing it.

PR #2561, opened 2022-09-29
Now solved the problem When we select the borrowed books so only the borrowed books are displaying

JCArya:
Targets #2535
Description
Now solved the problem When we select the borrowed books so only the borrowed books are displaying


gautam071 (approved):
Looks good

Issue #2533, opened 2022-09-24
Display the count of the borrowed books and wish-listed books


gautam071:
**Is your feature request related to a problem? Please describe.**
- Display the count of books preset in the whitelisted and borrowed books 
<img width="1440" alt="Screenshot 2022-09-24 at 12 28 27 PM" src="https://user-images.githubusercontent.com/73119570/192084658-deafb486-40b3-4690-9b0b-9af2f0eb452c.png">

deepankar-sajwan:
closing as task is complete
<img width="1101" alt="image" src="https://github.com/ColoredCow/portal/assets/96309135/1064dd7e-3641-4e8e-9013-6790e334ec98">

PR #2562, opened 2022-09-29
Show the count of wishlisted and borrowed books

nitinfaujdar:
Targets #2533 

Description:
Show the count of Wish listed books and borrowed books.


gautam071 (approved):
Looks good

PR #2578, opened 2022-10-08
Added two new category as stripe invoice for indian and international

pankaj-ag:
- Added two new categories to fix the P&L report.


P4NK4J (approved):
Looks good.

Issue #2602, opened 2022-10-13
Update feedback and update schedule HR applications


P4NK4J:
**Describe the task**
When we update feedback or update schedule for applications in the HR module, the page redirects back to the application listing page. We would require if the update feedback or update schedule button is clicked the redirection should be on the same application edit blade. The rest of the actions should lead to the application listing page itself.

<img width="730" alt="image" src="https://user-images.githubusercontent.com/35001775/195508003-9370939c-aa4b-4b31-8853-7471cc76fef6.png">

PR #2606, opened 2022-10-14
update feedback and schedule

kanuj365:
Targets #2602

### Description
update feedback and schedule 
- clean up the code and remove extra changes
- for some reason app.js code not going back as before, I tried many time but it was not changed.
- Issue has been resolved with the help of Abhishek Pokhriyal.

### Checklist:


Modules/HR/Http/Controllers/Recruitment/ApplicationRoundController.php
```diff
@@ -26,9 +26,7 @@ public function update(ApplicationRoundRequest $request, ApplicationRound $round
             $round->updateOrCreateEvaluation($request->validated()['round_evaluation']);
         }
 
-        $routeName = $round->application->job->type == 'internship' ? 'applications.internship.index' : 'applications.job.index';
-
-        return redirect()->route($routeName)->with('status', 'Application updated successfully!');
+        return redirect()->back()->with('status', 'Application updated successfully!');
```
P4NK4J, line 29: This will return to the edit page, for all the actions, we only need when updatefeedback and updateSchedule buttons are clicked. Please change the logic

P4NK4J (changes requested):

pokhiii:
> for some reason app.js code not going back as before, I tried many time but it was not changed

@kanuj365 Just checking about this:

1. Do you want to make `resources/js/app.js` exactly like it is in the `master` branch?
2. What have you tried so far?

P4NK4J (changes requested):
@kanuj365 All the actions are redirecting user to edit application page, even if we reject the application. We only want update Schedule button and update Feedback button to be redirected to this page, rest of them can go to the applications listing page.

Modules/HR/Http/Controllers/Recruitment/ApplicationRoundController.php
```diff
@@ -27,8 +27,12 @@ public function update(ApplicationRoundRequest $request, ApplicationRound $round
         }
 
         $routeName = $round->application->job->type == 'internship' ? 'applications.internship.index' : 'applications.job.index';
-
-        return redirect()->route($routeName)->with('status', 'Application updated successfully!');
+        $var = $request->validated();
```
P4NK4J, line 30: I see `$request->validated()` is being used on line number 24 as well. Can we define it in a single variable and then use at both the places ? Also, let's rename this to something meaningful, like `$validatedData` or so ?

P4NK4J (changes requested):
Left a comment, please check

Issue #2597, opened 2022-10-12
Code refactoring


P4NK4J:
**Describe the task**
Line number 71-87, there is a logic to show designation and domain on users' profiles. Currently its written in a very unoptimized way. We would like to have get attributes for designation and domains using the `$user->employee` object. The data can then be fetched directly using the accessors. Something like `getDesignationAttribute()` for employee model and it can be fetched using `$user->employee->designation` in the blade file. Similarly domain can go in Designation Model, and accessed using `$user->employee->designation->domain`

https://github.com/ColoredCow/portal/blob/master/Modules/User/Resources/views/profile/show-basic-details.blade.php

Reference
https://laravel.com/docs/9.x/eloquent-mutators#defining-an-accessor
https://dev.to/dalelantowork/laravel-8-accessors-and-mutators-4d8m

PR #2607, opened 2022-10-14
Code refactoring for user profile

vipul-chandan:
Targets #2597 

### Description

Code optimization is done for user profile. The logic written to show designation and domain on users' profiles is in a very unoptimized way. It should be like to have get attributes for designation and domains using the "$user->employee" object. The data can then be fetched directly using the accessors something like "getDesignationAttribute()" and "getDomainAttribute()". And then it is fetched on user profile blade file.

### Checklist:

### DIFF
--- a/Modules/User/Resources/views/profile/show-basic-details.blade.php
+++ b/Modules/User/Resources/views/profile/show-basic-details.blade.php
@@ -78,15 +78,10 @@
             <div class="form-group">
                 <label class="font-weight-bold" for="">Designation:</label>
                 <span>{{ $user->employee->designation }}</span>
-                @foreach ($designations as $designation )
-                    <span>{{ $user->employee->designation_id == $designation['id'] ? $designation['designation'] : "" }}</span>
-                @endforeach
             </div>
             <div class="form-group">
                 <label class="font-weight-bold" for="">Domain:</label>
-                @foreach ($domains as $domain )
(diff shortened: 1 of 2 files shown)


Modules/User/Resources/views/profile/show-basic-details.blade.php
```diff
@@ -78,15 +78,10 @@
...
-                @foreach ($designations as $designation )
-                    <span>{{ $user->employee->designation_id == $designation['id'] ? $designation['designation'] : "" }}</span>
-                @endforeach
             </div>
             <div class="form-group">
                 <label class="font-weight-bold" for="">Domain:</label>
-                @foreach ($domains as $domain )
-                    <span>{{ $user->employee->domain_id == $domain['id'] ? $domain['domain'] : "" }}</span>
-                @endforeach
+                <span>{{ $user->employee->domain }}</span>
```
P4NK4J, line 84: We will be removing domain column from employee table in future, hence the get attribute should fetch domain corresponding to the designation. Read about eloquent relationships a bit, and fix the logic

P4NK4J (changes requested):
Left a comment

Modules/User/Resources/views/profile/show-basic-details.blade.php
Vipul90009: @P4NK4J Sir there isn't any attribute of domain connected with designation table. Should I need to make a new column in designation

Issue #2309, opened 2022-08-29
We need to show total number of book in KnowledgeCafe module.


Ayush8923:
**Describe the task**
Go to KnowledgeCafe --> Library
The library module does not currently have a total book count. Due to this, we do not know what total number of books we have.

**Expected behavior**
We need to show total number of book in KnowledgeCafe module. As shown in the image below.
<img width="2551" alt="count" src="https://user-images.githubusercontent.com/80516839/187202166-55491717-99c1-4010-b5d8-eea70a06d6f5.png">

Ayush8923:
Working Fine.

PR #2610, opened 2022-10-15
we need to show total number of book in knowledgeCafe module

shivam333rawat:
Targets #2309

# Description
we need to show total number of book in KnowledgeCafe  module 

### Checklist:


Shivam-Samant (approved):
Lgtm.

Issue #2544, opened 2022-09-27
Search bar on opportunities page


P4NK4J:
**Describe the task**
We would require adding a search bar on the opportunities page, next to the new opportunity button.

<img width="1310" alt="image" src="https://user-images.githubusercontent.com/35001775/192431397-f55a5ecd-bf6f-4f3c-a71e-934189977a9e.png">

PR #2614, opened 2022-10-17
create-search-bar

ditya0504:
Targets #2544 

### Description
Created the search bar on the opportunity page.

### Checklist:

### DIFF
--- a/Modules/HR/Http/Controllers/Recruitment/RecruitmentOpportunityController.php
+++ b/Modules/HR/Http/Controllers/Recruitment/RecruitmentOpportunityController.php
@@ -21,9 +21,9 @@ class RecruitmentOpportunityController extends JobController
     {
         $this->authorize('list', Job::class);
         $search = request()->query('title') ?? '';
-        if($search != '') {
+        if ($search != '') {
             $jobs = Job::where('title', 'LIKE', "%$search%")
-                    ->paginate(config('constants.pagination_size'));
+                ->paginate(config('constants.pagination_size'));
         } else {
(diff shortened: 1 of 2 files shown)


Modules/HR/Http/Controllers/Recruitment/RecruitmentOpportunityController.php
```diff
@@ -20,12 +20,17 @@ public function getOpportunityType()
...
+        if ($search != '') {
+            $jobs = Job::where('title', 'LIKE', "%$search%")
+                ->paginate(config('constants.pagination_size'));
+        } else {
+            $jobs = Job::with('applications', 'applications.applicant', 'jobRequisition')
+                ->typeRecruitment()
+                ->latest()
+                ->paginate(config('constants.pagination_size'))
+                ->appends(Request::except('page'));
+        }
```
GauravGusain98, line 33: Why you have not added `->typeRecruitment()` and `->latest()` filters in the query when there is a search term? Any specific reason for that? 

Also instead doing it like this you can use the below suggestion. This way we can avoid duplicate code
```suggestion
        $search = request()->query('title') ?? '';
        
        $jobs = Job::with('applications', 'applications.applicant', 'jobRequisition')
            ->typeRecruitment()
            ->latest();
        
        if ($search != '') {
            $jobs = $jobs->where('title', 'LIKE', "%$search%")
                ->paginate(config('constants.pagination_size'));
        }
                
        $jobs = $jobs->paginate(config('constants.pagination_size'))
            ->appends(Request::except('page'));
```

GauravGusain98 (changes requested):
please check the comments

Issue #2536, opened 2022-09-24
UI issue in Home screen


AkhileshNegi:
**Describe the bug**
On the home screen under `Books` on hovering over `Recommend` the highlight is not the same size as the other two buttons namely `Your Wishlist ` and `You have read`

**To Reproduce**
1. Go to Portal home
2. Hover over `Recommend`
3. Highlight is not of same size as other buttons adjacent to same

**Expected behavior**
All buttons should have equal highlights

**Screenshots**
<img width="601" alt="image" src="https://user-images.githubusercontent.com/40158831/192086284-ba95f421-277c-47e2-a8d6-2bef45f2f9ea.png">

PR #2618, opened 2022-10-18
UI issue in Home screen

ShivanshGairola:
Targets #2536

### Description
the highlight is not the same size as the other two buttons.

### Checklist:


resources/js/components/Dashboard/UserDashboardLibrary.vue
```diff
@@ -22,9 +22,11 @@
         <li class="nav-item">
           <a
             id="recommend"
-            class="nav-link c-pointer"
+            class="nav-link c-pointer"style="
+            padding-bottom: 20px;
+            padding-top: 20px;"
```
gautam071, line 27: Can we use bootstrap class?

gautam071 (dismissed):
what is the purpose of removing the development.md file

Can you add it back?

resources/js/components/Dashboard/UserDashboardLibrary.vue
ShivanshGairola: done , I use py-3 for both top and bottom padding .

resources/js/components/Dashboard/UserDashboardLibrary.vue
```diff
@@ -22,9 +22,8 @@
         <li class="nav-item">
           <a
             id="recommend"
-            class="nav-link c-pointer"
-            @click="setActiveTile('recommend')"
-            >Recommend</a
+            class="nav-link c-pointer py-3"
```
GauravGusain98, line 25: Why you removed this `@click="setActiveTile('recommend')"`

It was calling the function `setActiveTile` when the element is clicked. Do we not need that functionality?

ShivanshGairola: Changes have been reverted

GauravGusain98 (approved):
LGTM

GauravGusain98 (changes requested):
The `recommend` button doesnot seem to have fix
image

GauravGusain98 (approved):
LGTM

GauravGusain98:
> The `recommend` button doesnot seem to have fix image

This is fixed by running `npm run dev`

Issue #2615, opened 2022-10-17
Roundwise interview duration report


P4NK4J:
**Describe the task**
Currently, we have 3 rounds, where we capture interview duration. This includes - Introductory Call, Basic Technical Interview, Detailed Technical Interview. There can be more interview rounds in the future. We capture the duration for these rounds, to get a sense, what's the average time taken by HR team member for each of these rounds interviews. Duration can be found using `scheduled_end_date` (expected end time) and `actual_end_date` (actual ending time, difference of this with expected will give duration) column in the database for the `hr_application_round` table. This table has a round wise mapping of applications, each round the application goes, it creates a new row for the evaluation. 
For average time, we would require a BarGraph in the reports section. When you click on HR->Recruitment->Reports, we need to add a white card, saying `Roundwise interview duration` clicking on which, a new page would appear,  having a bar graph. It will have the following durations plotted against the rounds, where the value would be average duration value per round. 
0-30 minutes
30-45 minutes
45-60 minutes
60+ minutes

Hovering on a bar, should appear a tooltip which would display the exact average time for each round interview.

P4NK4J:
Let's update the logic to get difference between `scheduled_date` and `actual_end_date`, this will bring us the accurate results on the duration of interview

PR #2619, opened 2022-10-18
Feature 2615 interview duration

11karangm:
Targets #2615

### Checklist:

### DIFF
--- a/Modules/HR/Routes/web.php
+++ b/Modules/HR/Routes/web.php
@@ -55,7 +55,7 @@ Route::middleware('auth')->group(function () {
             Route::get('Dailyapplicationcount', 'ReportsController@index')->name('recruitment.reports.index');
             Route::get('reportsCard', 'ReportsController@showReportCard')->name('recruitment.daily-applications-count');
             Route::get('rejected-applications', 'ReportsController@rejectedApplications')->name('recruitment.rejected-applications');
-            Route::get('applications-rounds-duration', 'ReportsController@ApplicationRoundsDuration')->name('recruitment.application-rounds');
+            Route::get('reports/applications-rounds-duration', 'ReportsController@ApplicationRoundsDuration')->name('recruitment.reports.application-rounds');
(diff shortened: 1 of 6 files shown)


Modules/HR/Routes/web.php
```diff
@@ -55,6 +55,7 @@
             Route::get('Dailyapplicationcount', 'ReportsController@index')->name('recruitment.reports.index');
             Route::get('reportsCard', 'ReportsController@showReportCard')->name('recruitment.daily-applications-count');
             Route::get('rejected-applications', 'ReportsController@rejectedApplications')->name('recruitment.rejected-applications');
+            Route::get('applications-rounds-duration', 'ReportsController@ApplicationRoundsDuration')->name('recruitment.application-rounds');
```
GauravGusain98, line 58: This way it is easier to understand what this route is doing. It is a good practice to keep the naming convention such that the name of the variable/route/controller, etc. defines its purpose
```suggestion
            Route::get('reports/applications-rounds-duration', 'ReportsController@ApplicationRoundsDuration')->name('recruitment.reports.application-rounds');
```

P4NK4J (changes requested):
Let's remove the button here, and add a title instead
image
Other than that, I have added a logic change in the issue description, because due to inconsistent data, its populating incorrect results. Let's change that.

resources/js/app.js
```diff
@@ -197,6 +197,9 @@ $(document).ready(() => {
 	if ($("#myGraph").length) {
 		roundWiseRejectionsGraph();
 	}
+	if ($("#myGraphDuration").length) {
+		applicationRoundsDurationGraph();
+	}
```
GauravGusain98, line 202: The name `myGraph` is sounding very unprofessional. It looks like that it is a dummy project. We should not use names like this. The one who has kept this name has done it incorrectly, can you please correct it.

GauravGusain98 (changes requested):
Left some feedbacks. Will be reviewing it further once all the feedbacks till now have been implemented.

Issue #2631, opened 2022-10-30
User not able to access `My-Profile` showing 500 error


Ayush8923:
**Describe the task**
- Go to My Profile.
- User not able to access `My-Profile` Showing 500 error
![image](https://user-images.githubusercontent.com/80516839/198878378-d031af47-c38d-4169-8c29-803327764d1c.png)

**Expected behavior**
- We need to solve this error. So that the user can see his profile.

PR #2641, opened 2022-11-14
bug of accessing profile fixed

Vimalnegi03:
Targets #2631

### Description
fixed the bug of accessing the profile

### Checklist:

### DIFF
--- a/Modules/User/Resources/views/profile/show-basic-details.blade.php
+++ b/Modules/User/Resources/views/profile/show-basic-details.blade.php
@@ -77,7 +77,7 @@
             <div class="form-group">
                 <label class="font-weight-bold" for="">Domain:</label>
                 @foreach ($domains as $domain )
-                    <span>{{ $domainIndex == $domain['id'] ? $domain['domain'] : "" }}</span>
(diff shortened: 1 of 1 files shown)


Modules/User/Resources/views/profile/show-basic-details.blade.php
```diff
@@ -77,7 +77,7 @@
             <div class="form-group">
                 <label class="font-weight-bold" for="">Domain:</label>
                 @foreach ($domains as $domain )
-                    <span>{{ $domainIndex == $domain['id'] ? $domain['domain'] : "" }}</span>
+                    <span>{{ $domains == $domain['id'] ? $domain['domain'] : "" }}</span>
```
Ayush8923, line 80: What data is coming in this `$domains` variable? You are comparing an array with a `$domain['id']`. And as far as I think this condition will never be true.

Ayush8923 (changes requested):
Please check the comment.

Ayush8923:
Not anyone is Working on the Feedbacks that's why Closing this PR.

Issue #2642, opened 2022-11-16
Add a search bar on the HR designations page


P4NK4J:
**Describe the task**
On the HR designations page, we would need a search bar similar to the one on HR universities page.

PR #2643, opened 2022-11-16
Added a search bar on the HR designations page

rajatsajwan:
Targets #2642

### Description
added a search bar for HR designation page

### Checklist:

### DIFF
--- a/Modules/HR/Http/Controllers/HrJobDesignationController.php
+++ b/Modules/HR/Http/Controllers/HrJobDesignationController.php
@@ -7,6 +7,8 @@ use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 use Modules\HR\Services\HrJobDesignationService;
 use Modules\HR\Entities\HrJobDesignation;
 use Illuminate\Routing\Controller;
+use Illuminate\Http\Request;
+
 
 class HrJobDesignationController extends Controller
 {
@@ -46,5 +48,10 @@ class HrJobDesignationController extends Controller
         $this->service->destroy($request, $id);
 
         return redirect()->back();
+    }
+
+    public function searchDesignation(Request $request)
+    {
+    $searchDesignation = $request->input('search');
     }
 }
--- a/Modules/HR/Http/Controllers/HrJobDesignationController.php
+++ b/Modules/HR/Http/Controllers/HrJobDesignationController.php
@@ -22,7 +22,7 @@ class HrJobDesignationController extends Controller
     }
 
     public function index(Request $request)
-    {   
+    {
         $this->authorize(HrJobDesignation::class);
 
         return view('hr.designations.index', $this->service->index($request));
--- a/Modules/HR/Http/Controllers/HrJobDesignationController.php
+++ b/Modules/HR/Http/Controllers/HrJobDesignationController.php
@@ -52,6 +52,6 @@ class HrJobDesignationController extends Controller
 
     public function searchDesignation(Request $request)
     {
-    $searchDesignation = $request->input('search');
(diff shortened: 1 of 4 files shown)


Modules/HR/Routes/web.php
```diff
@@ -145,6 +145,7 @@
         Route::post('/delete/{id}', 'HrJobDesignationController@destroy')->name('designation.delete');
         Route::get('/{id}/edit', 'HrJobDesignationController@edit')->name('designation.edit');
         Route::post('/store', 'HrJobDesignationController@storeDesignation')->name('hr-job-designation.storeJobDesignation');
+        Route::get('/search', 'HrJobDesignationController@index')->name('hr-job-designation.searchDesignation');
```
P4NK4J, line 148: We don't need to create a separate route for this, it can be used as a GET parameter in index function of the resource controller

Modules/HR/Http/Controllers/HrJobDesignationController.php
```diff
@@ -20,11 +21,11 @@ public function __construct(HrJobDesignationService $service)
         $this->service = $service;
     }
 
-    public function index()
+    public function index(Request $request)
     {
         $this->authorize(HrJobDesignation::class);
 
-        return view('hr.designations.index', $this->service->index(request()->all()));
```
P4NK4J, line 27: Why this is changed to $request ?

Modules/HR/Services/HrJobDesignationService.php
```diff
@@ -5,13 +5,20 @@
...
+use Illuminate\Http\Request;
 use Modules\HR\Entities\HrJobDomain;
 
 class HrJobDesignationService
 {
-    public function index()
+    public function index(Request $request)
     {
-        $designations = HrJobDesignation::all();
+        $searchDesignation = $request->input('search');
```
P4NK4J, line 15: what would happen if `$request->input` doesn't have a search param value ?

Modules/HR/Services/HrJobDesignationService.php
```diff
@@ -5,13 +5,20 @@
...
+        $design = HrJobDesignation::where('designation', $searchDesignation)->get();
```
P4NK4J, line 16: There may be chances, that users may not enter the correct designation, let's change this to match a string set rather than calculating exact match, try use LIKE

P4NK4J, line 16: Let's change the variable name, it doesn't make sense

resources/views/hr/designations/index.blade.php
```diff
@@ -6,6 +6,12 @@
         <div>
             <h2 class="mb-3">New Designations</h2>
         </div>
+        <div class="col-md-5">
+            <form class="d-flex" method="GET" action="{{route('hr-job-designation.searchDesignation')}}">
```
P4NK4J, line 10: You can use designation index as a action route here

P4NK4J (changes requested):

Modules/HR/Http/Controllers/HrJobDesignationController.php
rajatsajwan: ``

rajatsajwan: @P4NK4J   we are using request in service so if we dont use a request it is giving us error

Modules/HR/Services/HrJobDesignationService.php
rajatsajwan: it will show all the values

P4NK4J (approved):

PR #2645, opened 2022-11-20
[HOTFIX] - Fix issue with invoice term calculation

GauravGusain98:
### Changes 
- Fix invoice term calculation logic
- Add bank charges while calculating amount for invoice


GauravGusain98:
Self reviewed

Issue #2599, opened 2022-10-12
We need to add Features to Media Module


Kirannegi02:
**Describe the task**
- Go to Media module 
- Here all Media will be shown, a modal will be opened by clicking on Add New Post.
![Screenshot (188)](https://user-images.githubusercontent.com/99318732/195358593-b5c11594-0cde-492b-9aec-2158f1738f86.png)

**Expected behavior**
- We need to change 
    1. modal text from Add New Post to **Add New Media**.
    2. Heading from Media to **Media Gallery**.
 - We need to add text - **Uploaded by** in front of Image icon
 - Add **Count**
 
![Screenshot (191)](https://user-images.githubusercontent.com/99318732/195360033-56af5f5e-4f9b-48b6-aa12-0bac59a31c49.png)

- Go to Add New Post Modal
 - When we open the Add New Media modal, it appears as there are two modals. So, we need to adjust the modal in  such a way that it appears as a single modal.
 - 
![Screenshot (193)](https://user-images.githubusercontent.com/99318732/195368912-823f226c-4a7a-4bde-bfc3-929008201df9.png)

PR #2668, opened 2022-11-30
add Features to Media Module

ShivanshGairola:
Targets #2599

### Description
Added Features to Media Module

### Checklist:

### DIFF
--- a/Modules/Media/Http/Controllers/MediaController.php
+++ b/Modules/Media/Http/Controllers/MediaController.php
@@ -31,15 +31,19 @@ class MediaController extends Controller
      * @return RedirectResponse
      */
     public function store(MediaRequest $request)
-    {
-        $validated = $request->validated();
-        $path = config('media.path');
-        $fileName = time() . '.' . $request->file->extension();
+    {   
+        $files=$request->file;
+        if($files)
+        {
 
-        $request->file->storeAs(
-            $path,
-            $fileName,
-        );
+        $names=[];
+
+foreach ($files as $file) {
+    $validated = $request->validated();
+    $path = config('media.path');
+    $fileName = $file->getClientOriginalName();
+    $file->move($path, $fileName);
+    array_push($names, $fileName);
 
         $postData = [
             'event_name' => $validated['event_name'],
@@ -48,7 +52,8 @@ class MediaController extends Controller
             'uploaded_by' => Auth()->user()->id
         ];
         Media::create($postData);
-
+    }
+}
         return redirect(route('media.index'))->with(['message', 'status' => 'Media added successfully!']);
     }
 
--- a/Modules/Media/Http/Controllers/MediaController.php
+++ b/Modules/Media/Http/Controllers/MediaController.php
@@ -52,7 +52,7 @@ class MediaController extends Controller
                 Media::create($postData);
             }
         }
-        
+
         return redirect(route('media.index'))->with(['message', 'status' => 'Media added successfully!']);
     }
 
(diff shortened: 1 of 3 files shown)


Modules/Media/Resources/views/media/index.blade.php
```diff
@@ -78,14 +71,13 @@
...
                         <img src="{{ asset('storage/media/'.$row->file_url) }}" class="card-img-top img-fluid ">
                     @endif
                 </a>
                 <div class="card-body">
-                    <p>
-                        <td class="">
-                            <img src="{{auth()->user()->avatar}}" alt="{{auth()->user()->name}}" class="w-25 h-25 rounded-circle" data-toggle="tooltip" data-placement="top" title="{{auth()->user()->name}}">
+                    <p>Uploaded By
+                        <td class=""> <img src="{{auth()->user()->avatar}}" alt="{{auth()->user()->name}}" class="w-25 h-25 rounded-circle" data-toggle="tooltip" data-placement="top" title="{{auth()->user()->name}}">
                         </td>
```
GauravGusain98, line 81: Why you have used `<td>` tag when there is no table. Also remove the class attribute if there is no class we are using

Modules/Media/Resources/views/media/index.blade.php
```diff
@@ -78,14 +71,13 @@
                             <source src="{{asset('storage/media/'.$row->file_url)}}" type="video/mp4">
                         </video>
                     @endif
-                    @if(pathinfo($row->file_url, PATHINFO_EXTENSION) == 'jpg' || pathinfo($row->file_url, PATHINFO_EXTENSION) == 'jpeg' || pathinfo($row->file_url, PATHINFO_EXTENSION) == 'png' )
+                    @if(pathinfo($row->file_url, PATHINFO_EXTENSION) == 'jpg' || pathinfo($row->file_url, PATHINFO_EXTENSION) == 'jpeg' || pathinfo($row->file_url, PATHINFO_EXTENSION) == 'PNG' )
```
GauravGusain98, line 74: Why only `PNG` is in uppercase, while `jpg` and `jpeg` are lowercase? Can you confirm if the syntax is correct

Modules/Media/Http/Controllers/MediaController.php
```diff
@@ -32,22 +32,26 @@ public function index()
...
-        );
+            foreach ($files as $file) {
+                $validated = $request->validated();
+                $path = config('media.path');
+                $fileName = $file->getClientOriginalName();
+                $file->move($path, $fileName);
+                array_push($names, $fileName);
 
-        $postData = [
+                $postData = [
```
GauravGusain98, line 46: Please correct the indentation

GauravGusain98 (changes requested):
Please check the comments. The feedbacks mostly include formatting issues. I'll do another round of review once these issues are fixed. Will also need some help in the testing of the workflow on local

Modules/Media/Resources/views/media/index.blade.php
ShivanshGairola: Sir png image have extension in uppercase only.thats why used in upper case

Modules/Media/Http/Controllers/MediaController.php
GauravGusain98: This is not implemented

ShivanshGairola: After implementing it failing in CI check.It is showing larstan error

GauravGusain98 (changes requested):
- The text of this button is not updated.
    image
- The modal title and button text is also not updated
    image
- On Saving the media the image is not getting saved.

Modules/Media/Http/Controllers/MediaController.php
GauravGusain98: ok

GauravGusain98:
The image is not getting saved and is not getting displayed on the page. Please test it once in your system.

Issue #2650, opened 2022-11-24
Show pending invoices grouped by clients by default in invoice page


GauravGusain98:
**Enhancement Requested by**
@GauravGusain98 

**Is your feature request related to a problem? Please describe.**
Whenever a Finance team member visits the invoice page he sees the list of pending invoices for that month sorted by sent date. This makes it difficult for him to see all the pending invoices for a client in one place. To see this he needs to do some filtering using some available filter combinations 

For the Finance team being aware of the pending invoices is an important thing as if they miss following up for any pending invoice then it would be a huge loss for the company. Therefore it would be better if all these pending invoices should be available to them and grouped according to clients whenever they open the invoice page.

This will help the finance team to not struggle with filters and execute their work smoothly 

**Describe the solution you'd like**
- [x] We need to group the invoice according to the clients sorted by sent date in descending order
- [x] We need to change the default filters as per the below table

| Filter | expected default filter value |
| --- | --- |
| Year | All years |
| Month | All months |
| Status | Pending |
| Client | All clients | 

**Acceptance Criteria**
- [x] The default filters should be applied as we open the invoice page
- [x] If a filter is applied the result should be grouped according to the clients only. 
- [x] If a filter is applied and we refresh a page the filter should not be clear.

**Additional context**
This task is fulfilling our goal to achieve an effortless invoicing system for the organization.

**Contact Person for Guidance**
@GauravGusain98 @gautam071

PR #2673, opened 2022-12-02
Apply default filter invoice page

ShivanshGairola:
Targets #2650

### Description
Show pending invoices grouped by clients by default in invoice page

### Checklist:

### DIFF
--- a/Modules/Invoice/Services/InvoiceService.php
+++ b/Modules/Invoice/Services/InvoiceService.php
@@ -38,7 +38,7 @@ class InvoiceService implements InvoiceServiceContract
 
         if ($invoiceStatus == 'sent') {
             $invoices = Invoice::query()->applyFilters($filters)
-            ->orderBy('client_id','desc')
+            ->orderBy('client_id', 'desc')
             ->get();
             $clientsReadyToSendInvoicesData = [];
             $projectsReadyToSendInvoicesData = [];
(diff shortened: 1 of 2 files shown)


GauravGusain98 (changes requested):
This is half done. There was one other requirement as well that was mentioned in the issue. It was to group the invoices according to clients. This means that all the invoices of a client should come together

Modules/Invoice/Services/InvoiceService.php
```diff
@@ -38,8 +38,8 @@ public function index($filters = [], $invoiceStatus = 'sent')
 
         if ($invoiceStatus == 'sent') {
             $invoices = Invoice::query()->applyFilters($filters)
-                ->orderBy('sent_on', 'desc')
-                ->get();
+            ->orderBy('client_id', 'desc')
+            ->get();
```
GauravGusain98, line 42: This will only provide you the invoices according to clients but how we can make sure the invoices list is in descending order of the sent date. 

For each client, the invoice should be sorted by sent data. So there will be 2 levels of sorting One at the client level and the other at the client invoice level.

GauravGusain98: The same is done [here](https://github.com/ColoredCow/portal/blob/01dffc71b0adb36691f0b3fb8ff633c73fed574b/Modules/Client/Services/ClientService.php#L21). Please take reference and check how this code is running.

GauravGusain98 (approved):
LGTM

Issue #2657, opened 2022-11-28
Team Members Role in a Project


P4NK4J:
**Describe the task**
A clear and concise description of the task

**Expected behavior**
A clear and concise description of what needs to be done.

**References**
Specify any reference material / links

PR #2720, opened 2022-12-21
Feature/resource pipeline model in a project #2657

kanuj365:
Targets #2657

### Description
-  Transform the designation column to designation_id inside of this populated each id from hr_job_designation table.
-  On UI of ProjectTeamMember, all designation should available in dropdown .
-  Remove team members' roles in a project on the tooltip.

### Checklist:


Modules/Project/Database/Factories/ProjectTeamMemberFactory.php
```diff
@@ -31,7 +31,6 @@ public function definition()
             'team_member_id' => function () {
                 return User::factory()->create()->id;
             },
-            'designation' => array_rand(config('project.designation')),
```
P4NK4J, line 34: Why this is removed ? I guess this was to seed the data in designations table for the first time

Modules/Project/Resources/views/index.blade.php
```diff
@@ -93,7 +93,7 @@
                                     <td class="w-20p">
                                         @foreach ($project->getTeamMembers ?: [] as $teamMember)
                                             <span class="content tooltip-wrapper" data-html="true" data-toggle="tooltip"
-                                                title="{{ $teamMember->user->name }} - {{ config('project.designation')[$teamMember->designation] }} <br>    Efforts: {{ $teamMember->current_actual_effort }} Hours">
```
P4NK4J, line 96: Why team member designation is removed from here ?

database/migrations/2022_12_24_125621_drop_designation_from_project_team_members.php
```diff
@@ -0,0 +1,43 @@
...
+class DropDesignationFromProjectTeamMembers extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        $this->getDesignationid();
```
P4NK4J, line 18: What's the use of this line, if this is used to migrate existing designation data, can we rename this function to something meaningful ?

P4NK4J (changes requested):

Modules/Project/Database/Factories/ProjectTeamMemberFactory.php
kanuj365: yes when i'm using this it's showing error in run seeders because it was searching that column so that's why i removed that.
that error->
image

Modules/Project/Resources/views/index.blade.php
kanuj365: it's requirement of 3rd phase of task ->
On the project listing page, we would like to remove the designation of a member from the tooltip.

database/migrations/2022_12_24_125621_drop_designation_from_project_team_members.php
kanuj365: i called that function because, before remove that column it migrate all designation id

Modules/Project/Database/Factories/ProjectTeamMemberFactory.php
P4NK4J: You need to find out solution to fix this, rather than removing the line of codes. Please do that

P4NK4J (changes requested):
Left a comment please check

P4NK4J:
@kanuj365 Also there was requirement to replace the dropdown items and bring them from database in project team members page. You have simply added the items to config and not modified the dropdown. Please read the requirements and make changes accordingly

Modules/HR/Database/Seeders/ResumeScreeningEvaluationSeeder.php
```diff
@@ -20,38 +20,38 @@ public function run()
...
                 ],
 
                 [
-                    'id' => 4,
+                    // 'id' => 4,
                     'name' => 'Proceed to next round?',
                     'segment_id' => 1,
                     'marks' => 1,
                     'parent_id' => 1,
                 ],
```
P4NK4J, line 52: Why this code is commented out ? will this not impact the workflow if someone runs the seeder first time ?

Modules/Project/Database/Factories/ProjectTeamMemberFactory.php
```diff
@@ -31,7 +32,7 @@ public function definition()
...
+            'designation_id' =>  HrJobDesignation::where('designation', array_rand(config('project.designation')))->value('id'),
```
P4NK4J, line 35: If there are no designations in the table how this will work ?

Modules/HR/Database/Seeders/HRDatabaseSeeder.php
```diff
@@ -15,6 +15,7 @@ class HRDatabaseSeeder extends Seeder
     public function run()
     {
         Model::unguard();
+        $this->call(HrDesignationTableSeeder::class);
```
P4NK4J, line 18: Designations have domains column in the table, will changing the sequence of the seeders would be impacted on a fresh database here ?

P4NK4J (changes requested):

Modules/HR/Database/Seeders/ResumeScreeningEvaluationSeeder.php
kanuj365: when i run module:Seed then in this file created duplicate id. it's impacted module:Seed for that i removed this.

Modules/HR/Database/Seeders/HRDatabaseSeeder.php
kanuj365: No, it was last changes.

Modules/Project/Database/Factories/ProjectTeamMemberFactory.php
kanuj365: 1 - if there is no designation then it will return null,

kanuj365:
image
image

PR #2725, opened 2022-12-22
Make invoice number same everywhere

GauravGusain98:
### Description

- Not trimming the month and year numbers from the invoice. It will be the same everywhere
- The added year in the term. This was needed in case the billing period date range lies in between 2 years.

### Checklist:

### DIFF
--- a/Modules/Invoice/Http/Controllers/InvoiceController.php
+++ b/Modules/Invoice/Http/Controllers/InvoiceController.php
@@ -95,12 +95,12 @@ class InvoiceController extends Controller
         $invoiceNumber = $data['invoiceNumber'];
         $pdf = $this->showInvoicePdf($data);
 
-        return $pdf->inline(str_replace('-', '', $invoiceNumber) . '.pdf');
+        return $pdf->inline($invoiceNumber . '.pdf');
     }
 
     public function showInvoicePdf($data)
     {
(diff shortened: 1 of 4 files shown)


Modules/Invoice/Http/Controllers/InvoiceController.php
```diff
@@ -95,12 +95,12 @@ public function generateInvoice(Request $request)
         $invoiceNumber = $data['invoiceNumber'];
         $pdf = $this->showInvoicePdf($data);
 
-        return $pdf->inline(str_replace('-', '', $invoiceNumber) . '.pdf');
+        return $pdf->inline($invoiceNumber . '.pdf');
```
GauravGusain98, line 98: Now we don't require `-` in invoice number.

Modules/Invoice/Http/Controllers/InvoiceController.php
```diff
@@ -95,12 +95,12 @@ public function generateInvoice(Request $request)
...
     }
 
     public function showInvoicePdf($data)
     {
-        $data['invoiceNumber'] = substr($data['invoiceNumber'], 0, -5);
+        $data['invoiceNumber'] = $data['invoiceNumber'];
```
GauravGusain98, line 103: We should not trim year and month number from invoice number

P4NK4J (approved):
Looks good.

Issue #2700, opened 2022-12-20
Inactive client not getting shown when creating a new project


GauravGusain98:
**Issue raised by** 
@maniSHarma7575 

**Priority**
High as people are facing issues while creating new projects

**Describe the bug**
When all the project corresponding to a client gets `inactive` then the client automatically gets `inactive` and the client cannot be marked as active until any one of the projects is active. 

But in case we want to create a new project for a client then in the project creation list, we also need to select the client corresponding to which we are creating the new project. In this list currently, only active clients are shown therefore if we need to add a project corresponding to a client that is inactive then we are unable to add it.

**To Reproduce**
1. Go to a Client's details
2. Mark it as inactive, this will automatically mark all the projects of the client as inactive
3. Try creating a new project for the inactive from the project page
4. You will not see the client in the dropdown list

**Expected behavior**
The client dropdown list should also include inactive clients.

**Environment where the issue is occurring**
All

PR #2726, opened 2022-12-23
Inactive client not getting shown when creating a new project

race45:
Targets #2700 

### Description

    when we create a new project for inactive client . in select client option it show active client detail . I changed it and show all 
     the inactive clients.
### Checklist:

### DIFF
--- a/Modules/Project/Services/ProjectService.php
+++ b/Modules/Project/Services/ProjectService.php
@@ -23,28 +23,37 @@ class ProjectService implements ProjectServiceContract
     {
         $filters = [
             'status' => $data['status'] ?? 'active',
-            'is_amc' => $data['is_amc'] ?? 0
+            'is_amc' => $data['is_amc'] ?? 0,
         ];
 
         if ($nameFilter = $data['name'] ?? false) {
             $filters['name'] = $nameFilter;
         }
 
-        $showAllProjects = Arr::get($data, 'projects', 'my-projects') != 'my-projects';
+        $showAllProjects =
+            Arr::get($data, 'projects', 'my-projects') != 'my-projects';
 
         $memberId = Auth::id();
 
-        $projectClauseClosure = function ($query) use ($filters, $showAllProjects, $memberId) {
+        $projectClauseClosure = function ($query) use (
+            $filters,
+            $showAllProjects,
+            $memberId
+        ) {
             $query->applyFilter($filters);
             $showAllProjects ? $query : $query->linkedToTeamMember($memberId);
         };
 
         $projectsData = Client::query()
-        ->with('projects', $projectClauseClosure)
-        ->whereHas('projects', $projectClauseClosure)
-        ->orderBy('name')
-        ->paginate(config('constants.pagination_size'));
-        $tabCounts = $this->getListTabCounts($filters, $showAllProjects, $memberId);
+            ->with('projects', $projectClauseClosure)
+            ->whereHas('projects', $projectClauseClosure)
+            ->orderBy('name')
+            ->paginate(config('constants.pagination_size'));
+        $tabCounts = $this->getListTabCounts(
+            $filters,
+            $showAllProjects,
+            $memberId
+        );
         return array_merge(['clients' => $projectsData], $tabCounts);
     }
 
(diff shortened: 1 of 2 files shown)


Modules/Project/Services/ProjectService.php
```diff
@@ -109,7 +109,9 @@ private function getListTabCounts($filters, $showAllProjects, $userId)
 
     public function getClients()
     {
-        return Client::where('status', 'active')->orderBy('name')->get();
+        return Client::where('status', 'active')
+                       ->orWhere('status', 'inactive')
+                       ->orderBy('name')->get();
```
GauravGusain98, line 114: 1. There are only 2 status for client `inactive` or `active`. If you want both then why you are using where clause? It should work without any of the where clause
2. Its possible that this function is used at other places as well in the code where only active clients are needed. If you make any changes in this function definition then it will affect all other places. 
3. You have now 2 options:
    - Create a new function that will give you only active clients.
    - Modify this function in such a way that can you can pass some parameters to it that will decide if you want active client, inactive clients or all clients.  But the default behavior of this function should be the same as of now so that it does not affect other places.

Option 2 looks good as that way the code will be optimize and we don't have to make 2 functions for status filter. Please implement it as suggested

GauravGusain98 (changes requested):
Please check the comment

Modules/Project/Services/ProjectService.php
```diff
@@ -107,9 +107,20 @@ private function getListTabCounts($filters, $showAllProjects, $userId)
         return $counts;
     }
 
-    public function getClients()
-    {
-        return Client::where('status', 'active')->orderBy('name')->get();
+    public function getClients($status = 'active')
```
GauravGusain98, line 110: This is correct now

Modules/Project/Services/ProjectService.php
```diff
@@ -107,9 +107,20 @@ private function getListTabCounts($filters, $showAllProjects, $userId)
...
+                $client = Client::where('status', 'active')->orderBy('name')->get();
+        }
+            elseif($status == 'inactive'){
+                 $client = Client::where('status', 'inactive')->orderBy('name')->get();
+        }
+            else{
+                 $client = Client::where('status', 'active')
+                       ->orWhere('status', 'inactive')
+                       ->orderBy('name')->get();
+        }
```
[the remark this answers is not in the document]
GauravGusain98: Also while fetching clients, I am not sure why you are using the `where` clause. Without a where clause it will automatically fetch all the clients, right?

Modules/Project/Services/ProjectService.php
```diff
@@ -96,20 +95,25 @@ private function getListTabCounts($filters, $showAllProjects, $userId)
...
+    public function getClients($status = 'active')
     {
-        return Client::where('status', 'active')->orderBy('name')->get();
+        if ($status == 'active') {
+            $client = Client::where('status', 'active')->orderBy('name')->get();
+        } elseif ($status == 'inactive') {
+            $client = Client::where('status', 'inactive')->orderBy('name')->get();
+        } else {
+            $client = Client::where('status', 'active')->orWhere('status', 'inactive')->orderBy('name')->get();
+        }
```
GauravGusain98, line 115: It's good practice to make our code as optimized as we can. This piece of code can be more optimized by reducing the conditions like this:

If status == `all` then fetch all clients
else fetch clients where status = `$status`

GauravGusain98 (changes requested):
Please check the comments

Modules/Project/Services/ProjectService.php
GauravGusain98: For your information, the error in the CI checks is telling to add a blank line before return statement. So you need to add a blank line before each return statement. Note the there should not be any spaces in the blank line, else the CI checks will again throw error

Modules/Project/Services/ProjectService.php
```diff
@@ -94,22 +94,28 @@ private function getListTabCounts($filters, $showAllProjects, $userId)
...
 
         return $counts;
     }
 
-    public function getClients()
+    public function getClients($status = 'active')
     {
-        return Client::where('status', 'active')->orderBy('name')->get();
+        if ($status == 'all') {
+            $client = Client::where('status', 'active')->orWhere('status', 'inactive')->orderBy('name')->get();
```
GauravGusain98, line 113: This is still not fixed

GauravGusain98 (changes requested):
Left a comment please check

Modules/Project/Services/ProjectService.php
race45: I am sorry, I missed it.

GauravGusain98 (approved):
LGTM

PR #2730, opened 2022-12-26
Added client wise invoice report

GauravGusain98:
### Description

- Added chart for client wise invoice 

### Checklist:

### DIFF
--- a/Modules/Report/Http/Controllers/FinanceReportController.php
+++ b/Modules/Report/Http/Controllers/FinanceReportController.php
@@ -14,10 +14,12 @@ class FinanceReportController extends Controller
     use AuthorizesRequests;
 
     protected $service;
+    protected $reportDataService;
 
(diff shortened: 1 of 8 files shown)


Modules/Report/Http/Controllers/FinanceReportController.php
```diff
@@ -24,11 +25,21 @@ public function dashboard()
...
+        ]);
+    }
+
     public function getReportData(Request $request)
     {
         $type = $request->type;
         $filters = $request->filters;
 
-        return app(ReportDataService::class)->getData($type, $filters);
+        return app(ReportDataService::class)->getData($type, json_decode($filters, true));
```
GauravGusain98, line 43: So that we get the filters as an array instead of object

P4NK4J (approved):
Looks good.

Issue #2732, opened 2022-12-27
Add term_start_date and term_end_date column in invoice table


GauravGusain98:
**Enhancement suggested by**
@GauravGusain98 

**Is your feature request related to a problem? Please describe.**
To manage finance we should have all the information of the invoice captured with us. This will help us to fetch all the necessary information regarding an invoice whenever needed. 

One of the important such information is term of the invoice. The term of the invoice defines the period for which the invoice is raised. 

**Describe the solution you'd like**
Currently, the term of the invoice is calculated in the logic but we also want it to be a part of the invoice table as we don't want to calculate it again and again. This is because if we want to export data from the database directly like incase of analytics then we don't have this information available. 

To resolve this issue we need to add 2 columns in the invoice table. Below is the schema 
| Column Name | Type | Comment |
| --- | --- | --- |
| term_start_date | date | Nullable |
| term_end_date | date | Nullable |

**Action items**
- [ ] Create a new migration file to add these 2 columns in the table. Take reference from [here](https://nwidart.com/laravel-modules/v1/advanced-tools/artisan-commands#modulemake-migration)
- [ ] Understand how migration file works. How it is named? What is the meaning of up and down functions inside the migration file? Take help from [Laravel documentation](https://laravel.com/docs/8.x/migrations)

**Additional context**
Once this is implemented we will then work on the logic to fill values in these columns. 

**Point of contact for guidance**
@GauravGusain98

PR #2735, opened 2022-12-27
Add 2 columns in a invoices table 

rahulpanwar16:
Targets #2732

### Description
Task 1 - I have added term_start_date column and term_end_date column to the invoices table.
Task 2 -  Laravel migration are like version control for your database that allows you to create a table in your database, without actually going to the database manager such as phpmyadmin.
If you would like to specify a custom path for the generated migration, you may use the --path option when executing the make:migration command. The given path should be relative to your application's base path.
The up() is used to add new tables, columns to the database, while the down () should reverse the operations performed by the up method.

### Checklist:


database/migrations/2022_12_27_165111_add_term_start_date_and_term_end_date_to_users_table.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddTermStartDateAndTermEndDateToUsersTable extends Migration
```
GauravGusain98, line 7: The file name and class name should be corrected as we are adding these columns in invoice table not user.
```suggestion
class AddTermStartDateAndTermEndDateToInvoicesTable extends Migration
```
Also correct the file name

GauravGusain98 (changes requested):
please check the comment

database/migrations/2022_12_27_165111_add_term_start_date_and_term_end_date_to_users_table.php
GauravGusain98: Also can you add details about second action item in the PR description.
image

database/migrations/2022_12_28_114940_add_term_start_date_and_term_end_date_to__invoices_table.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('invoices', function (Blueprint $table) {
+            $table->date('term_start_date');
+            $table->date('term_end_date');
```
GauravGusain98, line 18: These 2 columns should also be nullable.

GauravGusain98 (changes requested):
Left some comments. Please check

GauravGusain98 (approved):
LGMT

Issue #2722, opened 2022-12-21
Add client country while creating a client


GauravGusain98:
**Issue Raised  By**
@rathorevaibhav 

**Is your feature request related to a problem? Please describe.**
When we are creating a client, the country is required data that should be needed because based on the country the currency and billing details of the client can be decided.

Currently, we don't ask for the client country whenever we are creating a client and due to this some of the pages throws error because the client country is needed at those places.

So we need to add a dropdown field for choosing the country of the client. 

**Describe the solution you'd like**
- Add a dropdown field `Country` on the `Add Client` page for the country similar to what we have to `Parent Organization`
- Create an entry for the new client in the `client_addresses` table and store the country value to the table

<img width="1165" alt="image" src="https://user-images.githubusercontent.com/54376313/208959778-18836a97-fa04-40e0-adf2-9202fcf16192.png">

For dropdown logic, you can take reference from the client details page under addresses section

<img width="1410" alt="image" src="https://user-images.githubusercontent.com/54376313/208960320-33ee1637-498f-4f04-9841-ce0a6c20b1c3.png">

**Additional context**
- We have a country table in the database from where we get the country list
- The country of the client is stored in `client_addresses` table.

GauravGusain98:
### Testing Feedbacks
The functionality is working fine. But there is not label for country in the form. Please add label `Select Country`.
<img width="1356" alt="image" src="https://user-images.githubusercontent.com/54376313/215064116-cb2a6c73-c6ff-42d5-a183-c8d1787c7da0.png">

PR #2738, opened 2022-12-28
Adding a dropdown field on the Add Client page for the Country.

race45:
Targets issue  #2722 

### Description

Add a dropdown field Country on the Add Client page for the country similar to what we have for Parent Organization.
Create an entry for the new client in the client_addresses table and store the country value in the table.

### Checklist:


Modules/Client/Http/Controllers/ClientController.php
```diff
@@ -33,8 +34,9 @@ public function index()
     public function create()
     {
         $this->authorize('create', Client::class);
+        $countries = Country::all();
```
GauravGusain98, line 37: This should come from the clientService. Then you don't have to add it separately inside the return() function as you have done now because the data is directly coming from clientService.

Modules/Client/Resources/views/subviews/create-client-details.blade.php
```diff
@@ -18,7 +18,8 @@
         </div>
     </div>
     <br>
-    <div class="form-row">
+    <div class="form-row" action="{{url('/')}}/ClientController@store" method="posts">
```
GauravGusain98, line 21: Why you have added `action` and `method` here? The form was getting saved earlier also. I think it was getting submitted using Javascript and not using PHP. 

I think it should work without adding action also.

Modules/Client/Resources/views/subviews/create-client-details.blade.php
```diff
@@ -58,10 +59,20 @@ class="checkbox-custom mb-1.9 mb-1.67 mr-3"
                 @endforeach
             </select>
         </div>
-
+        <div class="form-group col-md-5">
+            <label for="country_id">Country
+                <span data-toggle="tooltip" data-placement="right" title="If this client came via a channel partner then link that client from here."><i class="fa fa-question-circle"></i>&nbsp;</span>
+            </label>
+            <select name="country_id" id="country_id" class="form-coutrol">
```
GauravGusain98, line 66: Check the naming convention used in portal from [here](https://github.com/ColoredCow/portal/blob/master/docs/variable-naming-convention.md)
```suggestion
            <select name="country_id" id="countryId" class="form-control">
```

GauravGusain98 (changes requested):
Please check the comments

Modules/Client/Resources/views/subviews/create-client-details.blade.php
GauravGusain98: If its not working without action then you need to find how the form was getting submitted earlier. You can look at submit button below and search whole codebase for the id that is used in submit button. We might have written a JS code somewhere that will use the Id to submit the form

GauravGusain98:
It's not working. I've created a new client and added the country to it but in client details, the country was not showing. Also in database `client_addresses` table, no entry corresponding to that client was created.
image

Modules/Client/Http/Controllers/ClientController.php
```diff
@@ -60,11 +62,16 @@ public function show($id)
...
      * Show the form for editing the specified resource.
      */
-    public function edit(Client $client, $section = null)
+    public function edit(Client $client, $section = null, ClientAddress $clientAddress)
     {
+
+        $clientAddress = DB::table('clients')
+                               ->join('client_addresses', 'clients.client_id', '=', 'client_addresses.client_id')
+                               ->join('countries', 'client_addresses.country_id', '=', 'countries.id')
+                               ->get();
```
GauravGusain98, line 71: This is not needed as we don't want to change the view file

Modules/Client/Services/ClientService.php
```diff
@@ -42,15 +48,19 @@ public function index(array $data = [])
...
         return [
             'keyAccountManagers' => $this->getKeyAccountManagers(),
             'channelPartners' => $this->getChannelPartners(),
             'parentOrganisations' => $this->getParentOrganisations(),
+            'countries' => $countries,
         ];
     }
 
-    public function edit($client, $section)
+    public function edit($client, $section, $clientAddress)
```
GauravGusain98, line 61: Why changes are done to edit function? We just need to do changes for creating a client. Functionality to update country by editing the client details is already present and working correct. So this should not be changed at all.

Modules/Client/Services/ClientService.php
```diff
@@ -147,7 +158,14 @@ public function store($data)
         $data['status'] = 'active';
         $data['client_id'] = Client::max('client_id') + 1;
 
-        return Client::create($data);
+        $store = Client::create($data);
+        $clientAddress = new clientAddress();
+        $clientAddress->country_id = $data['country_id'];
+        $clientAddress->client_id = $data['client_id'];
```
GauravGusain98, line 164: Are you sure that this client_id is same as the id of the client that is created? We also have a client_id column in client table.

database/migrations/2023_01_05_195430_add_country_field_to_clients.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
GauravGusain98, line 1: This file is not needed. Rollback the changes that have been done due to this file in your database and then remove this file

database/migrations/2023_01_05_204310_rename_country_in_clients_table.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
```
GauravGusain98, line 2: This is also not needed. rollback the changes if any were made in database due to this file and then delete this file

database/migrations/2023_01_09_121655_drop_column_country_id_from_clients_table.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
```
GauravGusain98, line 2: File not needed

Modules/Client/Services/ClientService.php
race45: In `clients` & `client_addresses` both table have `client_id`. so which client_id I will use

GauravGusain98: When you are creating a client did you try to print you are getting $store variable?

race45: yes, i was print $store variable into that only client detail had passed. client_addresses will not pass with $store variable.

Modules/Client/Resources/views/subviews/create-client-details.blade.php
```diff
@@ -58,7 +57,17 @@ class="checkbox-custom mb-1.9 mb-1.67 mr-3"
                 @endforeach
             </select>
         </div>
-
+        <div class="form-group offset-md-6 col-md-5">
+            <label for="country_id">Country
+                <span data-toggle="tooltip" data-placement="right" title="If this client came via a channel partner then link that client from here."><i class="fa fa-question-circle"></i>&nbsp;</span>
```
GauravGusain98, line 62: I think we don't need this

Modules/Client/Services/ClientService.php
```diff
@@ -147,7 +156,14 @@ public function store($data)
         $data['status'] = 'active';
         $data['client_id'] = Client::max('client_id') + 1;
 
-        return Client::create($data);
+        $store = Client::create($data);
```
GauravGusain98, line 159: Its good practice to use meaningful variable name. 
```suggestion
        $newClient = Client::create($data);
```

Modules/Client/Services/ClientService.php
```diff
@@ -147,7 +156,14 @@ public function store($data)
...
+        $clientId = $store->id;
```
GauravGusain98, line 160: No need to create a variable for this. Directly put value on line 163

GauravGusain98 (changes requested):
Minor suggestions, please have a look

Modules/Client/Services/ClientService.php
race45: ok.

GauravGusain98 (approved):
LGTM

Issue #2739, opened 2022-12-28
When a user is an administrator, invoices can be deleted.


belwalshubham:
When a user deletes specific invoice information as an administrator, it is not deleted; however, if the user is a super-admin or another, it is deleted.

We have invoices in the finance module, and after clicking on the invoices and then on a specific invoice number, that invoice information can be deleted. On the network tab, you can see the error shown in the screenshot below.

![portl](https://user-images.githubusercontent.com/61574835/209783631-9a5d4420-97dc-43a3-ab7d-47aa8afed221.PNG)

Ayush8923:
This issue is completed. Closing it.

PR #2740, opened 2022-12-29
When the user is an administrator, invoices can be deleted

belwalshubham:
Targets #2739  

### Description
-  Invoices can be deleted when a user is designated as an administrator. It should happen by passing the module class in the destroy function of invoiceController.
reference : https://github.com/laravel/framework/issues/22847#issuecomment-521308861

### Checklist:


GauravGusain98 (approved):
LGTM

GauravGusain98 (changes requested):
This is not working. Still getting unauthorized error when deleting invoice using admin user.

GauravGusain98 (approved):
LGTM

Issue #2688, opened 2022-12-08
Fix the alignment of report tiles in the finance report 


pankaj-ag:
**Issue Raised by**
@pankaj-ag 

**Issue Description**

If you go to Finance -> Reports -> Profit and Loss page you will see the tiles are having weird with.  The padding and margin of the tabs should be equal and the UI should look symmetrical.

<img width="1059" alt="image" src="https://user-images.githubusercontent.com/10670690/206426092-7a00816d-acff-472c-b03b-2bf903444c22.png">

PR #2748, opened 2023-01-03
Fix the alignment of report tiles in the finance report

Akshiteshnaithani:
Targets #2688 

### Description
I have Fixed the alignment of report tiles in the finance report.
### Checklist:


GauravGusain98 (changes requested):
In you code changes, package.lock.json should not exist as it is not related to the issue. So remove that file from the code changes. 

In case you face issue with this, please take help from anyone near you.

Modules/Report/Resources/views/finance/profit-and-loss/index.blade.php
```diff
@@ -55,4 +56,6 @@
         </div>
     </div>
     </div>
+    </div>
+    </div>
```
GauravGusain98, line 60: The indentation looks incorrect here.

GauravGusain98 (changes requested):
Minor suggestion. Please check

Modules/Report/Resources/views/finance/profit-and-loss/index.blade.php
Akshiteshnaithani: I dont get what do you mean by tiles ? for all box and their inside content padding are same .

GauravGusain98: These 2 gap looks different in size, No?
image

Akshiteshnaithani: done

GauravGusain98 (approved):
LGTM

Issue #2678, opened 2022-12-03
Show project details link to all team members


GauravGusain98:
**Feature Requested By**
@GauravGusain98 

**Describe the task**
Whenever a team member comes to the project listing page he is not able to open his project and see project details. This creates a problem for the key account manager of that project to edit the project as he is not able to open it. 

Currently, only the admin and super admin can open the project details from the project listing page using the link as shown in the image below
<img width="623" alt="image" src="https://user-images.githubusercontent.com/54376313/205438420-84ddf491-566d-4f65-9240-42138608bb81.png">

The link on the project name is not available for users with roles below admin. This should be fixed and the link should be available to all members of that project

**Expected behavior**
- [x] The link should be available to super-admin, admin and, all members of that project

**Additional info**
The link is disabled for users with the role below admin. So there might be a permission check that has been added to enable and disable the link. We just need to remove that check.

GauravGusain98:
### Action items

- [ ] Find and understand the logic that is used to show and hide the link on the project name
- [ ] Allow the link to all types of user. So that the link of all projects should be available to you
- [ ] Read how we can print a value in Laravel using `dd()` 
- [ ] Try printing `hello` using `dd` function
- [ ] Figure out how you can get details of the login user. Try printing it on the screen
- [ ] Figure out how you can find role of the user. Try printing it on the screen
- [ ] Learn about Model/Entities in Laravel. Put your understanding in the comment in your own language.
- [ ] Learn about View In Laravel. Put your understanding in the comment in your own language.
- [ ] Learn about Controller in laravel. Put your understanding in the comment in your own language.
- [ ] Learn about relationships in laravel. [Link](https://laravel.com/docs/8.x/eloquent-relationships#main-content)
- [ ] Read about accessors in Laravel. [Link](https://laravel.com/docs/8.x/eloquent-mutators#accessors-and-mutators)
- [ ] Now based on your understanding of above topic, try to check if the user is a member of a project or not
- [ ] Based on output, show and hide the link in the project name, so that the link is shown only on projects where the user is a team member of the project 
- [ ] Compare the functionality with the expected behaviour of the task, it should work on all cases

Gauravchamoli1997:
As my understanding till now Model is used for questioning data from the table within the database as well as insert new records in the table

Gauravchamoli1997:
view provide us a convenient  way to put or place all of our HTML in a separate files. View show us user interface.

Gauravchamoli1997:
[...]

PR #2761, opened 2023-01-09
Show project details link to all team members

Gauravchamoli1997:
Targets #2678

### Description
Previously , the project link was shown for all members either he/she is member of project or not and I have to fixed this , so that the link is shown only on projects where the user is a team member of the project. I create a check for this.

### Checklist:

### DIFF
--- a/Modules/Project/Resources/views/index.blade.php
+++ b/Modules/Project/Resources/views/index.blade.php
@@ -79,8 +79,13 @@
                                 </td>
                             </tr>
                             @foreach ($client->projects as $project)
-                                 <tr>
-                                    @if('projects.update')
+                                <tr>
+                                    @can('projects.update')
+                                        <td class="w-33p">
+                                            <div class="pl-2 pl-xl-3"><a
+                                                    href="{{ route('project.show', $project) }}">{{ $project->name }}</a></div>
+                                        </td>
+                                    @else
                                         <td class="w-33p">
                                             <div class="pl-2 pl-xl-3">
                                                 @if($client->key_account_manager_id === auth()->user()->id)
@@ -90,14 +95,9 @@
                                                 @else
                                                      <span>{{ $project->name }}</span>
                                                  @endif
-                                            </div     
+                                            </div 
                                         </td>
-                                    @else
-                                        <td class="w-33p">
-                                            <div class="pl-2 pl-xl-3">{{ $project->name }}</div>
-                                        </td>
-                                    
-                                    @endif
+                                    @endcan
                                     <td class="w-20p">
                                         @foreach ($project->getTeamMembers ?: [] as $teamMember)
                                             <span class="content tooltip-wrapper" data-html="true" data-toggle="tooltip"
--- a/Modules/Project/Resources/views/index.blade.php
+++ b/Modules/Project/Resources/views/index.blade.php
@@ -88,14 +88,15 @@
                                     @else
                                         <td class="w-33p">
                                             <div class="pl-2 pl-xl-3">
-                                                @if($client->key_account_manager_id === auth()->user()->id)
-                                                 <div class="pl-2 pl-xl-3">
-                                                    <a
-                                                   href="{{ route('project.show', $project) }}">{{ $project->name }}</a></div> 
-                                                @else
-                                                     <span>{{ $project->name }}</span>
-                                                 @endif
-                                            </div 
+                                                @php
+                                                    $team_member_ids = $project->getTeamMembers->pluck('team_member_id')->toArray();
+                                                    @endphp
+                                                    @if(in_array(auth()->user()->id, $team_member_ids))
+                                                        <a href="{{ route('project.show', $project) }}">{{ $project->name }}</a>
+                                                    @else
+                                                         <div class="pl-2 pl-xl-3"> {{ $project->name }}</div>
+                                                    @endif
+                                            </div>   
                                         </td>
                                     @endcan
                                     <td class="w-20p">
--- a/Modules/Project/Resources/views/index.blade.php
+++ b/Modules/Project/Resources/views/index.blade.php
@@ -80,22 +80,24 @@
                             </tr>
                             @foreach ($client->projects as $project)
                                  <tr>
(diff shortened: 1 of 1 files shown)


GauravGusain98 (changes requested):
The functionality is not working correct. Try adding more than 1 member in a project and see. The project row is being duplicate for each member of the team. 

Please test it properly.

Modules/Project/Resources/views/index.blade.php
```diff
@@ -79,16 +79,22 @@
...
+                                 <tr>
                                     @can('projects.update')
                                         <td class="w-33p">
                                             <div class="pl-2 pl-xl-3"><a
                                                     href="{{ route('project.show', $project) }}">{{ $project->name }}</a></div>
                                         </td>
                                     @else
                                         <td class="w-33p">
-                                            <div class="pl-2 pl-xl-3">{{ $project->name }}</div>
+                                            @if($client->key_account_manager_id === auth()->user()->id)
```
GauravGusain98, line 90: This line was correct earlier. Please revert it

Modules/Project/Resources/views/index.blade.php
```diff
@@ -79,27 +79,39 @@
...
                             </tr>
                             @foreach ($client->projects as $project)
+                                @foreach ($project->getTeamMembers ?: [] as $teamMember)
+                                {{-- @dd($project->users) --}}
+                                   {{-- @dd($project) --}}
+                                   {{-- @dd($client) --}}
+                                   {{-- @dd($teamMember) --}}
+                                   {{-- @dd(auth()->user()->id) --}}
                                 <tr>
                                     @can('projects.update')
```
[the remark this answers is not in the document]
GauravGusain98: Check why we use @can function. 
Also remember that not every project member have permission `'projects.update'` but we want to show them the project link.

GauravGusain98 (changes requested):
Left some comments

Gauravchamoli1997:
@can  is used to check weather a user have specific permission or not.

Gauravchamoli1997:
Currently those who have not permission of (project.update) they can also saw the project link. 
Screenshot (13)

GauravGusain98:
> Currently those who have not permission of (project.update) they can also saw the project link.

what will happen if we not used can as we want to show URL in both case i.e if user has permission and if he does not have (just in this case there is other condition that he should be part to team)

Modules/Project/Resources/views/index.blade.php
```diff
@@ -79,17 +79,25 @@
                                 </td>
                             </tr>
                             @foreach ($client->projects as $project)
-                                <tr>
-                                    @can('projects.update')
+                                 <tr>
+                                    @if('projects.update')
```
GauravGusain98, line 83: Why you have added if condition here and what exactly is this condition doing? I am not able to understand. 

Just think like that that if I need to print "hello world", then is the below code implementation looks correct or incorrect

if (true) {
    print("hello world")
} else {
    print("hello world")
}

Gauravchamoli1997: incorrect

Gauravchamoli1997:
> > Currently those who have not permission of (project.update) they can also saw the project link.
> 
> what will happen if we don't use @can() because we want to show URL in both case i.e if user has permission and if he does not have (just in this case there is other condition that he should be part to team)

when we have to show link in both cases then why we need the permission check

Modules/Project/Resources/views/index.blade.php
Gauravchamoli1997: > > > Currently those who have not permission of (project.update) they can also saw the project link.
> > 
> > 
> > what will happen if we don't use @can() because we want to show URL in both case i.e if user has permission and if he does not have (just in this case there is other condition that he should be part to team)
> 
> when we have to show link in both cases then why we need the permission check

 as far as I understand we have to show the link in both the case whether the member has permission or not. So for this we have to give link to @can as well as inside else. The condition inside @can will be applied when the member has permission, otherwise the else condition will work in which only the project link will be visible.

GauravGusain98: So if in both condition you are printing the URL, then why you need those conditions? It should would work without them as well, no?

Gauravchamoli1997:
In @can(project.update) I don't have to change anything in this because if a user have permission to update a project then he/she already see the project link.
I have to change only in (else) condition where if a user is member of project then user see the link otherwise user only can see project name.

Modules/Project/Resources/views/index.blade.php
```diff
@@ -87,7 +87,16 @@
...
-                                            <div class="pl-2 pl-xl-3">{{ $project->name }}</div>
+                                            <div class="pl-2 pl-xl-3">
+                                                @php
+                                                    $team_member_ids = $project->getTeamMembers->pluck('team_member_id')->toArray();
+                                                    @endphp
+                                                    @if(in_array(auth()->user()->id, $team_member_ids))
+                                                        <a href="{{ route('project.show', $project) }}">{{ $project->name }}</a>
+                                                    @else
+                                                         <div class="pl-2 pl-xl-3"> {{ $project->name }}</div>
+                                                    @endif
```
GauravGusain98, line 98: This is correct now

Modules/Project/Resources/views/index.blade.php
```diff
@@ -87,7 +87,16 @@
                                         </td>
                                     @else
                                         <td class="w-33p">
-                                            <div class="pl-2 pl-xl-3">{{ $project->name }}</div>
+                                            <div class="pl-2 pl-xl-3">
+                                                @php
+                                                    $team_member_ids = $project->getTeamMembers->pluck('team_member_id')->toArray();
+                                                    @endphp
```
GauravGusain98, line 93: Correct the indentation of @php and @endphp and also other related code

GauravGusain98 (changes requested):
Looks good now, just fix the indentation

GauravGusain98 (approved):
LGTM

Issue #2757, opened 2023-01-06
Error occurring while running project seeders


GauravGusain98:
**Bug Issued By**
@Raj504 

**Describe the bug**
While running the project seeder, there is an error occurring in some windows devices. The error related to a column `reason_for_deletion` in the projects table where the column is not nullable.  Ideally it should be nullable column but in schma it is not due to which whenever we are trying to create an entry in the project table, it is expecting `reason_for_deletion` column to have a value, and as no value is passed so it is showing this error.

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/54376313/211010624-9992e84b-0c3a-4ad2-a288-4af8735163de.png">

**Solution**
- We need to make a migration that will make the `reason_for_deletion` column nullable in the table

**Environment where the issue is occurring**
Os - Windows
Environment - Local

**To Reproduce**
1. Try running project seeder. find how you can run a seeder corresponding to a particular module

**Additional context**
Due to this issue, many interns may face difficulty in project installation.

Hemrajbharti123:
![image](https://user-images.githubusercontent.com/121417679/212827477-fff38db1-2791-4c08-b97e-f6f9b8888db9.png)
 can i share the PR?

GauravGusain98:
The down method is incorrect. It should be opposite of up function. So if in up function you are making the column nullable then in the down function you should make the column not nullable instead of deleting the column

GauravGusain98:
For raising the PR, you don't need to take permission. You can open the PR anytime and give it for review once you feel that the code changes have been done. Make sure to check your code changes before assigning the PR for review.

PR #2774, opened 2023-01-17
Error occurring while running project seeders

Hemrajbharti123:
Targets #2757 
### Description
Solving error occurring while running project seeders

### Checklist:

### DIFF
--- a/database/migrations/2022_08_17_120508_add_reason_for_deletion_to_projects_table.php
+++ b/database/migrations/2022_08_17_120508_add_reason_for_deletion_to_projects_table.php
@@ -14,7 +14,7 @@ class AddReasonForDeletionToProjectsTable extends Migration
     public function up()
     {
         Schema::table('projects', function (Blueprint $table) {
-            $table->string('reason_for_deletion')->nullable();
+            $table->string('reason_for_deletion');
         });
     }
 
@@ -26,7 +26,7 @@ class AddReasonForDeletionToProjectsTable extends Migration
     public function down()
     {
         Schema::table('projects', function (Blueprint $table) {
-            $table->string('reason_for_deletion')->nullable(false);
+            $table->dropColumn('reason_for_deletion');
         });
     }
 }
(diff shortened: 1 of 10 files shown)


database/migrations/2022_08_17_120508_add_reason_for_deletion_to_projects_table.php
```diff
@@ -14,7 +14,7 @@ class AddReasonForDeletionToProjectsTable extends Migration
     public function up()
     {
         Schema::table('projects', function (Blueprint $table) {
-            $table->string('reason_for_deletion');
+            $table->string('reason_for_deletion')->nullable();
```
GauravGusain98, line 17: We don't update any existing migration file. Whenever we want to do some updates in database we create new migration file. 

Read about [why we create new migration file when making any database change](https://stackoverflow.com/questions/49132212/laravel-5-is-it-need-to-create-new-migration-everytime-while-changes-any-colum).

database/migrations/2022_08_17_120508_add_reason_for_deletion_to_projects_table.php
```diff
@@ -26,7 +26,7 @@ public function up()
     public function down()
     {
         Schema::table('projects', function (Blueprint $table) {
-            $table->dropColumn('reason_for_deletion');
+            $table->string('reason_for_deletion')->nullable(false);
```
GauravGusain98, line 29: while testing also test the down function. For this you will need to run rollback command.

GauravGusain98 (changes requested):
Please check the comment. and follow these steps 
- [x] revert all file changes 
- [x] Run the migration fresh. This will delete all the data from database and re migrate the migration files
- [x] Make changes in new migration

database/migrations/2023_01_17_162703_append_projects.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AppendProjects extends Migration
```
GauravGusain98, line 7: The name of this class should be
```suggestion
class MakeReasonForDeletionNullableInProjectTable extends Migration
```
also update the file name according to it. make sure the file name should be in snake case only

GauravGusain98 (changes requested):
looks fine. Just a minor suggestion

GauravGusain98 (approved):
LGTM

Issue #2773, opened 2023-01-17
Ordering Client dropdown by name


GauravGusain98:
**Enhancement suggested by**
@GauravGusain98 

**Is your feature request related to a problem? Please describe.**
In any application, the UI should be user-friendly i.e user should not find any difficulty using different components in the application. In the portal, any user who wants to edit project details faces difficulty finding the client while changing the client from the dropdown. Currently, the client list is not ordered by name which is causing trouble, if we sort the list by client name in ascending order then it would be easier to find the client for that list. 

So we want to order the client list by client name in ascending order.

**To replicate the problem**
- Go to edit project screen 
    <img width="1295" alt="image" src="https://user-images.githubusercontent.com/54376313/212808789-f168f6cd-0455-4540-a29a-af6fea0cf2e8.png">
- Check the client dropdown and see the client order, it is not sorted by names

**Additional context**
This task will improve the user experience while using the portal. 

**Point of Contact for Guidance**
@GauravGusain98

GauravGusain98:
### Action Items
**Basics - 1 pointer**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function
- [ ] On the project details page print details of the project using `dd` function
- [ ] Read about HTML forms and how we submit forms. Also, share the use of the below provide items
    - [ ] Action attribute in the form
    - [ ] Method attribute in the form 
    - [ ] Name attribute in any HTML tag and its importance while form submission

**Sorting - 1 pointer**
- [ ] Figure out the function from where we are fetching the client's details/name and passing them to frontend
- [ ] Read about [Laravel queries](https://laravel.com/docs/8.x/queries) and some basic queries like
    - [ ] [Running Database Queries](https://laravel.com/docs/8.x/queries#running-database-queries)
    - [ ] [Basic Where Clauses](https://laravel.com/docs/8.x/queries#basic-where-clauses)
    - [ ] [Select Statements](https://laravel.com/docs/8.x/queries#select-statements)
[...]

PR #2776, opened 2023-01-18
ordering client dropdown

nishant22029:
Targets #2773 

### Description
In this PR  ,I updated the query that is used to fetch client details in a way that is to order the client's result by client name in ascending order.
 The client list is not ordered by name which is causing trouble, if we sort the list by client name in ascending order then it would be easier to find the client for that list.

### Checklist:

### DIFF
--- a/Modules/Project/Http/Controllers/ProjectController.php
+++ b/Modules/Project/Http/Controllers/ProjectController.php
@@ -2,7 +2,7 @@
 
 namespace Modules\Project\Http\Controllers;
 
-use Google\Service\AnalyticsData\OrderBy;
+// use Google\Service\AnalyticsData\OrderBy;
 use Illuminate\Routing\Controller;
 use Modules\Client\Entities\Client;
 use Modules\Project\Entities\Project;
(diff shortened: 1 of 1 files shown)


Modules/Project/Http/Controllers/ProjectController.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Modules\Project\Http\Controllers;
 
+use Google\Service\AnalyticsData\OrderBy;
```
GauravGusain98, line 5: This is not needed. Why it was imported?

GauravGusain98 (changes requested):
Looks good minor suggestion

Modules/Project/Http/Controllers/ProjectController.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Modules\Project\Http\Controllers;
 
+// use Google\Service\AnalyticsData\OrderBy;
```
GauravGusain98, line 5: Commenting code is not a good practice. Remove this line as it was not there earlier.

GauravGusain98 (changes requested):
Check the comment

Modules/Project/Http/Controllers/ProjectController.php
nishant22029: ok sir

GauravGusain98 (approved):
LGTM

Issue #2690, opened 2022-12-08
Incorrect FTE calculation for project whose billing date starts in between of month


GauravGusain98:
### Bug Raised By
@GauravGusain98 

### Background
For ColoredCow Full Time Effort(FTE) of an employee is very important. FTE is used to check the engagement of employees and can also be used to plan teams for projects. 

The expected hours from an employee is 8 hours per day. And therefore the formula of FTE is
 ```
FTE = Sum of hours booked by an employee in all the projects / (Total number of days in month till today  * 8)
 ```
### Describe the bug
The user FTE is shown in the user dashboard (Steps provided in the next section). Ideally, the FTE is calculated for a current month that is from the first day of the month to the end of the month, but in some projects where the billing happens in between the FTE calculation comes incorrect. 

Due to this for some users, the FTE is not getting calculated correctly.

### How to open the user dashboard
To see the user's FTE, one can click on the image of his in the project list or can open the Working Staff option from the HR dropdown menu as shown below

<img width="400" alt="image" src="https://user-images.githubusercontent.com/54376313/206443793-21b7f15b-191b-4e28-ae0a-0865c49fb6e8.png">
<img width="400" alt="image" src="https://user-images.githubusercontent.com/54376313/206443691-86a97729-333e-4139-99de-ebc694c99b1a.png">

### To Reproduce
1. Run project module seeder. This will create a new client and project in portal. Check [here](https://nwidart.com/laravel-modules/v6/advanced-tools/artisan-commands#moduleseed) how to run seeder for a specific module
3. Go to the newly created client 
4. Edit the client and fill details of the client
5. In the financial details choose the billing period `monthly` and billing data to `1`
6. Go to the newly created project of that client
7. Check the team member's booked hours. They must have some hours booked for the month
8. Now Go to the user dashboard of that team member 
9. See his FTE. The FTE should come correct.
10. Now again edit client details and set the billing date to the current date 
11. Again see the FTE in the user dashboard the FTE is changed now. This should not happen

### Expected behavior
- [x] The FTE calculation in the user dashboard should not depend on the project billing date. 

### Environment details where the bug is occurring
 - OS: Ubuntu, Mac, Window
 - Environment - Production, Staging, Local

### Additional context
- You need to create a function that calculates the FTE for a user for that month

### Contact person for guidance
@GauravGusain98

GauravGusain98:
### Action items
- [ ] Find the code that is showing FTE of a team member in the Team member details page
[...]

PR #2778, opened 2023-01-19
 bugfix/incorrect FTE calculation where billing date starts in between of month

rahulpanwar16:
Targets #2690 

Description:

I have correct the FTE calculation of the user if the project billing date lies in between the month. 
To do so I have created new variable($firstDayOfMonth) whose value is 2022/01/01 and declared that variable in the FTE calculation function so that in variable $daysTillToday  , workingDaysList() can pass the value of  $firstDayOfMonth and be able to return the desired dates of the working days.

For the hours booked there is a CurrentActualEffort() who returns to the projectTeamMemberEffort() where 'added on' ,'>=',month_start_date ->sum('actual effort'). Here, month_start_date represents project billing date and I have replaced it with $firstDayOfMonth so that FTE calculation should not depend on the project billing date. Therefore, it will sum the actual effort from day 1st to the current date from the database table name "project_team_member_effort". 

The changes are required because the calculation of FTE is from the month start date to the month end date. 
This PR resolves the behavior of the FTE when project billing date lies in between of the month  as the FTE value is now constant when you add any random numbers till the current date of that project. 

 Checklist:

### DIFF
--- a/Modules/Project/Entities/ProjectTeamMember.php
+++ b/Modules/Project/Entities/ProjectTeamMember.php
@@ -47,19 +47,21 @@ class ProjectTeamMember extends Model
 
     public function getCurrentActualEffortAttribute()
     {
-        return $this->projectTeamMemberEffort()->where('added_on', '>=', $this->project->client->month_start_date)->sum('actual_effort');
+        $firstDayOfMonth = date('y-m-01');
+        return $this->projectTeamMemberEffort()->where('added_on', '>=',$firstDayOfMonth)->sum('actual_effort');
     }
 
     public function getCurrentExpectedEffortAttribute()
     {
         $project = new Project;
         $currentDate = today(config('constants.timezone.indian'));
+        $firstDayOfMonth = date('y-m-01');
 
         if (now(config('constants.timezone.indian'))->format('H:i:s') < config('efforttracking.update_date_count_after_time')) {
             $currentDate = $currentDate->subDay();
         }
 
-        $daysTillToday = count($project->getWorkingDaysList($this->project->client->month_start_date, $currentDate));
+        $daysTillToday = count($project->getWorkingDaysList($firstDayOfMonth,$currentDate));
 
         return $this->daily_expected_effort * $daysTillToday;
     }
@@ -81,12 +83,13 @@ class ProjectTeamMember extends Model
     {
         $project = new Project;
         $currentDate = today(config('constants.timezone.indian'));
+        $firstDayOfMonth = date('y-m-01');
 
         if (now(config('constants.timezone.indian'))->format('H:i:s') < config('efforttracking.update_date_count_after_time')) {
             $currentDate = $currentDate->subDay();
         }
 
-        $daysTillToday = count($project->getWorkingDaysList($this->project->client->month_start_date, $currentDate));
+        $daysTillToday = count($project->getWorkingDaysList($firstDayOfMonth,$currentDate));
         if ($daysTillToday == 0) {
             return 0;
         }
--- a/Modules/Project/Entities/ProjectTeamMember.php
+++ b/Modules/Project/Entities/ProjectTeamMember.php
@@ -45,31 +45,32 @@ class ProjectTeamMember extends Model
         return $query->whereNull('ended_on');
     }
 
-    public function getCurrentActualEffortAttribute()
+    public function getCurrentActualEffortAttribute($startDate = null)
     {
-        $firstDayOfMonth = date('y-m-01');
+        $startDate = $startDate ?? $this->project->client->month_start_date;
 
-        return $this->projectTeamMemberEffort()->where('added_on', '>=', $firstDayOfMonth)->sum('actual_effort');
+        return $this->projectTeamMemberEffort()->where('added_on', '>=', $this->project->client->month_start_date)->sum('actual_effort');
     }
 
-    public function getCurrentExpectedEffortAttribute()
+    public function getCurrentExpectedEffortAttribute($startDate = null)
     {
         $project = new Project;
         $currentDate = today(config('constants.timezone.indian'));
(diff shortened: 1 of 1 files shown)


Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -47,19 +47,22 @@ public function scopeActive($query)
...
         $project = new Project;
         $currentDate = today(config('constants.timezone.indian'));
+        $firstDayOfMonth = date('y-m-01');
 
         if (now(config('constants.timezone.indian'))->format('H:i:s') < config('efforttracking.update_date_count_after_time')) {
             $currentDate = $currentDate->subDay();
         }
 
-        $daysTillToday = count($project->getWorkingDaysList($this->project->client->month_start_date, $currentDate));
+        $daysTillToday = count($project->getWorkingDaysList($firstDayOfMonth, $currentDate));
```
GauravGusain98, line 65: We needed this for calculating the Velocity of the project. The velocity is calculated from the billing date. So this should not be changed

Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -47,19 +47,22 @@ public function scopeActive($query)
 
     public function getCurrentActualEffortAttribute()
     {
-        return $this->projectTeamMemberEffort()->where('added_on', '>=', $this->project->client->month_start_date)->sum('actual_effort');
+        $firstDayOfMonth = date('y-m-01');
+
+        return $this->projectTeamMemberEffort()->where('added_on', '>=', $firstDayOfMonth)->sum('actual_effort');
     }
```
GauravGusain98, line 54: This function is use to get the effort of user for the billing cycle i.e we need to use the billing date here.

GauravGusain98: Maybe you can just make this function take a $startDate argument whose initial value will be null. inside the function just check if the $startDate is null then assign it the value `$this->project->client->month_start_date` else the value should be $startDate only.

Modules/Project/Entities/ProjectTeamMember.php
GauravGusain98: Maybe you can just make this function take a $startDate argument whose initial value will be null. inside the function just check if the $startDate is null then assign it the value $this->project->client->month_start_date else the value should be $startDate only.

Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -81,12 +84,13 @@ public function getFteAttribute()
...
 
         if (now(config('constants.timezone.indian'))->format('H:i:s') < config('efforttracking.update_date_count_after_time')) {
             $currentDate = $currentDate->subDay();
         }
 
-        $daysTillToday = count($project->getWorkingDaysList($this->project->client->month_start_date, $currentDate));
+        $daysTillToday = count($project->getWorkingDaysList($firstDayOfMonth, $currentDate));
         if ($daysTillToday == 0) {
             return 0;
         }
```
GauravGusain98, line 96: On line 98 just pass $firstDayOfMonth to `$this->getCurrentExpectedEffortAttribute()` and it will give you the booked hours from first day of the month

Modules/Project/Entities/ProjectTeamMember.php
GauravGusain98: Do the same with the function `getExpectedEffortTillTodayAttribute()` at line 70

GauravGusain98 (changes requested):
Great, you are very close. Left some suggestion please check

Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -45,15 +45,18 @@ public function scopeActive($query)
         return $query->whereNull('ended_on');
     }
 
-    public function getCurrentActualEffortAttribute()
+    public function getCurrentActualEffortAttribute($startDate = null)
     {
+        $startDate = $startDate ?? $this->project->client->month_start_date;
```
GauravGusain98, line 50: Nice. You know why we user `??` right?

Modules/Project/Entities/ProjectTeamMember.php
GauravGusain98: On line 99, we want to calculate the hours booked by user from first day of the month. So, also pass the firstDateOfMonth to `$this->current_actual_effort`

GauravGusain98 (changes requested):
minor suggestion

Modules/Project/Entities/ProjectTeamMember.php
rahulpanwar16: Yes, ?? is a null coalescing operator which means it will check if the value of $startDate is null or not and if null then it will assign a value to the $startDate

Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -81,17 +85,18 @@ public function getFteAttribute()
...
         }
 
-        $daysTillToday = count($project->getWorkingDaysList($this->project->client->month_start_date, $currentDate));
+        $daysTillToday = count($project->getWorkingDaysList($firstDayOfMonth, $currentDate));
         if ($daysTillToday == 0) {
             return 0;
         }
 
-        return round($this->current_actual_effort / ($daysTillToday * config('efforttracking.minimum_expected_hours')), 2);
+        return round($firstDayOfMonth / ($daysTillToday * config('efforttracking.minimum_expected_hours')), 2);
```
GauravGusain98, line 99: Why numerator is number of days, it should be actual effort that you will get from the existing function, you just need to pass the start date to it. That's why we added the startDate field to above functions.

GauravGusain98 (changes requested):

GauravGusain98 (approved):
LGTM

Issue #2736, opened 2022-12-27
Phase 1.1 CodeTrek Module


Ayush8923:
**Feature Suggested By**
@Ayush8923 

**Parent Issue:-** https://github.com/ColoredCow/portal/issues/2704

**Related Issue**
https://github.com/ColoredCow/portal/issues/2734

**Is your feature request related to a problem? Please describe.**
In ColoredCow, CodeTrek is a journey that most of the interns do. In it, they learn the basics of software development and hone their skills. Now, we want to track and capture this journey of all the interns in Portal. This will help us to monitor the progress of each individual and see where we can or they can do better. 

For this, we need to introduce Codetrek Module in Portal.

**Point of contact for guidance**
@GauravGusain98

GauravGusain98:
### Action Items
**Basics - 1 pointer**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function

**Laravel Modules - 1 pointer**
- [ ] Read about the [Laravel module](https://nwidart.com/laravel-modules/v6/basic-usage/creating-a-module) and what is benefit of it. Share your understanding in your own language in the task comment. 
- [ ] Create a new module in Portal Project with the name CodeTrek

**Add Home page - Pointer**
- [ ] Add CodeTrek option in Menu bar as shown in below image
    ![image](https://user-images.githubusercontent.com/80516839/209699220-878efd98-61e3-4b55-ae5b-09ee912ae8dc.png)
- [ ]  When Clicking on the `CodeTrek` Module link, then a page should open, in which the header and its center should have `Coming Soon` written.
    ![image](https://user-images.githubusercontent.com/80516839/209700206-9babd74c-516c-4284-af52-44e80a295ade.png)

nishant22029:
models
•Models are used for creating and managing database tables.
•Models are used to query the database, insert, update and delete data.
•Models allows developers to write less code and make their applications more efficient and maintainable.
[...]

PR #2782, opened 2023-01-21
phase 1.1 codetrek module

nishant22029:
Targets #2736 

### Description
Created a new module in Portal Project with the name CodeTrek ,to track and capture journey of all the interns in Portal. This will help us to monitor the progress of each individual and see where we can or they can do better.


resources/views/layouts/navbar.blade.php
```diff
@@ -148,6 +148,12 @@ class="caret"></span>
         </li>
     @endcan
 
+    @can('codetrek.view')
```
GauravGusain98, line 151: You have added this check for permission but we don't have any such permission yet, so this option will not available for users other than superadmin. For now please remove this check

GauravGusain98 (changes requested):
Please remove package.lock.json as this is used only when you have added any new package.

resources/views/layouts/navbar.blade.php
nishant22029: ok sir

GauravGusain98:
The `package.lock.json` is not removed yet.

resources/views/layouts/navbar.blade.php
nishant22029: sir should i have to remove package.json file from codetrek module

GauravGusain98 (approved):
LGTM

PR #2795, opened 2023-01-27
Feature/2722/add client country

race45:
Targets #2722 

### Description

 add one label for country dropdown.

### Checklist:


Modules/Client/Resources/views/subviews/create-client-details.blade.php
```diff
@@ -58,6 +58,7 @@ class="checkbox-custom mb-1.9 mb-1.67 mr-3"
             </select>
         </div>
         <div class="form-group offset-md-6 col-md-5">
+            <label for="channel_partner_id">Select Country</label>
```
GauravGusain98, line 61: Read about what is the use of `for` attribute and use the correct value here
```suggestion
            <label for="channel_partner_id">Select Country</label>
```

GauravGusain98 (changes requested):

GauravGusain98 (approved):
LGTM

Issue #2796, opened 2023-01-27
Error occurring in Monthly Tax report page


GauravGusain98:
**Issue Raised By**
@mohitsharma-22 

**How to Replicate**
1. Go to finance->reports menu from the menubar
2. Select `Monthly Tax` option
    ![Image](https://user-images.githubusercontent.com/54376313/215080397-b1f900b5-67bd-4c3a-b282-b0d5bffe76c9.png)
3. Select the `all financial` option 
    ![Image](https://user-images.githubusercontent.com/54376313/215080616-82c6a592-fccb-47c5-892b-04dc23f790f8.png) 
4. Check the error 
    ![Image](https://user-images.githubusercontent.com/54376313/215080720-0b090a87-5ea9-4e31-a417-575b21a7eb7a.png)

**Additional Information**
Due to this issue, we are not able to do our GST reporting. So it needs to be fixed on a priority

**Point of contact for guidance**
@GauravGusain98

PR #2799, opened 2023-01-31
2796/fixed-error-in-monthly-tax-report-page

rajneeshweb:
Targets #2796

### Description

Changes are made to fix error in monthly tax report page.
### Checklist:

### DIFF
--- a/Modules/Invoice/Entities/Invoice.php
+++ b/Modules/Invoice/Entities/Invoice.php
@@ -114,7 +114,6 @@ class Invoice extends Model implements Auditable
         if ($invoiceYear = Arr::get($filters, 'invoiceYear', '')) {
             $query = $query->InvoiceInaYear($invoiceYear);
         }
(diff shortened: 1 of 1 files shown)


Modules/Invoice/Entities/Invoice.php
```diff
@@ -70,7 +70,7 @@ public function scopeClient($query, $clientId)
     }
     public function scopeInvoiceInaYear($query, $invoiceYear)
     {
```
GauravGusain98, line 72: The whereBetween function is used to compare between 2 times. But in case of all years can you check what value we are getting in $invoiceYear? If it is not a number then you should remove the whereBetween filter incase of all years and apply it only when you are getting a valid value in $invoiceyear

GauravGusain98 (changes requested):

GauravGusain98 (approved):
LGTM

PR #2800, opened 2023-01-31
Feature/zero expected effort mail coming for inactive projects

kanuj365:
Targets ##2787

### Description
- A bug was generated because we're getting all project in which all types of projects whose status active, inactive, halted we are getting.
- Now we fix that by allowing status only active.

### Checklist:


kanuj365:
**### Basics Concept**
 **Models in Laravel** -> 
In laravel a Modal represents a database table and provides an easy way to interact with that table. It can also be used to insert, retrieve and update data in the database.to create a model we can use the artisan command-
(php artisan make:modal ModalName)

**Views in Laravel** -> 
In laravel Views are used to represent the HTML content that is sent to the browser. and it is also used to display the dynamic data by controller. The views contain plain html with this also includes Blade templates.
  public function show()
  {
return view('welcome')
}

**Routes in Laravel** ->
 In laravel, Routes define the url structure of our application and map urls to actions in our controllers. The routes are defined in the "web.php". here are different types of methods are available like 

Route ::post() when we try to submit the form we can use post method to maps a url to a function.
Route:: put()  when we try to update the a resources on the server then it is used.
Route::get()  when we try get data from database then we use this get method.
Route::resources() when we need a set of routes in controller for a crud operation then we can use this. etc.

**Controller in Laravel**->
 In laravel, a controller is a class that defines HTTP endpoint behavior. it handle the incoming http requests and returning appropriate response. even Controller perform different tasks like data validation, accessing models to retrieve data and returning views  .
here is different types of controller like-
Single action controller 
Resource controllers

**Services files in Laravel** ->
 In laravel , a service file use for wide range of tasks, such as processing data , shared  logic and it can also be used to share logic between multiple parts of the application, such as controllers, models and other services.

**How we can print a value in Laravel using dd()**->
public function hello($id)
{
$showclient=Client::where('id',$id)->get();
dd($showclient);
}
in this function i'm getting all client whose id is equal to id i.e. passing as parameter and i can see that by dd.

**printing hello world using the dd function**->
 dd("hello world") main php file
 @dd("hello world")->in blade file
 

### Scheduling 
 Scheduling custom command-> Laravel provides a convenient way to schedule repetitive tasks by using Task Schedular, -
how we can create this->
1- create a console command by using this command (php artisan make:command MyCommand).
2- after the command run a file will be generate where a handle function will be available in MyCommand class.
3- In handle function you can define your task .
4- Open the kernal.php file and in the schedule method add a call to the command method, passing the name of the console command you want to schedule and use the cron or every method to set the schedule for running the command ex-($schedule->command('project:zero-expected-hours-in-project')->weekly()->tuesdays()->at('11:00');).
5-Run the following command to register the scheduled tasks. (php artisan schedule:run).

kanuj365:
> `understanding of the workflow`

image

GauravGusain98 (approved):

Issue #2794, opened 2023-01-26
Phase 1.3: CodeTrek Module


GauravGusain98:
**Feature Suggested By**
@Ayush8923 

**Parent Issue:-** https://github.com/ColoredCow/portal/issues/2704

**Related Issue**
https://github.com/ColoredCow/portal/issues/2734
https://github.com/ColoredCow/portal/issues/2736
https://github.com/ColoredCow/portal/issues/2743

**Is your feature request related to a problem? Please describe.**
In ColoredCow, CodeTrek is a journey that most of the interns do. In it, they learn the basics of software development and hone their skills. Now, we want to track and capture this journey of all the interns in Portal. This will help us to monitor the progress of each individual and see where we can or they can do better. 

For this, we have introduced Codetrek Module and created a form to add applicants in previous the [task](https://github.com/ColoredCow/portal/issues/2743). Now we will create DB schema to store details corresponding to applicant

**Point of contact for guidance**
@GauravGusain98

GauravGusain98:
### Action Items
**Basics - 1 pointer (Skip if you've already done these in your previous tasks)**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function

**Database Migrations - 1 pointer**
- [ ] Read about [Laravel Migration](https://laravel.com/docs/8.x/migrations#creating-tables). Write your understanding in the comments. Also mention the use of `up` and `down` function in migration file.
- [ ] Create a table `codetrek_applicants` in the employee portal database using this [Schema](https://docs.google.com/presentation/d/1WC_lr2P1C6O6L6pobD2sxtL-A3YIfDr8AYOeU1x9CVQ/edit#slide=id.g1bc21c5bde4_0_5)

**Form Submission - 1 pointer**
- [ ] Read about how forms are submitted in Laravel. You can take reference from [here](https://vegibit.com/how-to-set-up-form-submission-in-laravel/)
- [ ] Now make a function in Controller to store the applicant data 
[...]

PR #2801, opened 2023-01-31
phase 1.3 module added

nishant22029:
Targets #2794 

### Description

I have created codetrek_applicants migration and table , a controller with store() that will store applicants
data into table. Also created service file to fetch the applicants data and sent to codetrek page.
### Checklist:

### DIFF
--- a/Modules/CodeTrek/Config/config.php
+++ b/Modules/CodeTrek/Config/config.php
@@ -3,7 +3,13 @@
 return [
     'name' => 'CodeTrek',
     'status' => [
-        'active' => 'Active',
-        'inactive' => 'Inactive'
+        'active' =>[
+            'label' => 'Active',
+            'slug'  => 'active'
+        ],
+        'inactive' =>[
+            'label' => 'Inactive',
+            'slug'  => 'inactive'
+        ],
     ]
 ];
--- a/Modules/CodeTrek/Config/config.php
+++ b/Modules/CodeTrek/Config/config.php
@@ -1,5 +1,9 @@
 <?php
 
 return [
-    'name' => 'CodeTrek'
+    'name' => 'CodeTrek',
+    'status' => [
+        'active' => 'Active',
+        'inactive' => 'Inactive'
+    ]
 ];
--- a/Modules/CodeTrek/Entities/CodeTrekApplicant.php
+++ b/Modules/CodeTrek/Entities/CodeTrekApplicant.php
@@ -6,7 +6,7 @@ use Illuminate\Database\Eloquent\Model;
 
 class CodeTrekApplicant extends Model
 {
-   protected $fillable=[
-    'first_name','last_name','email','github_user_name','phone','course','start_date','graduation_year','university'
-   ];
+   protected $fillable = [
+      'first_name', 'last_name', 'email', 'github_user_name', 'phone', 'course', 'start_date', 'graduation_year', 'university'
+  ];
 }
--- a/Modules/CodeTrek/Entities/CodeTrekApplicant.php
+++ b/Modules/CodeTrek/Entities/CodeTrekApplicant.php
@@ -6,7 +6,5 @@ use Illuminate\Database\Eloquent\Model;
 
 class CodeTrekApplicant extends Model
 {
-    protected $fillable = [
-       'first_name', 'last_name', 'email', 'github_user_name', 'phone', 'course', 'start_date', 'graduation_year', 'university'
-  ];
+  protected $guarded = [];
 }
--- a/Modules/CodeTrek/Entities/CodeTrekApplicant.php
+++ b/Modules/CodeTrek/Entities/CodeTrekApplicant.php
@@ -6,5 +6,5 @@ use Illuminate\Database\Eloquent\Model;
 
 class CodeTrekApplicant extends Model
 {
-     protected $guarded = [];
+    protected $guarded = [];
 }
--- a/Modules/CodeTrek/Entities/CodeTrekApplicant.php
+++ b/Modules/CodeTrek/Entities/CodeTrekApplicant.php
@@ -6,5 +6,5 @@ use Illuminate\Database\Eloquent\Model;
 
 class CodeTrekApplicant extends Model
 {
-  protected $guarded = [];
+     protected $guarded = [];
 }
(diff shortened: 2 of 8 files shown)


GauravGusain98:
### Testing Feedbacks
1. Lets make these field as not required, as they are also not required in database level
    image
2. Good job, You've added the success banner, I think it can be made better if you use vue-toastification toggle as we have used in projects edit page. Just figure out how it is being used there and implement it here as well
    image

nishant22029:
I am stuck with this task to dispaly success banner using vue-toastification toggle ,please help on this sir.
image

Modules/CodeTrek/Entities/CodeTrekApplicant.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Modules\CodeTrek\Entities;
+
+use Illuminate\Database\Eloquent\Model;
+
+class CodeTrekApplicant extends Model
+{
+    protected $fillable = [
+       'first_name', 'last_name', 'email', 'github_user_name', 'phone', 'course', 'start_date', 'graduation_year', 'university'
+  ];
```
GauravGusain98, line 11: If you want all the fields to be fillable then you can use this 
`protected $guarded = []`

Modules/CodeTrek/Http/Controllers/CodeTrekController.php
```diff
@@ -24,10 +32,23 @@ public function create()
...
+        $applicant->first_name = $request['first_name'];
+        $applicant->last_name = $request['last_name'];
+        $applicant->email = $request['email_id'];
+        $applicant->phone = $request['phone'];
+        $applicant->github_user_name = $request['github_username'];
+        $applicant->course = $request['course'];
+        $applicant->start_date = $request['start_date'];
+        $applicant->graduation_year = $request['graduation_year'];
+        $applicant->university = $request['university_name'];
+        $applicant->save();
```
GauravGusain98, line 48: We usually do these kinds of operations in the service file. Can you create a function inside the CodeTrekService file and do this operation in it instead of here. 

Controllers are just use to call functions and return views.

Modules/CodeTrek/Resources/views/index.blade.php
```diff
@@ -42,7 +42,33 @@
...
                 </thead>
                 <tbody>
-                    <td>No applicant found</td>
+                    <tr>
+                        <td>
+                            @foreach ($applicants as $applicant )
+                                {{$applicant->first_name}}
+                                {{$applicant->last_name}}
+                                <br><br>
+                            @endforeach
```
GauravGusain98, line 51: Instead of adding foreach in each <td>, we can just use it outside <tr>, that way we don't have to duplicate the  foreach loop. Duplicacy is not a good practice in programming.

Modules/CodeTrek/Resources/views/index.blade.php
```diff
@@ -54,3 +80,46 @@
...
+                phone: '',
+                github_username: '',
+                start_date: '',
+                university: '',
+                course: '',
+                graduation_year: ''
+            }
+        },
+        methods: {
+            submitForm: async function(formId) {
```
GauravGusain98, line 101: The form is getting saved even when I have not filled any data in required fields
image


Check by inspecting the page and see errors in console.

Modules/CodeTrek/Resources/views/index.blade.php
```diff
@@ -54,3 +80,46 @@
     @endif
 </div>
 @endsection
+@section('vue_scripts')
```
GauravGusain98, line 83: I've a doubt that we already have some code defined to handle ajax form submission. let me check and get back to you on this. Till then please fix other feedbacks

GauravGusain98 (changes requested):
Please check the comments

GauravGusain98:
The toast is also not getting displayed when submitting the form

Modules/CodeTrek/Resources/views/index.blade.php
nishant22029: image
sir i'm trying to display toggle notifation but after submission data storing twice ,need help on this
image
image

nishant22029: image

Modules/CodeTrek/Resources/views/index.blade.php
```diff
@@ -42,7 +42,14 @@
                     </tr>
                 </thead>
                 <tbody>
-                    <td>No applicant found</td>
+                    @foreach ($applicants as $applicant )
+                        <tr>
+                            <td>{{$applicant->first_name}} {{$applicant->last_name}}</td>
+                            <td>-</td>
+                            <td>{{$applicant->status}}</td>
```
GauravGusain98, line 49: You need to create a status key in config.py file insider Codetrek module and for different status you need to set the label and slug. Then based on status value from database you need to show th label in the UI instead of slug
image

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -0,0 +1,32 @@
...
+        return ['applicants'=> $applicants];
+    }
+    public function store($data)
+    {
+        $applicant = new CodeTrekApplicant();
+
+        $applicant->first_name = $data['first_name'];
+        $applicant->last_name = $data['last_name'];
+        $applicant->email = $data['email_id'];
+        $applicant->phone = $data['phone'];
```
GauravGusain98, line 22: Use null coalescing in fields that are nullable
```suggestion
        $applicant->phone = $data['phone'] ?? null;
```

GauravGusain98 (changes requested):
Minor feedbacks, else looks good

1. The key name should not be status. Lets rename it to `tab`
    image

2. On submitting the applicant form the toast is not coming.

nishant22029:
image
toast message is coming on my system

Modules/CodeTrek/Config/config.php
```diff
@@ -1,5 +1,9 @@
 <?php
 
 return [
-    'name' => 'CodeTrek'
+    'name' => 'CodeTrek',
+    'status' => [
+        'active' => 'Active',
+        'inactive' => 'Inactive'
+    ]
```
GauravGusain98, line 8:
```suggestion
    'status' => [
        'active' => [
            'label' => 'Active',
            'slug' => 'active'
         ],
        'inactive' => [
            'label' => 'Inactive',
            'slug' => 'inactive'
         ]
    ]
```

GauravGusain98 (commented):
minor fix

GauravGusain98 (approved):
LGTM

Issue #2729, opened 2022-12-26
Balance in Invoice Payment


GauravGusain98:
**Feature Requested by**
@PrateekNarang 

**Is your feature request related to a problem? Please describe.**
There are some cases where the client doesn't pay the total amount mentioned in the invoice and promises to pay the remaining amount with the next invoice. 

For this type of case, we want to capture the balance amount corresponding to a client's invoice and should add this balance amount to the next invoice. This way we can have a track of the invoice balances. 

**Describe the solution you'd like**
- [ ] We need to create a new table `invoice_details` where we can store all the information for an invoice. _Note: There are some columns currently present in the invoice table that should ideally be a part of this new table. We will migrate them in a separate task later._
- [ ] Add an auto-calculated balance field in the payment received form. This balance field will be used to capture the balance amount. It will be calculated by subtracting the total received amount from the total invoice amount (including taxes)

**Describe alternatives you've considered**
NA

**Additional context**
NA

GauravGusain98:
### Action Items

**Pre-Requisite**
- [ ] Choose a client and its project which you are planning to use for testing
- [ ] Fill in details of that client and the project related to it
- [ ] Add details related to client like country, service rate, address, contact person
- [ ] Set billing frequency to monthly
- [ ] Edit project
    - [ ] Set billing level to Project level
    - [ ] Billing type to monthly billing
- [ ] Now run this query in database 
    ```sql
       INSERT INTO `settings` (`module`, `setting_key`, `setting_value`, `created_at`, `updated_at`) VALUES
    ('invoice', 'send_invoice_subject', '|*project_name*| Invoice |*term*| |*year*|', '2022-06-29 11:50:09', '2022-06-29 11:50:09'),
[...]

PR #2804, opened 2023-02-03
Clients invoice in installments

kanuj365:
Targets #2729

### Description
- In this feature we provide a functionality that Clients can pay partial payments or has paid the invoice in installments.

### Checklist:


GauravGusain98:
### Feedbacks
1. The partially paid option is not available 
image

2. I have not paid anything but the remaining amount is 0 by default
image

3. Also on adding the amount in the received amount field the remaining amount should change. This can be easily done  using vuejs
image

kanuj365:
image
if user paid in installments then every amount after installments amount can see.

kanuj365:
image

Modules/Invoice/Config/config.php
```diff
@@ -5,6 +5,7 @@
     'status' => [
         'sent' => 'Sent',
         'paid' => 'Paid',
+        'partially' => 'Partially Paid',
```
GauravGusain98, line 8:
```suggestion
        'partially_paid' => 'Partially Paid',
```

GauravGusain98 (changes requested):
### Testing feedbacks
1. Partially paid Status should also be shown here
    image

2. The status should come as `Partially Paid` in orange color
image


3. The amount related columns should be encrypted
    image

I think I'll update the requirement of this task a bit. So please give me some time to do that. Till then please fix the given feedbacks

GauravGusain98:
### Testing Feedbacks
- [ ] I am getting this error. Mostly looks like php issue. 
    image

- [ ] The status field is not coming up correct
    image

- [ ] The partially paid filter is not showing correct data
    image

- [ ] If an invoice is partially paid then we should not show overdue status, instead show partially paid status
    image

GauravGusain98:
### Additonal requirements
- [ ] We need to make some changes in the current table [schema](https://docs.google.com/presentation/d/1djNuUlhABvhZieWBax39jshAGEGsComSzvzMKtMQv6w/edit#slide=id.g1c14b30e04f_0_134) use this file to update the `invoice_details` table
- [ ] If any invoice payment details is capture then make an entry for that payment in this table
- [ ] For partially paid invoices. The amount paid till now should be calculated by adding the amount paid information of all the entries for an invoice in this table
- [ ] If there is any previous payment details available then show the details like shown in below image
image

GauravGusain98:
### Testing feedbacks
1. Overdue is still showing
2. Remaining amount is only decreasing. Its not resetting if we change the paid amount to 0
3. Rename the button text to Payment History
4. Make all the field default values as empty
5. Make details in accordion
    image
6. Store paid payment details also in new table

Issue #2816, opened 2023-02-13
Project TeckStack not showing


GauravGusain98:
**Issue Raised By** 
@GauravGusain98 

**How to Replicate**
1. Go to Projects
2. Edit a project 
3. Go to `Project Techstack` section 
    <img width="1203" alt="image" src="https://user-images.githubusercontent.com/54376313/218367508-c423c3f3-da0a-4574-8ba2-d01de196f56e.png">
4. Fill Teckstack details
5. Save the details
6. The updated data is not shown in the fields

**Expected behavior**
One we save the teckstack form, the updated data should be displayed in the page

GauravGusain98:
### Action Items
**Basics Concept - 1 pointer:**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function

**Form Submission**
- [ ] Read about HTML forms and how we submit forms. Also, share the use of the below provide items
- [ ] Action attribute in the form
- [ ] Method attribute in the form and different types of Methods
- [ ] Name attribute in any HTML tag and its importance while form submission
- [ ] Read about `old()` function in laravel

**Bugfix**
- [ ] Find if the data is getting stored in database 
- [ ] See why its not getting shown in the form
- [ ] Make chnages to display stored data in the form as well
- [ ] Create a PR
- [ ] Do self review of your code changes in GitHub
- [ ] Assign for review

rajneeshweb:
**Basics Concept - 1 pointer**
- [Previously Answered](https://github.com/ColoredCow/portal/issues/2781#issuecomment-1407372506)

- An HTML form is used to collect user input.
The HTML `<form>` element is used to create an HTML form.

- The action attribute defines the action to be performed when the form is submitted.
the form data is sent to a file on the server when clicking the submit button.

- The method attribute specifies the HTTP method to be used when submitting the form data.
The form-data can be sent as URL variables with method="get" or as HTTP post transaction with method="post". we never use GET method to send sensitive data.

[...]

PR #2827, opened 2023-02-17
fixed-Project-TeckStack-not-showing

rajneeshweb:
Targets #2816

### Description

Changes were made because Project TeckStack not showing the data.
### Checklist:

### DIFF
--- a/Modules/Project/Entities/ProjectMeta.php
+++ b/Modules/Project/Entities/ProjectMeta.php
@@ -13,18 +13,15 @@ class ProjectMeta extends Model
     {
         return $this->belongsTo(Project::class);
     }
-    
     public function getMetaValue($metaKey, $projectId)
     {
         $meta = $this->where('key', $metaKey)
         ->where('project_id', $projectId)
-        ->first();    
+        ->first(); 
 
         if ($meta) {
             return $meta->value;
         }
-
         return null;
     }
-
 }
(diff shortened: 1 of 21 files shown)


Modules/Project/Resources/views/subviews/edit-project-techstack-details.blade.php
```diff
@@ -7,30 +7,30 @@
             <div class="form-group row m-5">
                 <label for="language" class="col-sm-4 col-form-label m-4">Language</label>
                 <div class="col-sm-6 m-4">
-                    <input type="text" name="language" class="form-control" id="language" placeholder="Language" value="{{old('language')}}">
+                    <input type="text" name="language" class="form-control" id="language" placeholder="Language" value="{{ old('language', $projectmeta->where('key', 'language')->first()->value ?? '') }}">
```
GauravGusain98, line 10: Instead of using where condition every time can you create a function in project Meta model with name `getMetaValute(MetaKey)`. This will return the value of the meta key for a project. In case the key does not exist, it will simply return null.

GauravGusain98 (changes requested):

Modules/Project/Entities/ProjectMeta.php
```diff
@@ -13,4 +13,14 @@ public function project()
...
+    public function getMetaValue($metaKey, $projectId)
+    {
+        $meta = $this->where('key', $metaKey)
+                     ->where('project_id', $projectId)
+                     ->first();
+
+        if ($meta) {
+            return $meta->value;
+        }
+    }
```
GauravGusain98, line 25: I think this function should be inside Project Model, then you don't have to pass projectId.

GauravGusain98 (changes requested):
Please check comment

GauravGusain98:
Take latest pull from master and push it again

GauravGusain98 (approved):
LGTM

Issue #2825, opened 2023-02-16
Change the Effort Summary time


GauravGusain98:
### Enhancement suggested by
@GauravGusain98 

### Problem that your solution is solving
In ColoredCow we use effort sheet where we track the hour we spent on a particular task. This helps us to calculate our productivity and is also used for billing the client. 

We periodically migrate/update the data present in the effort sheet to the portal. For this, we use a Laravel command that refreshes the booked hours of team members from the effort sheet.
We also send the details of projects as show below at 10:30 PM on Google chat channel. So that the team member should be aware of their team progress.
![Image](https://user-images.githubusercontent.com/54376313/216901654-21b4380a-1b46-4595-8589-53be8e21ae45.png) 

There are many employees and interns who used to update their efforts late at night. Therefore we want to send the summary on next day at 8 AM. 

### Contact person for help
@GauravGusain98

GauravGusain98:
### Action items

**Basics Concept - 1 pointer:**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function

**Task Scheduling**
- [ ] Read about [task scheduling](https://laravel.com/docs/8.x/scheduling) in Laravel
- [ ] Find the command that is use to send the summary at 10:30 PM.
- [ ] Change the timing to 8 AM 
- [ ] Go through the command logic and see if the logic will work if the billing period of project changes because when a billing period changes then team member velocity is reset to 0

GauravGusain98:
Not needed now

PR #2831, opened 2023-02-21
change effort summary time

DiskshaChauhan:
Targets #2825

Description
Change effort summary time 

### Checklist:

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -52,7 +52,7 @@ class Kernel extends ConsoleKernel
         $schedule->command('project:zero-effort-in-project')->weekly()->mondays()->at('09:00');
         $schedule->command('project:ended-project')->dailyAt('09:00');
         $schedule->command('project:remind-to-update-effort')->weekdays()->at('19:00');
-        $schedule->command('project:send-daily-effort-summary-google-chat')->weekdays()->at('22:30');
+        $schedule->command('project:send-daily-effort-summary-google-chat')->weekdays()->at('08:00');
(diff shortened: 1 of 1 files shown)


app/Console/Kernel.php
```diff
@@ -52,7 +52,7 @@ protected function schedule(Schedule $schedule)
         $schedule->command('project:zero-effort-in-project')->weekly()->mondays()->at('09:00');
         $schedule->command('project:ended-project')->dailyAt('09:00');
         $schedule->command('project:remind-to-update-effort')->weekdays()->at('19:00');
-        $schedule->command('project:send-daily-effort-summary-google-chat')->weekdays()->at('22:30');
+        $schedule->command('project:send-daily-effort-summary-google-chat')->weekdays()->at('08:00');
```
GauravGusain98, line 55: Did you check the command logic? As now we are sending alert on next day morning instead of current day, so it might affect the logic that is used to calculate the effort summary of team members

GauravGusain98:
Can you take pull from master and push the code again

app/Console/Kernel.php
DiskshaChauhan: Sir can you explain me it as i'm not able to find the logic for calculating the effort summary of team member.

GauravGusain98: Check where we write command related files and find the file corresponding to this command
`send-daily-effort-summary-google-chat`

GauravGusain98: @DiskshaChauhan Can you check the above comment

PR #2832, opened 2023-02-21
CI fixes

rathorevaibhav:
### Description
1. Added legacy OpenSSL provider option before running npm command as the latest versions of node17 have switched to using OpenSSL3.0
2. The export option is based on this [StackOverflow answer](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported)
3. Removed npm installation from Larastan and PHPCS as these are purely PHP level checks and we don't need node modules. So we can save time when running these jobs.

### Checklist:


GauravGusain98 (approved):
LGTM

Issue #2737, opened 2022-12-27
Store invoice term in database


GauravGusain98:
**Enhancement Suggested By**
@GauravGusain98 

**Is your feature request related to a problem? Please describe.**
To manage finance we should have all the information of the invoice captured with us. This will help us to fetch all the necessary information regarding an invoice whenever needed. 

One of the important such information is terms of the invoice. The term of the invoice defines the period for which the invoice is raised. 

We already have the columns `term_start_date` and `term_end_date` present in the database where we want to store the start and end date of the invoice term. In this task we need to write the logic to populate those columns in the database.

**Additional context**
Once we are able to store the term of the invoice then we don't have to calculate the term every time and we can directly use this information while creating invoices and sending mails corresponding to invoices.

**Related issues**
https://github.com/ColoredCow/portal/issues/2732

**Point of Contact for guidance**
@GauravGusain98

GauravGusain98:
### Action Items

- [ ] Find the logic in the codebase where invoice entry is created. 
- [ ] Figure out the logic to calculate the term start date and end date.
- [ ] Discuss the term calculation logic with @GauravGusain98 
- [ ] Once the logic is reviewed and approved, add code to store term data in the `term_start_date` and `term_end_date` columns of invoices table

Hemrajbharti123:
### **term start date and term end date is added **
![image](https://user-images.githubusercontent.com/121417679/213139655-7a3b494b-78a9-4379-bfe0-8791f0f41068.png)

Hemrajbharti123:
### Form data is printing on screen using dd() function

![image](https://user-images.githubusercontent.com/121417679/213338379-ef567664-5744-490d-9bda-42e770669a03.png)

Hemrajbharti123:
Invoice data along with start date and end date 

![image](https://user-images.githubusercontent.com/121417679/213384315-32734523-37b0-4915-8dae-9a7d5747c406.png)

Hemrajbharti123:
### Dummy function which returns true.

![image](https://user-images.githubusercontent.com/121417679/213624407-12e42fd5-f733-44dd-b5ad-ea53bcff551d.png)

Hemrajbharti123:
### getTermStartAndEndDate function

![image](https://user-images.githubusercontent.com/121417679/213637451-dfe43650-3672-4cf3-994e-ca7ecca22666.png)

Hemrajbharti123:
### Ready to send invoice

[...]

PR #2843, opened 2023-02-23
Store invoice term in database

Hemrajbharti123:
Targets #2737 

### Description
Store Term Start Date and Term End Date in database 

### Checklist:

### DIFF
--- a/Modules/Invoice/Resources/views/subviews/create/invoice-details.blade.php
+++ b/Modules/Invoice/Resources/views/subviews/create/invoice-details.blade.php
@@ -1,8 +1,8 @@
 <div id="create_invoice_details_form">
     <div class="card-body">
         <div class="form-row mb-4">
-            <div class="col-md-5">
-                <div class="form-group position-relative">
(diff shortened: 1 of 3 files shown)


GauravGusain98:
Can you move the comment section at bottom
image

GauravGusain98 (changes requested):
minor UI suggestion please check

Modules/Invoice/Resources/views/subviews/create/invoice-details.blade.php
```diff
@@ -2,7 +2,7 @@
     <div class="card-body">
         <div class="form-row mb-4">
             <div class="col-md-5">
-                <div class="form-group">
+                <div class="form-group position-relative">
```
GauravGusain98, line 5: This is not correct way to do it as it will affect the placement of any other fields that we add in the future to this page in this section.

Please use the flex approach

GauravGusain98 (changes requested):
Please check the comment

GauravGusain98 (approved):
LGTM

Issue #2783, opened 2023-01-21
Office location and capacity - phase 1 


GauravGusain98:
**Enhancement suggested by**
@P4NK4J

**Parent Issue**
https://github.com/ColoredCow/portal/issues/2382

**Is your feature request related to a problem? Please describe.**
In ColoredCow we have a culture to freely move to different locations and work from there. As ColoredCow is growing and now we have multiple offices at different locations so we need to manage the capacity of an office so that is not overpopulated at a particular time. 

To manage these kinds of things, we need an operation Module where we will create functionalities related to operations. 

**Additional context**
On this module is set up in the portal, then we can start managing our operations-related tasks from the Portal, which will help us to easily manage our infrastructure, capacity, etc

**Point of Contact for guidance**
@GauravGusain98

GauravGusain98:
### Action Items
**Basics - 1 pointer**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function

**Laravel Modules - 1 pointer**
- [ ] Read about the [Laravel module](https://nwidart.com/laravel-modules/v6/basic-usage/creating-a-module) and what is benefit of it. Share your understanding in your own language in the task comment. 
- [ ] Create a new module in Portal with the name Operations

**Add Home page - 1 Pointer**
- [ ] Add Operations option in Menu bar as shown in below image
    <img width="1416" alt="image" src="https://user-images.githubusercontent.com/54376313/213862119-39d63b52-77f5-4267-9858-ca227408f228.png">
- [ ]  When Clicking on the `Office Location` link, then a page should open, in which the header and its center should have `Coming Soon` written.
    <img width="876" alt="image" src="https://user-images.githubusercontent.com/54376313/213862288-f098e18a-c209-46a4-b086-44be641a8e2d.png">

Harshvardhan80:
Laravel modules
-A Laravel module is a larvael package
-It has controller, model ,views or routes
-It  helps to make projects

create module in larvel
[...]

PR #2847, opened 2023-02-24
added operations module

Harshvardhan80:
Targets #2783 

### Checklist:


Modules/Operations/Resources/views/layouts/master.blade.php
```diff
@@ -0,0 +1,19 @@
...
+       {{-- <link rel="stylesheet" href="{{ mix('css/operations.css') }}"> --}}
+
+    </head>
+    <body>
+        @yield('content')
+
+        {{-- Laravel Mix - JS File --}}
+        {{-- <script src="{{ mix('js/operations.js') }}"></script> --}}
+    </body>
+</html>
```
GauravGusain98, line 19: This is not correct way to create a page. please check logic in other modules

GauravGusain98 (changes requested):
Along with the feedbacks, also Remove package-lock.json file changes.

Modules/Operations/Resources/views/layouts/master.blade.php
Harshvardhan80: ok sir

resources/views/index.blade.php
```diff
@@ -0,0 +1,4 @@
+@extends('layouts.app')
```
GauravGusain98, line 1: The blade file corresponding to Operations should be inside operation module

GauravGusain98: This file is not needed

resources/views/layouts/navbar.blade.php
```diff
@@ -113,14 +113,28 @@
             </div>
         </li>
     @endcan
+    @can('operations.view')
+        <li class="nav-item dropdown">
+            <a id="navbarDropdown_operations" class="nav-link dropdown-toggle" href="#" role="button"
+                data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" v-pre>Operations <span class="caret"></span>
+            </a>
+            <div class="dropdown-menu z-index-1100" aria-labelledby="navbarDropdown_operations">
+                <a class="dropdown-item" href="{{ url('/operations') }}">Office location</a>
```
GauravGusain98, line 122: We use route for URLs. PLease use route function

GauravGusain98 (changes requested):

Issue #2844, opened 2023-02-23
Export Profit and Loss data in Excel sheet


pankaj-ag:
If you go to Finance -> Reports-> Profit and Loss -> detailed P&L report

We have a report where we show the revenue with different filters. We need to download this report in an excel format with current filters applied. 

<img width="1400" alt="image" src="https://user-images.githubusercontent.com/10670690/220923457-091df710-cd2f-4e3d-be64-81ebdbe9519b.png">

PR #2850, opened 2023-02-27
Export Profit and Loss Report in Excel

yuvraj752:
Targets #2844

### Description

These changes are required to download profit and loss report data in excel.

### Checklist:

### DIFF
--- a/Modules/Report/Services/Finance/ProfitAndLossReportService.php
+++ b/Modules/Report/Services/Finance/ProfitAndLossReportService.php
@@ -36,7 +36,7 @@ class ProfitAndLossReportService
         $request = request()->all();
         $endYear = $request["year"];
         $startYear = $endYear - 1;
-        
+
         return Excel::download(new ProfitAndLossReportExport($reportData), "Profit And Loss Report $startYear-$endYear.xlsx");
     }
 
--- a/Modules/Report/Services/Finance/ProfitAndLossReportService.php
+++ b/Modules/Report/Services/Finance/ProfitAndLossReportService.php
@@ -34,7 +34,7 @@ class ProfitAndLossReportService
         $reportData = $this->profitAndLoss($filters);
         $reportData = $this->formatProfitAndLossForExportAll($reportData);
         $request = request()->all();
-        $endYear = $request["year"];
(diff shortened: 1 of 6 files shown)


pankaj-ag:
@yuvraj752 we need to use the same filters we are applying on the profit and loss data report. Currently I think it just downloads data for current Financial year.

pankaj-ag (changes requested):
@yuvraj752 we need to use the same filters we are applying on the profit and loss data report. Currently I think it just downloads data for current Financial year.

Modules/Report/Services/Finance/ProfitAndLossReportService.php
```diff
@@ -18,4 +21,73 @@ public function profitAndLoss(array $filters): array
...
+
+    public function profitAndLossReportExport()
+    {
+        $currentYear = date('m') > 03 ? date('Y') + 1 : date('Y');
+        $defaultFilters = [
+            'transaction' => 'revenue',
+            'year' => $currentYear,
+        ];
+
+        $filters = array_merge($defaultFilters, request()->all());
```
pankaj-ag, line 33: Let's keep the filter generation part in a single place or file. It will be easy to maintain that.

Modules/Report/Services/Finance/ProfitAndLossReportService.php
```diff
@@ -18,4 +21,73 @@ public function profitAndLoss(array $filters): array
...
+        $currentYear = date('m') > 03 ? date('Y') + 1 : date('Y');
+        $startYear = request()->input('year', $currentYear);
+        $lastYear = $startYear - 1;
+        $startYearVal = substr($startYear, -2);
+        $lastYearVal = substr((string) $lastYear, -2);
+        $allAmounts = array_map(function ($item) {
+            return $item['amounts'];
+        }, $reportData);
+
+        $profitAndLossData = [];
```
pankaj-ag, line 54: Since this function is only related to exports let's move this to an export file only.

pankaj-ag (changes requested):

pankaj-ag (approved):

Issue #2826, opened 2023-02-16
Show All projects by default to admin and super admin


GauravGusain98:
### Enhancement suggested by
@mohitsharma-22 

### Problem that your solution is solving
In project listing page by default the user project is shown. For admin and super admin user who usually are not in any project we want to show the all projects list by default. 

**Expected behavior**
- [x] #2839
- [x] For admin and super admin role users, the all project listing should be shown as default

### Contact person for help
@GauravGusain98

GauravGusain98:
### Action items

**Basics Concept - 1 pointer:**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function

**Roles and Permission**
- [ ] Read about roles and permission - [Link](https://spatie.be/docs/laravel-permission/v5/basic-usage/basic-usage)

**Default project listing**
- [ ] Fix the logic such that for admin and super admin the default page should be `all project` instead of `my project`
    <img width="1376" alt="image" src="https://user-images.githubusercontent.com/54376313/219437212-1df204bc-e7bc-4726-9f9c-be0590ed633e.png">

thisajaychauhan:
**views** - views is a graphical user interface by which users can interact.
all the things we can see on the display.

**models** - models are connected to controllers by which controllers interact with a database.

**controller**- controllers is a bridge between views and models. 
the controller contains the logical part. 
it is the heart of the MVC pattern.
receive updates from views then notify the model and apply crud methods. 
vice versa take updates from the database and display them in views.

**Services** - services also contain logical parts.
it is responsible for complex logic. like validation logic.

**routing** - routing is used in making a path to the particular page requested by the user.

thisajaychauhan:
we can print a value using dd() in .blade.php file by 
@dd("**enter the text...** ");

thisajaychauhan:
print **hello world** using the dd function

thisajaychauhan:
[...]

PR #2861, opened 2023-03-02
feature/#2826/Show_All_projects_by_default_to_admin_and_super_admin

thisajaychauhan:
Targets #2826

### Description
In project listing page by default the user project is shown. For admin and super admin user who usually are not in any project  show all projects list by default.

### Checklist:

### DIFF
--- a/Modules/Project/Resources/views/index.blade.php
+++ b/Modules/Project/Resources/views/index.blade.php
@@ -32,20 +32,22 @@
                     <input type="hidden" name="is_amc" value="{{ request()->input('is_amc', 0) }}">
                     <select class="fz-14 fz-lg-16 p-1 bg-info ml-3 my-auto text-white rounded border-0" name="projects"
                         onchange="this.form.submit()">
-                        @if(auth()->user()->isAdmin() ||
-                                auth()->user()->isSuperAdmin())
+                        @php
+                            $isUserAdmin =auth()->user()->isAdmin() ||
+                                auth()->user()->isSuperAdmin();
+                        @endphp
+                        <option value="my-projects"
+                            {{ request()->get('projects') == 'my-projects' || (!$isUserAdmin && !request()->has('projects')) ? 'selected' : '' }}>
+                            {{ __('My Projects') }} </option>
+
+                        @if ($isUserAdmin)
                             <option value="all-projects"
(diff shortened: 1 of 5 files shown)


Modules/Project/Services/ProjectService.php
```diff
@@ -29,9 +29,11 @@ public function index(array $data = [])
         if ($nameFilter = $data['name'] ?? false) {
             $filters['name'] = $nameFilter;
         }
-
-        $showAllProjects = Arr::get($data, 'projects', 'my-projects') != 'my-projects';
-
+        if (auth()->user()->roles[0]['id'] == 1 || auth()->user()->roles[0]['id'] == 2) {
```
GauravGusain98, line 32: Id will never be same in different system. It can vary, so we don't have to use this approach to get the role. I think you can use the function that you created in User Model here

Modules/User/Entities/User.php
```diff
@@ -61,6 +61,11 @@ public function isSuperAdmin()
         return $this->hasRole('super-admin');
     }
 
+    public function isAdmin()
```
GauravGusain98, line 64: Why you have added this function in both User.php? Do you know which User.php is used for user table

GauravGusain98: Keep the function only in correct model

Modules/Project/Resources/views/index.blade.php
```diff
@@ -32,11 +32,21 @@
...
+                            <option value="my-projects"
+                                {{ request()->get('projects') == 'my-projects'  ? 'selected' : '' }}>
+                                {{ __('My Projects') }} </option>
+                        @else
+                            <option value="my-projects"
+                                {{ request()->get('projects') == 'my-projects' || !request()->has('projects') ? 'selected' : '' }}>
+                                {{ __('My Projects') }} </option>
+                            <option value="all-projects" {{ request()->get('projects') == 'all-projects' ? 'selected' : '' }}>
+                                {{ __('All Projects') }} </option>  
+                        @endif
```
GauravGusain98, line 49: So much of code duplicacy. Can't you add a check only in the area where there is difference. 

We should not write same code multiple times

GauravGusain98 (changes requested):
Please check the comment

GauravGusain98 (approved):
LGTM

PR #2864, opened 2023-03-05
The My Profile button should open the user dashboard page where we se…

shreyakharola:
…e the FTE of the user

Targets #{ISSUE_ID}

### Checklist:

### DIFF
--- a/resources/views/layouts/app.blade.php
+++ b/resources/views/layouts/app.blade.php
@@ -61,7 +61,7 @@
 
                                 <div class="dropdown-menu" aria-labelledby="navbarDropdown">
                                     @if (Module::checkStatus('User'))
-                                        <a class="dropdown-item" href="{{ route('user.profile') }}">My profile</a>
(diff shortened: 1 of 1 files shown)


resources/views/layouts/app.blade.php
```diff
@@ -61,7 +61,7 @@
 
                                 <div class="dropdown-menu" aria-labelledby="navbarDropdown">
                                     @if (Module::checkStatus('User'))
-                                        <a class="dropdown-item" href="{{ route('user.profile') }}">My profile</a>
+                                        <a class="dropdown-item" href="{{ route('home') }}">My profile</a>
```
GauravGusain98, line 64: This is not the correct route to which we need to redirect the user

GauravGusain98 (changes requested):
Please check the comment

resources/views/layouts/app.blade.php
shreyakharola: sir, I am redirecting **my profile** button to the employees page. i.e. http://localhost/my profile/employees

PR #2869, opened 2023-03-11
added operations modules

Harshvardhan80:
Targets #2783 

Description
Checklist:


Modules/Operations/Resources/views/index.blade.php
```diff
@@ -0,0 +1,5 @@
+@extends('operations::layouts.master')
```
GauravGusain98, line 1: You should move this file inside Module/Operations/Resources/views/office-location and create index file inside that. Then in controller as well you should call that file.

Modules/Operations/Routes/web.php
```diff
@@ -0,0 +1,16 @@
...
+|--------------------------------------------------------------------------
+|
+| Here is where you can register web routes for your application. These
+| routes are loaded by the RouteServiceProvider within a group which
+| contains the "web" middleware group. Now create something great!
+|
+*/
+
+Route::prefix('operations')->group(function () {
+    Route::get('/', 'operationsController@index')->name('office_location.index');
```
GauravGusain98, line 15: the url should be `operation/office-location`

resources/views/layouts/navbar.blade.php
```diff
@@ -113,6 +113,17 @@
             </div>
         </li>
     @endcan
+    @can('operations.view')
```
GauravGusain98, line 116: Where is this permission created. You need to create this permission and assign this permission to `admin` and `super-admin`

GauravGusain98: Check how we have created other permissions and assign them to different role.

Harshvardhan80: i have done

Modules/Operations/Resources/views/index.blade.php
Harshvardhan80: i have done

Modules/Operations/Routes/web.php
```diff
@@ -0,0 +1,16 @@
...
+|--------------------------------------------------------------------------
+|
+| Here is where you can register web routes for your application. These
+| routes are loaded by the RouteServiceProvider within a group which
+| contains the "web" middleware group. Now create something great!
+|
+*/
+
+Route::prefix('operations')->group(function () {
+    Route::get('/operations/office_location', 'operationsController@index')->name('office-location.index');
```
GauravGusain98, line 15: As per your implementation, the URL for this route will become `operations/operations/office_location` but it should be `operations/office_location`

composer.lock
GauravGusain98, line 1: We don't need any changes in this file. So please revert the changes done to this file

package-lock.json
GauravGusain98, line 1: We don't need any changes in this file. So please revert the changes done to this file

Modules/Operations/Database/Seeders/OperationsDatabaseSeeder.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
+        Model::unguard();
+
```
GauravGusain98, line 18: I think here you need to call the `OperationsPermissionTableSeeder`. Check other modules like Project, HR etc.

GauravGusain98 (changes requested):
minor suggestions please check

Issue #2874, opened 2023-03-14
Correct project name is removed from the list  


Hemrajbharti123:
**Describe the bug**
There is an issue in `edit client information` ,whenever in projects where the first project name in the sequence is always being removed even if we are trying to remove a project name other than the first one.

**To Reproduce**
1. Go to 'edit client information'
2. Click on 'projects'
3. Remove any client

**Screenshots**
![image](https://user-images.githubusercontent.com/121417679/224969233-a7aaae85-c897-4992-8cb5-0dd56fe45775.png)

**Action Items**

- [x] Read about how data is deleted in Laravel.
- [x] Find how Project data is flow from remove button to the controller.
- [x] Use dd for debug.

PR #2879, opened 2023-03-16
bugfix/#2874/Correct-project-name-is-removed-from-the-list

Hemrajbharti123:
Targets #2874 

### Description
There is an issue in edit client information ,whenever in projects where the first project name in the sequence is always being removed even if we are trying to remove a project name other than the first one. This issue is related to the project id the correct id is not passes from the remove button. 

### Checklist:


GauravGusain98 (approved):
LGTM

Issue #2877, opened 2023-03-15
Show correct velocity of team members  according to their added on data in a project


Gauravchamoli1997:
**Describe the task**
Currently, the velocity of a team members is calculated from the start date of the project billing cycle . This doesn't depends on whether the team members are added in middle of the month or at the end of the month. Due to this the velocity of new members does not come correct in the first month.

Now we have to create a logic in which the velocity of a team members  to be calculated from the added on date instead of the start date of the month.

**Expected behavior**
- [x] The velocity of a team members to be calculated from the added on date instead of the start date of the billing cycle.

**Action Items**

- [x] Find the file in which velocity of team members is shown.
- [x] Go to that file and Check how we can get the velocity of a team member.
- [x] Find the logic from which the velocity of a team member is being calculated.

PR #2883, opened 2023-03-16
Show correct velocity of team members according to their added on data in a project

Gauravchamoli1997:
Targets #2877
### Description
Previously, the velocity of a team members is calculated from the start date of the project billing cycle . This doesn't depends on whether the team members are added in middle of the month or at the end of the month. Due to this the velocity of new members does not come correct in the first month. But now the velocity of  team members is calculated from the started_on date in the project for first month only  if team member team member is added in the middle of the month.

### Checklist:

### DIFF
--- a/Modules/Project/Entities/ProjectTeamMember.php
+++ b/Modules/Project/Entities/ProjectTeamMember.php
@@ -56,15 +56,16 @@ class ProjectTeamMember extends Model
     {
         $project = new Project;
         $currentDate = today(config('constants.timezone.indian'));
-        $startDate = $startDate ?? $this->project->client->month_start_date;
+        if ($this->started_on && $this->started_on > $startDate) {
+            $startDate = $this->started_on;
+        } else {
+            $startDate = $this->project->client->month_start_date;
+        }
         if (now(config('constants.timezone.indian'))->format('H:i:s') < config('efforttracking.update_date_count_after_time')) {
             $currentDate = $currentDate->subDay();
         }
 
-        $daysTillToday = count($project->getWorkingDaysList($this->project->client->month_start_date, $currentDate));
-        if ($this->started_on && $this->started_on > $startDate) {
-            $daysTillToday = count($project->getWorkingDaysList($this->started_on, $currentDate));
-        }
+        $daysTillToday = count($project->getWorkingDaysList($startDate, $currentDate));
 
         return $this->daily_expected_effort * $daysTillToday;
     }
--- a/Modules/Project/Entities/ProjectTeamMember.php
+++ b/Modules/Project/Entities/ProjectTeamMember.php
@@ -56,7 +56,7 @@ class ProjectTeamMember extends Model
     {
         $project = new Project;
         $currentDate = today(config('constants.timezone.indian'));
-        if ($this->started_on && $this->started_on > $startDate) {
+        if ($this->started_on && $this->started_on > $this->project->client->month_start_date) {
             $startDate = $this->started_on;
         } else {
             $startDate = $this->project->client->month_start_date;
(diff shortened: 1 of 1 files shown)


Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -47,22 +47,24 @@ public function scopeActive($query)
 
     public function getCurrentActualEffortAttribute($startDate = null)
     {
-        $startDate = $startDate ?? $this->project->client->month_start_date;
+        $startDate = $startDate ?? $this->started_on;
 
-        return $this->projectTeamMemberEffort()->where('added_on', '>=', $startDate)->sum('actual_effort');
+        return $this->projectTeamMemberEffort()->where('added_on', '>=', $this->started_on)->where('added_on', '>=', $startDate)->sum('actual_effort');
```
GauravGusain98, line 52: I think this was correct earlier. We don't need to make any change to this function because if we keep the start date before the team member was added to project, it will not affect the actual effort.

Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -47,22 +47,24 @@ public function scopeActive($query)
...
         $startDate = $startDate ?? $this->project->client->month_start_date;
-
         if (now(config('constants.timezone.indian'))->format('H:i:s') < config('efforttracking.update_date_count_after_time')) {
             $currentDate = $currentDate->subDay();
         }
 
         $daysTillToday = count($project->getWorkingDaysList($this->project->client->month_start_date, $currentDate));
+        if ($this->started_on && $this->started_on > $startDate) {
+            $daysTillToday = count($project->getWorkingDaysList($this->started_on, $currentDate));
+        }
```
GauravGusain98, line 67: Why you are calculating the days till today 2 times. It can be made more optimized. Just calculate start date first based on different cases and then calculate the days till today

GauravGusain98 (changes requested):
Please check the comment

Modules/Project/Entities/ProjectTeamMember.php
Gauravchamoli1997: Ok i will work on it.

Gauravchamoli1997:
This is my updated code . 
image

Modules/Project/Entities/ProjectTeamMember.php
```diff
@@ -56,13 +56,16 @@ public function getCurrentExpectedEffortAttribute($startDate = null)
     {
         $project = new Project;
         $currentDate = today(config('constants.timezone.indian'));
-        $startDate = $startDate ?? $this->project->client->month_start_date;
-
+        if ($this->started_on && $this->started_on > $startDate) {
+            $startDate = $this->started_on;
```
GauravGusain98, line 60: The start date will aways be team member added on date as started_on date will always be greater than null because the initial value of startDate is null. 

Due to this logic the expected efforts are coming very high  
image

GauravGusain98 (changes requested):
Please check the comment

Modules/Project/Entities/ProjectTeamMember.php
Gauravchamoli1997: @GauravGusain98 I think the expected efforts are coming very high due to this ```$this->started_on > $startDate``` . I think I have to take this ```$this->started_on > $this->project->client->month_start_date``` in the place of ```$this->started_on > $startDate```

GauravGusain98: Ok the try and check if it is working or not

Gauravchamoli1997: @GauravGusain98 It is working correctly in my system.
image

GauravGusain98 (approved):
LGTM

Issue #2792, opened 2023-01-25
Show client id in client page


GauravGusain98:
**Enhancement Requested By**
@mohitsharma-22 

**Problem Statement**
Every client that we create in the portal has an id corresponding to it. This id is not the `id` column of the clients table in the database instead it is the sequence number of the client. Whenever a new client comes we provide it a client_id so that we should know which number of clients it is for us.

We use this client_id in the invoice number as well. So we want to know the client id corresponding to the client in portal. 

**Point of contact for guidance**
@GauravGusain98

GauravGusain98:
### Action items

**Basics Concept - 1 pointer**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function
- [ ] Try printing the detail of any client on the client page using the dd() function

**Add client id - 0.5 pointer**
- [ ] On the client page add a new column `client id` before the client name. The value of this column should be taken from the `client_id` column 

**Sorting - 2 pointer**
- [ ] Add sorting icons in front of the name and id column similar to what we have on the Recruitment page
    ![image](https://user-images.githubusercontent.com/54376313/214581475-0cbefbd5-4674-45df-9e8a-52c35ff6719d.png)
- [ ] On hovering over these icons the cursor should work to change to a finger shape. _Note: For styling, we use HTML classes instead of adding CSS styles. We already have a defined class for the cursor with a pointer shape, just find it and add it to these icons_
- [ ] On clicking these icons the list clients should be sorted according to those columns. 
    - [ ] Find how the clients are sorted right now
    - [ ] Figure out the approach of how you are going to implement the solution so that the client should be ordered as per the selection. Write down the approach in the comment and share it with @GauravGusain98 
    - [ ] After approval, start working on solution

Priyanka-Kaintura:
sir, Can I try proceeding forward using Ajax if thi is the criteria for sorting.

[...]

PR #2890, opened 2023-03-20
Showed client-id in client page.

Priyanka-Kaintura:
Targets #2792

### Description
Showed the client-id in the client page . 

### Checklist:


Modules/Client/Resources/views/index.blade.php
```diff
@@ -59,6 +60,6 @@
             </tbody>
         </table>
 
-    </div>
```
GauravGusain98, line 62: why removed this div? Is it an extra div?

GauravGusain98 (commented):
Left a comment please check.

GauravGusain98 (approved):
LGTM

Issue #2842, opened 2023-02-23
Automated notification for invoice payment


Gauravchamoli1997:
**Describe the task**
Currently, we don't have any automated notification system feature which tells us that the invoice for a particular project is being paid.

**Expected behavior**
- [x] We need a automated notification system telling us that an invoice for  particular project is being paid through project channel.
- [x] The content for that notification can be: 
1. For Project :- `@all We have received the payment for {project-name} successfully!` 
2. For Client :- `@all We have received the payment for {client-project-name} successfully!`

**Point on Contact for Guidance
@GauravGusain98**

**Action Items**
- [ ] Find the file in which Invoice Information data is stored and understand the work flow. 
- [ ] Add a Checkbox **Send Payment Received Notification** like shown in image below.

![Screenshot (69)](https://user-images.githubusercontent.com/121275531/221537380-80f6c8d7-9335-43b3-973d-7dad7d5bde5c.png)

- [ ] Now Go to Invoice Module create a folder named ```Notifications\GoogleChat``` after that create a file named ```SendPaymentReceivedNotification.php``` inside ```Modules\Invoice\Notifications\GoogleChat``` folder.
- [ ] Once the file is created . Understand the logic and Create a functionality for GoogleChat. **Hint**:- We have similar functionality for ```RemindToUpdateEffort``` inside ```project module```.
- [ ] Create a logic and functionality how we can send notification through GoogleChat.

GauravGusain98:
The requirement looks fine @Gauravchamoli1997. Can you also add the breakdown of action items that you will be performing

Gauravchamoli1997:
> The requirement looks fine @Gauravchamoli1997. Can you also add the breakdown of action items that you will be performing

@GauravGusain98 Thanks. As I do this task, I will keep adding action items to the breakdown that I will perform.

GauravGusain98:
### Testing Feedbacks

This is what we received on channel when we tested this feature. The invoice details is being displayed
<img width="808" alt="image" src="https://user-images.githubusercontent.com/54376313/227014412-5318a375-2c46-4156-9bea-421e9308caf9.png">

PR #2919, opened 2023-03-30
Automated notification for invoice

Gauravchamoli1997:
Targets #2842

### Description
Add a check here for webhookURL because if a project doesn't have a webhook URL then the funtionality is throwing error. So we only want to send notification if a project has a webhook URL.

### Checklist:


GauravGusain98 (approved):
LGTM

GauravGusain98 (approved):
LGTM

Issue #2915, opened 2023-03-29
Responsive Book page 


mohitgusain:
**Describe the bug**
In mobile, The book page should be shown fine with proper font sizes and weights. So we need to make it responsive. 

**Screenshots**
<img width="261" alt="image" src="https://user-images.githubusercontent.com/54376313/229814661-3be6cca8-3406-43ab-9f8d-e1a8edeb7c19.png">

**Expected behaviour**
The book page should be shown properly in screen sizes
- [ ] 360px
- [ ] 768px
- [ ] 1200px
- [ ] 1440px

**Additional details**
- [ ] Read how variable.scss file works
- [x] #2982

PR #2926, opened 2023-04-01
Add new book button overlapping fixed

akshatsemwal146:
Targets #2915 
### Description
In mobile, on the book module theadd new button is overlapping so, i fixed it.
### Checklist:

### DIFF
--- a/resources/views/knowledgecafe/library/books/index.blade.php
+++ b/resources/views/knowledgecafe/library/books/index.blade.php
@@ -14,11 +14,11 @@
     <br>
     <div class="row">
         <div class="col-4">
-            <h3>Books({{count($books)}})</h3    >
+            <h3>Books({{count($books)}})</h3>
         </div>
         @can('library_books.create')
         <div class="col-8 xs-6">
-            <a href="{{ route('books.create') }}" class="btn btn-success float-right xs-2"><i class="fa fa-plus mr-1"></i> Add New Book</a>
+            <a href="{{ route('books.create') }}" class="btn btn-success float-right col-xs-2"><i class="fa fa-plus mr-1"></i> Add New Book</a>
         </div>
         @endcan
     </div>
--- a/resources/views/knowledgecafe/library/books/index.blade.php
+++ b/resources/views/knowledgecafe/library/books/index.blade.php
@@ -17,7 +17,7 @@
             <h3>Books({{count($books)}})</h3>
         </div>
         @can('library_books.create')
-        <div class="col-8 xs-6">
+        <div class="col-8 col-xs-6">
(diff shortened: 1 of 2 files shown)


GauravGusain98:
I think this is not looking correct.
1. The font size of `Books(count)` can be reduced so that the button text should come in one line
2. The margin below the `Books(count)` can be reduced 
4. Also the button text font size can be reduced so that the button size does not become too large
5. The margin bottom between search and sort fields can be reduced in mobile screens. Currently they are too far away
6. The space on the right side of books card is greater then left side. So the card can be placed in between for better UI
image
image

Hemrajbharti123:
### Local Testing Feedback
The button is still overlapping 

image

Hemrajbharti123:
Decrease the size of `short book by category ` 


image

resources/views/knowledgecafe/library/books/index.blade.php
```diff
@@ -13,17 +13,17 @@
...
-        <div class="col-6">
-            <h1>Books({{count($books)}})</h1>
+        <div class="col-4">
+            <h3>Books({{count($books)}})</h3    >
         </div>
         @can('library_books.create')
-        <div class="col-6">
-            <a href="{{ route('books.create') }}" class="btn btn-success float-right"><i class="fa fa-plus mr-1"></i> Add New Book</a>
+        <div class="col-8 xs-6">
+            <a href="{{ route('books.create') }}" class="btn btn-success float-right xs-2"><i class="fa fa-plus mr-1"></i> Add New Book</a>
```
Ayush8923, line 21: What are the means of this class `xs-2`? I think this class is not working.
The correct class is this `col-xs-2`
```suggestion
            <a href="{{ route('books.create') }}" class="btn btn-success float-right xs-2"><i class="fa fa-plus mr-1"></i> Add New Book</a>
```

Ayush8923 (changes requested):
Left some comments.

Ayush8923 (changes requested):
## Testing Feedbacks

**Current Behaviour**
- As far as I can see, the `Books` and the `Search` bar are not aligned, please take refer below screenshot.
image
&#8204;

**Expected Behaviour**

image

resources/views/knowledgecafe/library/books/index.blade.php
```diff
@@ -13,17 +13,17 @@
     <br>
     <br>
     <div class="row">
-        <div class="col-6">
-            <h1>Books({{count($books)}})</h1>
+        <div class="col-4 col-xs-6 pl-4">
+            <h4>Books({{count($books)}})</h4>
```
Ayush8923, line 17:
```suggestion
            <h3>Books({{count($books)}})</h3>
```

Ayush8923 (changes requested):
One small comment. else looks good to me.

Ayush8923 (approved):

Issue #2917, opened 2023-03-29
Books tiles should align centre - Mobile


mohitgusain:
**Describe the bug**
In Mobile, book tiles are left aligned, they should come to the centre of the screen in mobile 

**Screenshots**
![image](https://user-images.githubusercontent.com/30264862/228542220-87824566-b0cc-492d-b905-86b3dd9f94c9.png)

Issue #2941, opened 2023-04-04
Responsive HR module opportunities


rajatsajwan:
**Describe the bug**
In mobile, The opportunities page should be shown fine with proper font sizes and weights. So we need to make it responsive.

ScreensShots
![image](https://user-images.githubusercontent.com/117888235/229852443-1ac16ee4-848a-4c46-a414-2c320a14f5bd.png)

Expected behaviour
The book page should be shown properly in screen sizes

 

- [ ] 360px

 

- [ ] 768px

 

- [ ] 1200px

 

- [ ] 1440px

Additional details

- [x] Read how variable.scss file works
- [ ] You may need to read about [bootstrap](https://getbootstrap.com/docs/4.0/utilities/sizing/) for classes that are used to change sizes of elements and text

PR #2945, opened 2023-04-05
2917 books tiles should align centre mobile

MPandA00:
#2917

### Description

books tiles are center aligned - mobile

### Checklist:


GauravGusain98:
The implementation is not correct. 
1. You only need to add changes that are needed. In your code changes, you have removed a file and added many changes in app.css which I don't think would be needed for this task
2. I've assigned to task to make the complete book page responsive to someone. So you can skip this task and can start working on this task https://github.com/ColoredCow/portal/issues/2941

Issue #2948, opened 2023-04-05
Add Authorization For CodeTrek Page


Hemrajbharti123:
**Describe the Task**

1. So anyone with role employee, admin and super admin should be able to open codetrek page. Currently only admin and super admin has codetrek option in the menu
2. If someone is trying to open the page using link then you have implement laravel policies like we have in invoice module so that if a user doesnot have permission codetrek.view then the page should not be open and error page should be shown like below.

![image](https://user-images.githubusercontent.com/121417679/230071676-5a4bfb69-6ec2-4582-93cd-cfec48875dd1.png)

**Action Items**

- [ ] Read about Authorization in Laravel https://laravel.com/docs/8.x/authorization .
- [ ] Add  CodeTrek option in the menu for employees as well as admin and super admin.
- [ ] Implement Laravel policies for the CodeTrek page, and restrict access to users who don't have the "codetrek.view" permission.
- [ ] Display an error page if a user tries to access the codetrek page without the required permission.
- [ ] Update the documentation to reflect these changes.

PR #2954, opened 2023-04-06
feature/2948/add-authorization-for-code-codeTrek-page

Hemrajbharti123:
Targets #2948 

### Description

So anyone with role employee, admin and super admin should be able to open CodeTrek page. Currently only admin and super admin has CodeTrek option in the menu
If someone is trying to open the page using link then you have implement Laravel policies like we have in invoice module so that if a user does not have permission CodeTrek.view

### Checklist:

### DIFF
--- a/Modules/CodeTrek/Http/Controllers/CodeTrekController.php
+++ b/Modules/CodeTrek/Http/Controllers/CodeTrekController.php
@@ -7,13 +7,17 @@ use Illuminate\Http\Request;
 use Modules\CodeTrek\Entities\CodeTrekApplicant;
 use Modules\CodeTrek\Http\Requests\CodeTrekRequest;
 use Modules\CodeTrek\Services\CodeTrekService;
-use Illuminate\Support\Facades\Gate;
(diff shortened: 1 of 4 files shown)


Hemrajbharti123:
If an unauthorized user try to open CodeTrek.view directly by using URL this error is shown to him.

image

Modules/CodeTrek/Http/Controllers/CodeTrekController.php
```diff
@@ -20,6 +21,10 @@ public function __construct(CodeTrekService $service)
      */
     public function index()
     {
+        if (Gate::denies('codetrek_applicant.view')) {
+            abort(403, 'Unauthorized access');
+        }
+
```
GauravGusain98, line 27: Lets use the approach that is used throughout the portal for policies. You can search for invoicePolicy file and see how it is working. Based on it you can implement policies using the Policy file that exists for codetrek

GauravGusain98 (changes requested):
Please check the comment

GauravGusain98 (approved):
LGTM

PR #2962, opened 2023-04-09
responsive-ledger-accounts-page

MayankMukul:
Targets #{2944}

### Checklist:

### DIFF
--- a/Modules/Invoice/Resources/views/ledger-accounts/index.blade.php
+++ b/Modules/Invoice/Resources/views/ledger-accounts/index.blade.php
@@ -8,7 +8,7 @@
             <div class="card-body">
                 <form action="{{ route('ledger-accounts.index') }}" method="GET">
                     <div class='d-flex'>
-                        <div class='w-30p'>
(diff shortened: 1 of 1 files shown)


Hemrajbharti123:
### Local Testing Feedback

Compared to other pages in the portal width of the table is too large than others 

image

GauravGusain98:
> ### Local Testing Feedback
> Compared to other pages in the portal width of the table is too large than others
> 
> image

This is correct. For now it will not create any issue. The task was to make it responsive, so you need to check if the table is looking fine in small screen or not.

GauravGusain98:
@MayankMukul In small screen i.e 375px. The page UI is not looking fine. 

You need to decrease the font size as well so that they come in one line. I think you have just made changes to table. Also, if you are a user then do you think this UI is looking good? I guess not, So when submitting your PR don't just think of completing the task also see what you have made and how a user will feel by seeing the changes you have done
image

Ayush8923:
Closing this PR because suggested changes are not implemented and the Assignee is not working in this.

Issue #2960, opened 2023-04-08
Project Search Functionality in Project Resource Page


rahulpanwar16:
**Describe the task**
Currently we don't have any search functionality in project resource page. So we want to create a search functionality in which we want to search the project name. You can take reference from the projects page 

**Expected behavior**
Attached a screenshot below : 
![image1](https://user-images.githubusercontent.com/106467597/230722865-afbbd392-dc1c-4727-a830-b040a3dfccbd.jpg)

PR #2927, opened 2023-04-01
Added the link to project page on project resource page in project name

Priyanka-Kaintura:
Targets #2924 

### Description

Added the link to project page on project name in project resource page.
### Checklist:

- [x] I have performed a self-review of my own code.

PR #2991, opened 2023-04-15
Project search functionality in project resource page

Priyanka-Kaintura:
Targets #2960

### Description
Added the search functionality to the Project Resource page

### Checklist:


Modules/Project/Http/Controllers/ProjectController.php
```diff
@@ -148,7 +148,7 @@ public function projectFTEExport(Request $request)
 
     public function projectResource()
     {
-        $resourceData = $this->service->getProjectsWithTeamMemberRequirementData();
+        $resourceData = $this->service->getProjectsWithTeamMemberRequirementData(request()->input('name'));
```
Ayush8923, line 151:
```suggestion
        $resourceData = $this->service->getProjectsWithTeamMemberRequirementData(request()->all());
```

Modules/Project/Resources/views/resource-requirement.blade.php
```diff
@@ -25,7 +34,7 @@
                 </thead>
                 <tbody>
                     @can('projects.view')
-                        @foreach ($resourceData['data'] as $clientName =>$clientProjectsData)
+                        @forelse ($resourceData['data'] as $clientName =>$clientProjectsData)
```
Ayush8923, line 37: Why did you change `foreach` to `forelse` here?

Modules/Project/Services/ProjectService.php
```diff
@@ -462,13 +462,17 @@ private function formatProjectFTEFOrExportAll($employees)
         return $teamMembers;
     }
 
-    public function getProjectsWithTeamMemberRequirementData()
+    public function getProjectsWithTeamMemberRequirementData($input)
```
Ayush8923, line 465: Please update the all places.
```suggestion
    public function getProjectsWithTeamMemberRequirementData($request)
```

Modules/Project/Services/ProjectService.php
```diff
@@ -462,13 +462,17 @@ private function formatProjectFTEFOrExportAll($employees)
...
     {
-        $projectsWithTeamMemberRequirement = Project::with('client')->status('active')
+        $query = Project::with('client')->status('active')
```
Ayush8923, line 467: Please update the all places. Please use the correct naming convention.
```suggestion
        $projects = Project::with('client')->status('active')
```

Ayush8923 (changes requested):
Left some comments.

Modules/Project/Resources/views/resource-requirement.blade.php
Priyanka-Kaintura: Because sir, the result of the search may even be null.

Modules/Project/Http/Controllers/ProjectController.php
Priyanka-Kaintura: Okay sir, I will do that.

Modules/Project/Services/ProjectService.php
Priyanka-Kaintura: Okay sir.

Modules/Project/Services/ProjectService.php
Priyanka-Kaintura: SIr, I didn't name it '$projects' because the resultant is not projects data, rather the query is aggregated further and then after applying the get() function we get the the collection as named in the code. So do I still need to rename the variable?

Modules/Project/Resources/views/resource-requirement.blade.php
```diff
@@ -39,7 +48,15 @@
                             <tr>
                                 <td class="w-33p">
                                     <div class="pl-1 pl-xl-2">
-                                        {{ $projectName }}
+                                        @php
+                                            $project = $projectData['object'];
+                                            $team_member_ids = $project->getTeamMembers->pluck('team_member_id')->toArray();
+                                        @endphp
```
Ayush8923, line 54: Please you can do all these logic operations in the service file. Then here you can use the variable directly.

Modules/Project/Services/ProjectService.php
Ayush8923: Can we write like this? Please test locally once and also please test all the edge cases. and check this query work as expected.
Please let me know, it works or not.

```
$projectsWithTeamMemberRequirement = Project::query()
    ->with('client')
    ->status('active')
    ->withCount('getTeamMembers as team_member_count')
    ->withSum('resourceRequirement as team_member_needed', 'total_requirement')
    ->havingRaw('team_member_needed - team_member_count')
    ->when($request, function ($query, $request) {
        return $query->where('name', $request);
    })
    ->get();
```

Modules/Project/Services/ProjectService.php
```diff
@@ -462,13 +462,17 @@ private function formatProjectFTEFOrExportAll($employees)
...
             ->withCount('getTeamMembers as team_member_count')
             ->withSum('resourceRequirement as team_member_needed', 'total_requirement')
-            ->havingRaw('team_member_needed - team_member_count')
-            ->get();
+            ->havingRaw('team_member_needed - team_member_count');
+
+        if ($request != null) {
+            $query->where('name', $request);
+        }
+        $projectsWithTeamMemberRequirement = $query->get();
```
Ayush8923, line 475: If you are using the above-suggested code, then please remove this code.

and also one more suggestion in this code. You can write like this if the condition. `empty` is a PHP function that checks whether a variable is empty or not.
```suggestion
        if ((!empty($request->name)) {
            $query->where('name', $request);
        }
```

Ayush8923 (changes requested):
Left some suggestions @Priyanka-Kaintura.

Modules/Project/Services/ProjectService.php
Priyanka-Kaintura: Yes Sir, I'll keep that in mind.

Modules/Project/Resources/views/resource-requirement.blade.php
Priyanka-Kaintura: sir, I don't know why these changes are visible here because these are the changes from previous pr which is now merged and closed. Please tell me if I still need to make these changes 
PR: https://github.com/ColoredCow/portal/pull/2927

Modules/Project/Services/ProjectService.php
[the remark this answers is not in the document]
Priyanka-Kaintura: Sir, it gives the result as expected and works fine. Thank You.

Modules/Project/Services/ProjectService.php
Priyanka-Kaintura: Sir, it gives the result as expected and works fine. Thank You!

Modules/Project/Resources/views/resource-requirement.blade.php
GauravGusain98: Just take the pull from master branch and push it to this branch

Modules/Project/Services/ProjectService.php
```diff
@@ -462,13 +462,18 @@ private function formatProjectFTEFOrExportAll($employees)
...
-            ->get();
+        $projectsWithTeamMemberRequirement = Project::query()
+        ->with('client')
+        ->status('active')
+        ->withCount('getTeamMembers as team_member_count')
+        ->withSum('resourceRequirement as team_member_needed', 'total_requirement')
+        ->havingRaw('team_member_needed - team_member_count')
+        ->when($request, function ($query, $request) {
+            return $query->where('name', $request['name']);
+        })
```
GauravGusain98, line 475: It should not be a strict check. Even a word or as substring is present in name, show that data. Also when we search an empty string then filter should not be applied. Similar to how we do in project page search

GauravGusain98 (changes requested):

Modules/Project/Resources/views/resource-requirement.blade.php
Priyanka-Kaintura: okay sir.

GauravGusain98 (approved):
LGTM

Issue #2930, opened 2023-04-04
Make the invoice issue date ,end date of the invoice term.


mohitsharma-22:
**Is your feature request related to a problem? Please describe.**
As per CA, We need to create the invoice in the same month date when we have given the service. The main issue is at the end of the financial year. where March invoices are created in April. Many clients also requested to create the invoice with date of 31 March and not with 1st April.

**Describe the solution you'd like**
The invoice we create from portal has the invoice date on which we generate the invoice.  Please make the invoice issue date end date of Term. For example invoices of March month is of term 1st March - 31st March. The invoice issue date should be 31st March. and not the 1st April, even if we create and send the invoice on 1st April

<img width="482" alt="Screenshot 2023-04-04 at 4 18 09 PM" src="https://user-images.githubusercontent.com/16541548/229769540-d11d8abf-f3cb-402f-8ce5-d269744408e5.png">

AmishaBisht:
The issue date should be 31 if the invoice is created on the 1st what should be the date for other cases like the invoice is generated on 12 April?

AmishaBisht:
![Screenshot (157)](https://user-images.githubusercontent.com/90472056/232453363-fbcbf9b3-820a-44cc-853e-7924e3cbd293.png)

GauravGusain98:
@AmishaBisht 
So we store the billing date of a client in the client_billing_details table. This date is the day when the invoice should be sent.

Currently while generating an invoice we show the issue date as the date when we send the invoice but what we want is that the issue date should be 1 day less than the billing date of the invoice i.e (term_end_date)

AmishaBisht:
@GauravGusain98 in the above case the issue date should be 16 April?

GauravGusain98:
> @GauravGusain98 in the above case the issue date should be 16 April?

Yes, if the billing date of client is 17

PR #2996, opened 2023-04-17
fixed the issue date of invoice

AmishaBisht:
Targets #2930 

### Description
I have changed the issue date for invoice to the termEndDate that is the ending date of billing. 

### Checklist:

### DIFF
--- a/Modules/Invoice/Resources/views/render/invoice-template.blade.php
+++ b/Modules/Invoice/Resources/views/render/invoice-template.blade.php
@@ -302,7 +302,7 @@
                                         <td align="right">
                                             <p>{{ $termText }}</p>
                                             <p>{{ $invoiceNumber }}</p>
-                                            <p>{{ $termPeriod['term_end_date']->format('F d, Y') }}</p>
+                                            <p>{{ ($termEndDate) }}</p>
                                             <p>{{ date('F d, Y', strtotime($invoiceData['due_on'])) }}</p>
                                         </td>
                                     </tr>
--- a/Modules/Invoice/Http/Controllers/InvoiceController.php
+++ b/Modules/Invoice/Http/Controllers/InvoiceController.php
@@ -87,6 +87,7 @@ class InvoiceController extends Controller
             'client_id' => $request->client_id,
             'project_id' => $request->project_id,
             'term' => today(config('constants.timezone.indian'))->subMonth()->format('Y-m'),
+            'termEndDate' => $request->termEndDate,
             'sent_on' => today(config('constants.timezone.indian')),
             'due_on' => today(config('constants.timezone.indian'))->addDays(6),
             'period_start_date' => $request->period_start_date,
--- a/Modules/Invoice/Services/InvoiceService.php
+++ b/Modules/Invoice/Services/InvoiceService.php
@@ -518,6 +518,7 @@ class InvoiceService implements InvoiceServiceContract
             $billingStartMonth = Carbon::parse($data['period_start_date'])->format('M');
         }
 
+        $termEndDate = $client ? $client->getMonthEndDateAttribute(1)->format('F d, Y'): $project->client->getMonthEndDateAttribute(1)->format('F d, Y');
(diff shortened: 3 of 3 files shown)


Modules/Invoice/Resources/views/render/invoice-template.blade.php
```diff
@@ -302,7 +302,7 @@
                                         <td align="right">
                                             <p>{{ $termText }}</p>
                                             <p>{{ $invoiceNumber }}</p>
-                                            <p>{{ date('F d, Y', strtotime($invoiceData['sent_on'])) }}</p>
+                                            <p>{{ date('F d, Y', strtotime(now()->subMonthNoOverflow()->endOfMonth()->toDateString())) }}</p>
```
GauravGusain98, line 305: What it will return is billing date is 1 and today is 17 April. 

The expected output should be 31 march 2023

GauravGusain98 (changes requested):

Modules/Invoice/Resources/views/render/invoice-template.blade.php
AmishaBisht: yes it is giving 31st March, here its billing date is 1 in database
Screenshot (159)

GauravGusain98: OK what if the billing date is 15?

AmishaBisht: @GauravGusain98 I can directly use the term-end-date in the place of issue date?

Modules/Invoice/Resources/views/render/invoice-template.blade.php
```diff
@@ -302,7 +302,7 @@
                                         <td align="right">
                                             <p>{{ $termText }}</p>
                                             <p>{{ $invoiceNumber }}</p>
-                                            <p>{{ date('F d, Y', strtotime($invoiceData['sent_on'])) }}</p>
+                                            <p>{{ date('F d, Y', strtotime($invoiceData['term'])) }}</p>
```
GauravGusain98, line 305: are you sure it si giving correct answers in all different cases? Did you tried testing it with different cases. 

Can you share screenshot of the result for different cases

AmishaBisht: @GauravGusain98  this function is calculating the $termEndDate but when I am using this it is only returning the month. I need to make another function for date?
Screenshot (161)

GauravGusain98: No you don't need to create another function. Just check how the `$invoiceData['term']` is defined

AmishaBisht: @GauravGusain98 

there modifications have been made

AmishaBisht: Screenshot (163)
I wrote this query to get the term_end_date in the ready to send invoices.

GauravGusain98: But how you are getting term_end_date from the database when the invoice entry is not yet created?

GauravGusain98: The invoice that you are going to show in ready to send invoices is not yet stored in the database. From the database you can only get the end_date of invoices are sent.

AmishaBisht: @GauravGusain98 
Screenshot (167)
the preview URL is hitting this function according to my inderstanding.

AmishaBisht: @GauravGusain98 
Screenshot (168)
the preview URL is hitting this function according to my understanding.

GauravGusain98:
@AmishaBisht This is correct. As I've already done these changes and release the PR. So I'm closing this PR. 

But Congrats on understanding the logic that as important

Issue #2992, opened 2023-04-15
Clear the Name Filter From Search Functionality in CodeTrek Page


Himanshunegi0332:
**Describe the task**
Currently we don't have any feature that if a user has applied the name filter then how to remove that filter, for this we have to again apply the filter with no value to reset it. For this we have to give him a functionality to remove the filter also.

**Expected behavior**
In this is if a user click on the 'To clear the filter click here X' it redirect the user to CodeTrek page.
![Screenshot (88)](https://user-images.githubusercontent.com/108523408/232244809-50065034-a99a-42b8-bca0-d37e565614ae.png)

**References**
Specify any reference material / links

PR #2997, opened 2023-04-17
To Clear The Filter From Search Functionality

Himanshunegi0332:
Targets #2992 
### Description
Currently we don't have any feature that if a user has applied the name filter then how to remove that filter, for this I have created a functionality to remove the filter.

### Checklist:

### DIFF
--- a/Modules/CodeTrek/Resources/views/index.blade.php
+++ b/Modules/CodeTrek/Resources/views/index.blade.php
@@ -40,6 +40,9 @@
         </div>
         </form>
     </ul>
+    <div class="text-right fz-14 mr-11 mt-3">
+        <span>Result for 'search text'</span><a href="/codetrek"> To clear the filter click here X </a>
+    </div>
     @if (request()->input('tab', 'active') == 'active' || request()->tab == 'applicants')
         <div>
             <br>
--- a/Modules/CodeTrek/Resources/views/index.blade.php
+++ b/Modules/CodeTrek/Resources/views/index.blade.php
@@ -32,13 +32,13 @@
         </li>
         </div>
         <form class="d-md-flex justify-content-between ml-md-3"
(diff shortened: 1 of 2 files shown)


Modules/CodeTrek/Resources/views/index.blade.php
```diff
@@ -40,6 +40,9 @@
         </div>
         </form>
     </ul>
+    <div class="text-right fz-14 mr-11 mt-3">
+        <span>Result for 'search text'</span><a href="/codetrek"> To clear the filter click here X </a>
```
GauravGusain98, line 44: This `'search text'` should be replaced with the text that user has searched. If user has not search anything then the whole sentence should not be seen

GauravGusain98 (changes requested):
Please check the comment

Modules/CodeTrek/Resources/views/index.blade.php
```diff
@@ -40,6 +40,11 @@
         </div>
         </form>
     </ul>
+    @if (request()->get('name'))
+    <div class="text-right fz-14 mr-11 mt-3">
+        <span>Result for '{{ request()->get('name') }}'</span><a href="/codetrek"> To clear the filter click here X </a>
```
GauravGusain98, line 45: The UI is not looking good. Can we make it like this. Just add a simple clear button
image

GauravGusain98 (changes requested):

GauravGusain98:
### Testing Feedbacks
- I've searched `Gaurav Gusain` but the field is only showing `Gaurav`

Himanshunegi0332:
> ### Testing Feedbacks
> * I've searched `Gaurav Gusain` but the field is only showing `Gaurav`
> 
> image

GauravGusain98 (approved):
LGTM

Issue #2943, opened 2023-04-05
Responsive invoice page


rajatsajwan:
Describe the bug
In mobile, The invoice page(finance module) should be shown fine with proper font sizes and weights. So we need to make it responsive.

**Screenshots**

![image](https://user-images.githubusercontent.com/117888235/229981831-9251d50f-9bf7-4872-ac57-0eae9784285f.png)

**Expected behaviour**
The book page should be shown properly in screen sizes

- [x]  360px

**Additional details**

- [x]  Read how variable.scss file works

- [x]  You may need to read about [bootstrap](https://getbootstrap.com/docs/4.0/utilities/sizing/) for classes that are used to change sizes of elements and text

PR #2998, opened 2023-04-18
make the invoice page responsive

AdityaDabral:
Targets #2943 

### Description
Added a row in the invoice page and made some additional changes like changing the size of buttons to make the page responsive.

### Checklist:


Modules/Invoice/Resources/views/index-filters.blade.php
```diff
@@ -1,6 +1,8 @@
 <form action="{{ route('invoice.index')  }}" id="invoiceFilterForm">
     <input type="hidden" name="invoice_status" value="{{ request()->input('invoice_status', 'sent') }}">
+
```
Hemrajbharti123, line 3: remove this unnecessary space

Hemrajbharti123 (dismissed):

GauravGusain98:
@AdityaDabral The buttons and filters UI for small screen is not looking good. Maybe you can try to utilize the blank space and instead of showing filters options vertically put 2 filters in a row. 

Also button text can be reduced so that it comes in one line and the buttons can be shifted to right side
image

rahulpanwar16:
Looks good now.

rahulpanwar16 (dismissed):

GauravGusain98:
### Testing feedbacks
The table is going out of the page. Make it scrollable so that it stays within the page
image

GauravGusain98 (approved):
Not everything is made responsive but approving as we can go with the pieces that are fixed

PR #3020, opened 2023-04-24
Created a CodeTrek Seeder

MayankBisht8630:
Targets #{3016}

### Description
created a seeder 
### Checklist:


GauravGusain98:
@MayankBisht8630 

Remove the curly braces
image

config/database.php
GauravGusain98, line 1: Not needed. You should have provided it in the env file

GauravGusain98: If env data is not getting read then you should run `php artisan config:cache` to refresh the config cache

app/Providers/AppServiceProvider.php
```diff
@@ -16,6 +16,7 @@ class AppServiceProvider extends ServiceProvider
      */
     public function boot()
     {
+        Schema::defaultStringLength(191);
```
GauravGusain98, line 19: not needed

GauravGusain98 (changes requested):
Please check the comment and fix the checks that are getting failed

Ayush8923 (changes requested):
Please revert the `package.lock.json` file from this PR. This file is not needed.

Issue #2983, opened 2023-04-13
Responsive project client information page 


rahulpanwar16:
**Describe the bug**
In mobile, the edit client information project page the alignment is not correct. Hence, we need to make it responsive. 

**To Reproduce**
Go to the CRM nav bar > Clients > Select any active client > Projects 

**Screenshot**

![Screenshot (52)](https://user-images.githubusercontent.com/106467597/231659900-bb6f0786-aee7-4f42-9fa2-fdd10f94d6ff.png)

**Expected behavior**
The page should be shown properly in screen sizes : 
 - [x] 360px
- [x] 375px
- [x] 425px
- [x] 550px

**Additional context**

- [x] Read how variable.scss file works.
- [x] You may need to read about [bootstrap](https://getbootstrap.com/docs/4.0/utilities/sizing/) for classes that are used to change sizes of elements and text.

PR #3029, opened 2023-04-25
Corrected the alignment of client project page content

Priyanka-Kaintura:
Targets #2983

### Description
Corrected the alignment of client project page content.

### Checklist:


Priyanka-Kaintura:
@GauravGusain98 
Sir, I have checked the content alignment for almost all range of screen sizes and especially for edge cases with screen size xs, sm, md, l, xl as suggested. But if there is still a problem then please let me now.

Modules/Client/Resources/views/subviews/edit-client-projects.blade.php
```diff
@@ -1,4 +1,4 @@
-<div class="card">
+<div class="overflow-x-scroll card" style="min-width:660px;">
```
Ayush8923, line 1: Please sue bootstrap classes instead of inline-css.

Ayush8923 (changes requested):

Modules/Client/Resources/views/subviews/edit-client-projects.blade.php
Priyanka-Kaintura: @Ayush8923 Sir I didn't find any specific bootstrap class to set min-width and so used inline style. I even went through the documentation but didn't find any class of this sort.

resources/sass/_variables.scss
```diff
@@ -352,3 +353,7 @@ $grid-breakpoints: (
     xl: 1200px,
     hd: 1440px
 );
+
+$min-width: (
+    '660': 660px,
+);
```
Ayush8923, line 359: I don't think, we need this. Please remove this. Then test it locally. And let me know, it's working fine or not.

Ayush8923 (changes requested):
One comment else looks good to me.

Ayush8923 (approved):
Looks good to me.

PR #3070, opened 2023-05-08
Make the HR recruitment page responsive

Akshiteshnaithani:
Targets #{2942}
Here is the screenshot of the page 
image


Ayush8923:
@Akshiteshnaithani Can you move the `Application` heading slightly to the right? So that it aligns with the search bar.

Ayush8923 (approved):

PR #3117, opened 2023-06-02
Removed policy for CodeTrek module

GauravGusain98:
### Changes

- CodeTrek module existing policies are not working as expected. So removing them for now.


GauravGusain98:
Self reveiwed

Issue #3129, opened 2023-06-07
 Display Only Active Applicants in CodeTreak Applicant Sidebar and Resolve Overlapping with Other Modules.


Akshiteshnaithani:
**Description :**

In the CodeTrek module, there are both active and inactive applicants. However, the CodeTrek Applicant Sidebar currently displays all applicants. To enhance usability and efficiency, the sidebar should be modified to display only the active applicants. Additionally, when opening the sidebar, it overlaps with the tables in other modules, causing visual and functional issues.

**Current Behavior :**

- The CodeTreak Applicant Sidebar displays all applicants, including both active and inactive ones.
- Opening the sidebar overlaps with tables in other modules, resulting in a visually cluttered and problematic user experience, as illustrated in the attached screenshot.

**Screenshot :**
![image](https://github.com/ColoredCow/portal/assets/121417968/47a610c4-f166-4a9e-81bb-e5bca727644a)

**Expected Behavior:**

- The CodeTrek Applicant Sidebar should only display active applicants.
- When opening the sidebar, it should not overlap with the tables in other modules.

PR #3130, opened 2023-06-07
#3129 Display Only Active Applicants in CodeTreak Applicant Sidebar and Resolve Overlapping with Other Modules.

jaskaransingh304:
Targets #3129 

### Description

Made changes so only to display active applicants in CodeTreak Applicant Sidebar.
Resolved Overlapping of CodeTreak Applicant Sidebar with other modules.
These changes were required as the user will get to see only active applicant. Opening the sidebar does not overlaps with tables in other modules, resulting in better view for user.
 This pull request , solves the issue and displays Active applicants only in CodeTreak Applicant Sidebar and the sidebar does not overlaps with other modules.

### Checklist:

### Approach -
Checked the view page and tried to figure , how data is passed in view page.
Checked the css file for sidebar.

### Expected Time - 4-6  hours.
Code review and documentation: 0.5-1 hour.
then ,
Making the required changes : 1 hour.

For testing, code review, and documentation : 1 hour.

### Changes required in code -
As there was no condition to display active applicants, so added a condition where it displays active applicants only instead of all. 
Made changes in stack order of task i.e z-index. So sidebar opens up and does not overlaps.

### DIFF
--- a/Modules/CodeTrek/Resources/views/sidebar.blade.php
+++ b/Modules/CodeTrek/Resources/views/sidebar.blade.php
@@ -2,14 +2,12 @@
     <h5 class="fw-bold border-bottom pb-2">CodeTrek Applicants</h5>
     <ul class="applicant-list list-unstyled">
         @foreach ($codeTrekApplicants as $codeTrekApplicant)
-            @if ($codeTrekApplicant->status == 'active')
-                <li data-id="{{ $codeTrekApplicant->id }}" class="d-flex align-items-center">
-                    <i class="fa fa-user mr-1"></i>
-                    <h6 class="ml-1">
-                        {{ $codeTrekApplicant->first_name }} {{ $codeTrekApplicant->last_name }}
-                    </h6>
-                </li>
-            @endif
+        <li data-id="{{ $codeTrekApplicant->id }}" class="d-flex align-items-center">
+            <i class="fa fa-user mr-1"></i>
+            <h6 class="ml-1">
+                {{ $codeTrekApplicant->first_name }} {{ $codeTrekApplicant->last_name }}
+            </h6>
+        </li>
         @endforeach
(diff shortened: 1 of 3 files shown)


jaskaransingh304:
Attaching a screenshot for reference -
Screenshot (4)

Modules/CodeTrek/Resources/views/sidebar.blade.php
```diff
@@ -2,12 +2,14 @@
...
-            </h6>
-        </li>
+            @if ($codeTrekApplicant->status == 'active')
+                <li data-id="{{ $codeTrekApplicant->id }}" class="d-flex align-items-center">
+                    <i class="fa fa-user mr-1"></i>
+                    <h6 class="ml-1">
+                        {{ $codeTrekApplicant->first_name }} {{ $codeTrekApplicant->last_name }}
+                    </h6>
+                </li>
+            @endif
```
GauravGusain98, line 12: Instead of adding a check here. I guess we can only fetch active candidates from database i.e we can update the query from which we are getting this variable `$codeTrekApplicants`. 

The approach that I am suggesting will reduce the complexity as the loop will not run for all the candidates

GauravGusain98: You can check Harshvardhan PR that he had created for implementing the sidebar and see how he is getting the applicants from database.

GauravGusain98 (changes requested):

Modules/CodeTrek/Resources/views/sidebar.blade.php
jaskaransingh304: Sure , @GauravGusain98 , I will do that.

GauravGusain98 (approved):
LGTM

PR #3132, opened 2023-06-08
Update the CodeTrek installation guide line

Akshiteshnaithani:
### Description
Updated the CodeTrek Guide line.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -125,8 +125,8 @@ Before you start following the guidelines, make sure to go through the [prerequi
        - DB_WORDPRESS_DATABASE: The name of the WordPress database you want to connect to. For example `Coloredcow`.
        - DB_WORDPRESS_USERNAME: The user name which is define in the Wp-config.php. For example `root`.
        - DB_WORDPRESS_PASSWORD: If password is define set the password otherwise it should be null.
-       - DB_WORDPRESS_PREFIX  : The table prefix used by the WordPress installation. This is typically set to `cc_` by default, but it can be different depending on the configuration. 
-       - WORDPRESS_ENABLED: Set this to `true` to enable the integration with the ColoredCow website. If it is set to "false," the integration will not work.
+       - DB_WORDPRESS_PREFIX: The table prefix used by the WordPress installation. This is typically set to `cc_` by default, but it can be different depending on the configuration. 
+       - WORDPRESS_ENABLED: Set this to `true` to enable the integration with the ColoredCow website. If it is set to `false` the integration will not work.
 
     ```sh
     DB_WORDPRESS_DATABASE=Coloredcow
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -122,11 +122,11 @@ Before you start following the guidelines, make sure to go through the [prerequi
     4. _(Optional)_ ColoredCow website Configurations
     In case you want to use website integration functionality, then you need to enable `WORDPRESS_ENABLED` as `true` and add wordpress database configurations.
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -129,6 +129,12 @@ Before you start following the guidelines, make sure to go through the [prerequi
     DB_WORDPRESS_PREFIX=
     WORDPRESS_ENABLED=true
     ```
+    DB_WORDPRESS_DATABASE: The name of the WordPress database you want to connect to.For example `Coloredcow`.
+    DB_WORDPRESS_USERNAME: The user name which is difine in the Wp-config.php.For example `root`.
+    DB_WORDPRESS_PASSWORD: If password is define set the password otherwise it should be null.
+    DB_WORDPRESS_PREFIX=  The table prefix used by the WordPress installation. This is typically set to `cc_` by default, but it can be different depending on the configuration. 
+    WORDPRESS_ENABLED: Set this to `true` to enable the integration with the Colored Cow website. If it is set to "false," the integration will not work.
```
Ayush8923, line 136: Can you write these points next to the same points which are already written? I think you don't need to write the same points again.
Something like this:-
```
DB_WORDPRESS_PREFIX=cc_
WORDPRESS_ENABLED=true
```

Ayush8923 (changes requested):
Added one comment. Please check.

docs/installation.md
```diff
@@ -123,13 +123,12 @@ Before you start following the guidelines, make sure to go through the [prerequi
     In case you want to use website integration functionality, then you need to enable `WORDPRESS_ENABLED` as `true` and add wordpress database configurations.
 
     ```sh
-    DB_WORDPRESS_DATABASE=
-    DB_WORDPRESS_USERNAME=
+    DB_WORDPRESS_DATABASE=Coloredcow
+    DB_WORDPRESS_USERNAME=root
     DB_WORDPRESS_PASSWORD=
-    DB_WORDPRESS_PREFIX=
+    DB_WORDPRESS_PREFIX=cc_
```
GauravGusain98, line 129: Can you also explain the meaning of these variables as you had done earlier.

GauravGusain98: Just put them before the code block

GauravGusain98 (changes requested):

Ayush8923 (dismissed):
Added some more comments. Please check.

docs/installation.md
```diff
@@ -122,13 +122,21 @@ Before you start following the guidelines, make sure to go through the [prerequi
     4. _(Optional)_ ColoredCow website Configurations
     In case you want to use website integration functionality, then you need to enable `WORDPRESS_ENABLED` as `true` and add wordpress database configurations.
 
+       - DB_WORDPRESS_DATABASE: The name of the WordPress database you want to connect to. For example `Coloredcow`.
+       - DB_WORDPRESS_USERNAME: The user name which is define in the Wp-config.php. For example `root`.
+       - DB_WORDPRESS_PASSWORD: If password is define set the password otherwise it should be null.
+       - DB_WORDPRESS_PREFIX  : The table prefix used by the WordPress installation. This is typically set to `cc_` by default, but it can be different depending on the configuration. 
+       - WORDPRESS_ENABLED: Set this to `true` to enable the integration with the ColoredCow website. If it is set to "false," the integration will not work.
```
tarunnjoshi:
```suggestion
       - WORDPRESS_ENABLED: Set this to `true` to enable the integration with the ColoredCow website. If it is set to "false" the integration will not work.
```

GauravGusain98 (approved):

Issue #3167, opened 2023-06-19
Automated generation of assessment cards.


Bahugunajii:
**Current behaviour**
Currently the assessment cards being generated corresponding to an employee are somewhat manual.

**Expected behaviour**
 Since the review system is quarterly, we can have a functionality such that an assessment cards corresponding to an employee are generated automatically with the initial value of Self review, Mentor review, HR review and Manager review to be pending. Note that since the assessment duration can be modified in future (as it can be every 6 months or yearly) so we need to make the functionality such that the duration can be modified from config.

LavishRawat:
@Akshiteshnaithani Please add your understanding and breakdown of the task.

Akshiteshnaithani:
## Breakdown Of Issue 
- [x] Understand the current manual process of generating assessment cards for employees.
- [ ]Develop the necessary code logic to automate the generation of assessment cards. This may involve using a cron job or a 
    scheduled task to trigger quarterly.
- [ ] Create the assessment cards dynamically for each employee. Assign the initial values of Self review, Mentor review, HR review, and Manager review as "pending" to each card.
- [ ] Conduct thorough testing to ensure the automated generation of assessment cards is working as expected.

PR #3174, opened 2023-06-20
Automated review system

Akshiteshnaithani:
Targets #3167 

### Description
This task aims to automate the review system by implementing scheduled tasks using cron jobs. The current manual process of generating assessment cards for employees will be replaced with an automated system that generates assessment cards at regular intervals(Quarterly).

### Checklist:

### DIFF
--- a/Modules/HR/Console/SystemReviewQuaterly.php
+++ b/Modules/HR/Console/SystemReviewQuaterly.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Modules\HR\Console;
+
+use Carbon\Carbon;
+use Illuminate\Console\Command;
+use Modules\HR\Entities\Assessment;
+use Modules\HR\Entities\Employee;
+use Modules\HR\Entities\IndividualAssessment;
+
+class SystemReviewQuaterly extends Command
+{
+    /**
+     * The console command name.
+     *
+     * @var string
+     */
+    protected $name = 'hr:quarterly-review-system';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Generates quarterly assessment cards for employees.';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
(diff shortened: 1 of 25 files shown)


app/Console/Kernel.php
```diff
@@ -28,7 +29,8 @@ class Kernel extends ConsoleKernel
         EndedProject::class,
         FixedBudgetProject::class,
         NotificationToProjectTeamMembersToUpdateEffortOnGoogleChat::class,
-        JobExpiredEmailToHr::class
+        JobExpiredEmailToHr::class,
+        SystemReviewQuaterly::class
```
Ayush8923, line 33: Please update all places accordingly
```suggestion
        QuarterlyReviewSystemForEmployee::class
```

Modules/HR/Console/SystemReviewQuaterly.php
```diff
@@ -0,0 +1,72 @@
...
+use Modules\HR\Entities\IndividualAssessment;
+
+class SystemReviewQuaterly extends Command
+{
+    /**
+     * The console command name.
+     *
+     * @var string
+     */
+    protected $name = 'hr:quarterly-review-system';
```
Ayush8923, line 18:
```suggestion
    protected $name = 'employee:quarterly-review-system-for-employee';
```

Ayush8923 (changes requested):
Added some comments. Please check. Also I think, we don't need the `EmployeeController`, so please revert this file.

Akshiteshnaithani:
> Added some comments. Please check. Also I think, we don't need the `EmployeeController`, so please revert this file.

CI checks were getting fail due to `EmployeeController`  file.

Ayush8923:
As per the offline discussion, please use SRP(Single responsibility principal) principal.

Akshiteshnaithani:
> As per the offline discussion, please use SRP(Single responsibility principal) principal.

Implemented the SRP principal.

Akshiteshnaithani:
### New Requirement :
currently we have added a `Review Status` filter in our Working staff page so that we could filter out all the employees whose review is pending or in-progress or completed. Now we need to remove that filter and whenever I will land onto the working staff page, all the employee should show up whose review is pending or in-progress or completed

Bahugunajii (approved):

Issue #3113, opened 2023-05-31
Slow loading of HR recruitment page on the Portal


Harshvardhan80:
**Describe the task**
The HR recruitment page on the portal is experiencing significant performance issues, with page load times taking 10+ seconds. This slow loading makes it difficult and frustrating to use the page effectively.

**Steps to Reproduce:**

1. Access the HR recruitment page on the portal.
![Screenshot 2023-05-31 194505](https://github.com/ColoredCow/portal/assets/112154429/9d288e20-426f-4a3a-9e38-f5d640041bd2)

2. Observe the significant delay in page load times.

**Expected behaviour**
The HR recruitment page should load within a reasonable timeframe, preferably under a few seconds, to ensure smooth user interaction.

**References**
![Screenshot 2023-05-31 195223](https://github.com/ColoredCow/portal/assets/112154429/d8ef483f-502f-440c-98fd-d09a13acc2d5)

Ayush8923:
**Solution to solve this problem:-**

As far as I can check there is a lot of queries are running when the HR recruitment page is open. In the hr_application table, if there is a 150 - 200 rows in that table so the minimum time of the page is 1 - 2 seconds. 
And If there is 10,000 + data in that table so the minimum time of the page is open is 14 sec. So the total queries 11101 queries are executed. I think there are also unnecessary queries we can remove the ones we don't need. So we need to optimize the `index` function of the ApplicationController.

deepankar-sajwan:
@Ayush8923 as per the testing done is the portal it is tasking around `3-4 seconds` to load the `HR recruitment page`.  If the expected time for loading `HR recruitment page` then you can deploy the feature on the production

Ayush8923:
> @Ayush8923 as per the testing done is the portal it is tasking around `3-4 seconds` to load the `HR recruitment page`. If the expected time for loading `HR recruitment page` will be between 3-4 seconds, then you can deploy the feature on the production

@deepankar-sajwan Yes. Now it is taking 2 -3 seconds which is correct as compared to earlier. We will do more code cleanup and optimization in it, so I think it will be better.

deepankar-sajwan:
> Reference

@Ayush8923 you can proceed with the code deployment

PR #3180, opened 2023-06-21
Slow loading of HR recruitment page on the Portal

Harshvardhan80:
Targets #3113 

### Description
When I used to open the HR recruitment page, it would take more than 10 seconds to load. I would like to improve its performance.

### Checklist:

### DIFF
--- a/Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
+++ b/Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
@@ -71,12 +71,12 @@ abstract class ApplicationController extends Controller
             'should_skip_page' => false,
         ]);
 
-        $hr_job_id = request()->has('hr_job_id') ? '&hr_job_id=' . request('hr_job_id') : '';
-        $graduation_year = request()->has('end-year') ? '&end-year=' . request('end-year') : '';
+        $hrJobId = request()->has('hr_job_id') ? '&hr_job_id=' . request('hr_job_id') : '';
+        $graduationYear = request()->has('end-year') ? '&end-year=' . request('end-year') : '';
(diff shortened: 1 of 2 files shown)


Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -71,6 +71,13 @@ public function index()
             'should_skip_page' => false,
         ]);
 
+        $hr_job_id = request()->has('hr_job_id') ? '&hr_job_id=' . request('hr_job_id') : '';
```
LavishRawat, line 74: @Harshvardhan80 can we fix the naming convention here? Use camelCase.

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -71,6 +71,13 @@ public function index()
...
+        $graduation_year = request()->has('end-year') ? '&end-year=' . request('end-year') : '';
```
LavishRawat, line 75: @Harshvardhan80 same here. Fix the naming convention in all places.

LavishRawat (changes requested):

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
Harshvardhan80: yes, sure

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
Harshvardhan80: > yes, sure

Ayush8923 (approved):
Looks good..

LavishRawat (approved):

Issue #3184, opened 2023-06-22
Hide Setting Button for Users with No Entry in Employee Table


Akshiteshnaithani:
**Description:**

Currently, all employees are displayed in the user management system, irrespective of whether they have an entry in the employee table or not. This has led to a problem where the setting button, when clicked, opens a model to modify the working staff of the user. However, the requirement states that the setting button should only be visible if there is a corresponding entry in the employee table for the user.

**To address this issue, the following changes need to be implemented:**

- Update the user management system to check if there is an entry for each user in the employee table.
- If there is no entry in the employee table for a user, hide the setting button for that user in the user management interface.
- Implement a tooltip that appears when hovering over the hidden setting button, providing a message such as `No entry in the employee table for this user`.

These modifications will ensure that the setting button is only visible for users who have a valid entry in the employee table, and provide a tooltip message to inform users
To Reproduce

**Steps to reproduce the behavior:**

1. Go to Settings
2. Click on User Management
3. Click on the setting button.

**Screenshot**
![image](https://github.com/ColoredCow/portal/assets/121417968/6fcf00b3-e059-4807-befe-256459a8916e)

PR #3220, opened 2023-06-28
disabled the setting button for users with no entry in employee table

MohitGururani:
Targets #3184 

### Description
disabled the setting button for those users who has no entry in employee table and also added a tooltip that says "No entry in the employee table for this user" by making changes in the vue file.

### Checklist:

### DIFF
--- a/Modules/User/Entities/User.php
+++ b/Modules/User/Entities/User.php
@@ -163,7 +163,7 @@ class User extends Authenticatable
 
     public function getMonthTotalEffortAttribute()
     {
-        if (!$this->projectTeamMembers->first()) {
+        if (! $this->projectTeamMembers->first()) {
             return false;
         }
 
@@ -185,7 +185,7 @@ class User extends Authenticatable
         $fte = 0;
         $fteAmc = 0;
         foreach ($this->projectTeamMembers()->with('project')->get() as $projectTeamMember) {
-            if (!$projectTeamMember->project->is_amc) {
+            if (!  $projectTeamMember->project->is_amc) {
                 $fte += $projectTeamMember->fte;
             }
             if ($projectTeamMember->project->is_amc) {
--- a/Modules/User/Entities/User.php
+++ b/Modules/User/Entities/User.php
@@ -185,7 +185,7 @@ class User extends Authenticatable
         $fte = 0;
         $fteAmc = 0;
         foreach ($this->projectTeamMembers()->with('project')->get() as $projectTeamMember) {
-            if (!  $projectTeamMember->project->is_amc) {
+            if (! $projectTeamMember->project->is_amc) {
                 $fte += $projectTeamMember->fte;
             }
             if ($projectTeamMember->project->is_amc) {
--- a/Modules/User/Entities/User.php
+++ b/Modules/User/Entities/User.php
@@ -159,7 +159,7 @@ class User extends Authenticatable
 
     public function getMonthTotalEffortAttribute()
     {
-        if (!$this->projectTeamMembers->first()) {
+        if (! $this->projectTeamMembers->first()) {
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


Modules/User/Resources/assets/js/components/UserListing.vue
```diff
@@ -1,67 +1,83 @@
...
-                            <div class="mr-2 w-30">
-                                <img style="border-radius:50%"  class="w-full" :src="user.avatar" alt="">
-                            </div>
+			<tbody>
+				<tr v-for="(user, index) in filteredUsers" :key="index">
+					<td>
+						<span class="align-items-start d-flex justify-content-start">
+							<div class="mr-2 w-30">
+								<img style="border-radius:50%" class="w-full" :src="user.avatar" alt="">
+							</div>
```
GauravGusain98, line 20: In this whole file the indentation is too large. I think you have used indent using tabs in this file due to which the indentation size got increase. Please use indentation using spaces using below images

image
image
image

GauravGusain98 (changes requested):
Please check the comment

GauravGusain98:
The indentation is still not fixed. Try using tabs instead

GauravGusain98:
### Testing feedbacks
1. I removed this user entry from the employee table but the setting button was not disabled. PLease check it
image

Modules/User/Entities/User.php
```diff
@@ -106,6 +106,10 @@ public function getIsActiveEmployeeAttribute()
         return $this->provider == 'google' && strpos($this->email, config('constants.gsuite.client-hd')) !== false;
     }
 
+    public function getEmployeeIdAttribute()
+    {
+        return $this->employee ? true : false;
+    }
```
GauravGusain98, line 112: The function name tells that it will return employee id but you are returning true and false is does not match what the name is telling. 

Also, why we need this function. Can't we simply check if user.employee exist or not as I think we are also passing employee relationship in `with()` function when sending user data to the frontend.

Modules/User/Resources/assets/js/components/UserListing.vue
```diff
@@ -1,148 +1,152 @@
...
+                        <div>
+                            <button v-show="userPermissions['can-assign-roles']" class="btn btn-sm btn-outline-info mr-4"
+                                data-toggle="modal" data-target="#update_user_roles_modal"
+                                @click="updateUserRolesModal(index)">Manage user roles</button>
+                            <button v-show="userPermissions['can-delete'] && user.id !== authUser.id" type="button"
+                                class="btn btn-sm btn-outline-danger " data-toggle="modal" data-target="#deleteUserModal"
+                                @click="setIndex(index)">
+                            <i class="fa fa-trash fa-lg" aria-hidden="true"></i>
+                            </button>
+                            <button v-show="userPermissions['can-assign-roles']" :disabled="!user.name || !user.employeeId"
```
GauravGusain98, line 57: it is not possible that `user.name` doesnot exist therefore we don't need this check. 

Also instead of `user.employeeId` can't we simply check is a user has employee data or not. I guess we are sending employee in `with()` function from backend. So you can get the employee details using user.employee

GauravGusain98 (changes requested):

Modules/User/Resources/assets/js/components/UserListing.vue
```diff
@@ -1,129 +1,147 @@
...
-								<i class="fa fa-cog fa-lg" aria-hidden="true"></i>	
-							</button>						
-                        </div>
-                    </td>
-                </tr>
-            </tbody>
+							<button v-show="userPermissions['can-assign-roles']" :disabled="!user.employee"
+								class="btn btn-edit btn-outline-dark" data-toggle="modal"
+								data-target="#update_staff_type_modal" @click="updateUserRolesModal(index)">
+								<i class="fa fa-cog fa-lg" aria-hidden="true" data-toggle="tooltip" :title="!user.name || !user.employeeId
```
GauravGusain98, line 59:
```suggestion
								<i class="fa fa-cog fa-lg" aria-hidden="true" data-toggle="tooltip" :title="!user.employeeId
```

GauravGusain98 (changes requested):
Minor change suggested

GauravGusain98 (approved):
LGTM

Issue #3197, opened 2023-06-24
Add a dropdown menu for the status in the CodeTrek page


jaskaransingh304:
**Describe the task**

Currently, we are not getting a dropdown menu next to the status column in the applicants' table on the CodeTrek page, allowing us to filter applicants based on their status.
![image](https://github.com/ColoredCow/portal/assets/109968722/fe22a69f-c93a-48b1-beef-346f74eaea79)

**Expected behavior**
We need to add a dropdown menu next to the status column, clicking on which we can get all the options. Further, when we click on any level, we should get the same applicant list after filtering whose status is what we have chosen.

**References**

On the HR Recruitment page, there is a table called "Applications" that includes columns for "Name," "Details," "Assignee," and "Status." Next to the "Details," "Assignee," and "Status" columns, there are dropdown menus. Clicking these dropdown menus displays options for filtering the table based on the respective columns.

![image](https://github.com/ColoredCow/portal/assets/109968722/567ea7ff-3d33-4a7a-947e-a1b12aa02588)

Ayush8923:
@jaskaransingh304 Expected behavior will be something like this. Please let me know, if there are any questions and confusion.
When we choose any status in the status dropdown, we should get the same applicant list after filtering whose status is what we have chosen.

jaskaransingh304:
Thank You. I will surely ask if I face any issue.

jaskaransingh304:
Breakdown-

Code review and documentation: 0.5-1 hour.
Understanding the functionality: 2-3 hours.
Making the required changes: 2-3 hours.
Testing, code, review, and documentation: 1 hour.

Expected Time =6-8 hours.

GauravGusain98:
This has been deployed. Closing this now

PR #3230, opened 2023-06-30
#3197 Add a dropdown menu for the status in the CodeTrek page 

jaskaransingh304:
### Description
A dropdown menu has been added to the status column, allowing users to select an option from a list of available choices. When a user selects a particular option, the list of applicants is automatically filtered based on that selection.

## Approach
I reviewed the HR recruitment page and noticed that it had a dropdown menu for filtering. Then, I examined the controller, view page, and route to understand how the filter functionality was implemented. I replicated the same approach on the Codetrek applicant page's index. I created a route that would be requested, and I defined the corresponding functionality on the service page , which was returned to index by controller.

## Expected Time
- [x] Code review and documentation: 0.5-1 hour.
- [x] Understanding the functionality: 1-2 hours.
- [x] Making the required changes: 2-3 hours.
- [x] Testing, code, review, and documentation: 1 hour.

## Actual Time 
- [x] Code review and documentation: 0.5-1 hour.
- [x] Understanding the functionality: 5 hours.
Reading about routes and controllers: 5 hours.
Making the required changes: 2-3 hours.
Testing, code, review, and documentation: 1 hour.

### Total Time = 15 Hours.
---

### Checklist:


jaskaransingh304:
Attaching screenshot for reference, Before- image
After applying filter-image

Modules/CodeTrek/Resources/views/index.blade.php
```diff
@@ -146,7 +159,7 @@ class="btn-sm btn-primary mr-1 text-decoration-none"
                                 <td>
                                     <div class="d-flex justify-content-center">
                                         <span class="{{ config('codetrek.rounds.' . $applicant->latest_round_name . '.class') }} badge-pill mr-1 mb-1 fz-16">
-                                            {{ config('codetrek.rounds.' . $applicant->latest_round_name . '.label') }}
+                                            {{ config('codetrek.rounds.' . $applicant->round_name . '.label') }}
```
GauravGusain98, line 162: I think you have not taken latest pull from master branch or didn't run the migration, as the column name in database is `latest_round_name`  and not `round_name `

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -9,16 +9,21 @@
...
         $query = CodeTrekApplicant::where('status', $status)->orderBy('first_name');
-        $applicants = null;
+
         if ($centre) {
-            $applicants = $query->where('centre_id', $centre);
+            $query->where('centre_id', $centre);
+        }
+
+        if ($roundSlug) {
+            $query->where('round_name', $roundSlug);
```
GauravGusain98, line 24: correct here as well

GauravGusain98 (changes requested):

Modules/CodeTrek/Resources/views/index.blade.php
jaskaransingh304: Sure i will do that.

Modules/CodeTrek/Routes/web.php
```diff
@@ -21,4 +21,5 @@
     Route::post('/action/{applicant}', 'CodeTrekApplicantRoundDetailController@takeAction')->name('codetrek.action');
     Route::post('/update-feedback/{applicantDetail}', 'CodeTrekApplicantRoundDetailController@update')->name('codetrek.update-feedback');
     Route::post('/update-Status/{applicant}', 'CodeTrekApplicantRoundDetailController@updateStatus')->name('codetrek.updateStatus');
+    Route::get('/filter-by-round/{roundSlug}', 'CodeTrekController@index')->name('codetrek.filterByRound');
```
GauravGusain98, line 24: Why we have a new URL for filter? 

Opening same page using 2 different URL is not a good practice. Also, this way the current filters and search will not work along with filters. 

Please use the codetrek.index route only

GauravGusain98 (changes requested):

Modules/CodeTrek/Routes/web.php
jaskaransingh304: @GauravGusain98 , I have implemented the changes. Please check.

LavishRawat:
@jaskaransingh304  
**Testing Feedback**

1)The filters are not working when we apply them.

jaskaransingh304:
@LavishRawat , which filter is not working? Can you let me know?

LavishRawat:
> @LavishRawat , which filter is not working? Can you let me know?

The dropDown Filter to filter applicant according to round is not working.


Screenshot 2023-07-04 at 1 21 35 PM

jaskaransingh304:
I have tested , it is working .

LavishRawat:
> I have tested , it is working . Please run migration as a column name has been changed from round_name to latest_round_name

@jaskaransingh304 I ran the migrations still the dropdown filter is not working.

jaskaransingh304:
@LavishRawat , I'll check again

jaskaransingh304:
@LavishRawat , I have checked in another system, and it is working fine.

LavishRawat:
> @LavishRawat , I have checked in another system, and it is working fine.

Yes, the filters are working fine for Active Tab but when attempting to apply the filter for the "Inactive" tab, it automatically switches back to the "Active" tab. The same issue occurs when trying to apply the filter for the "Completed" tab.

GauravGusain98:
@jaskaransingh304 Please fix the merge conflicts and check if all the feedbacks are implemented or not

jaskaransingh304:
@GauravGusain98 , merge conflicts have been fixed , currently it is filtering for active status only as the index page is requested with active status..so I'm trying to fix it

GauravGusain98:
> @GauravGusain98 , merge conflicts have been fixed , currently it is filtering for active status only as the index page is requested with active status..so I'm trying to fix it

@jaskaransingh304 
1. I am still seeing the merge conflict 
2. Also the filter by level filter is getting removed if I am using other filters
3. The approach of creating a new route is not correct you should use the approach that is used for center filter

jaskaransingh304:
> > @GauravGusain98 , merge conflicts have been fixed , currently it is filtering for active status only as the index page is requested with active status..so I'm trying to fix it
> 
> @jaskaransingh304
> 
> 1. I am still seeing the merge conflict
> 2. Also the filter by level filter is getting removed if I am using other filters
> 3. The approach of creating a new route is not correct you should use the approach that is used for center filter

1. Merge conflicts fixed.

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -14,12 +14,16 @@ public function getCodeTrekApplicants($data = [])
...
+        $roundSlug = $data['roundSlug'] ?? null;
         $query = CodeTrekApplicant::where('status', $status);
         $applicants = null;
 
         if ($centre) {
             $applicants = $query->where('centre_id', $centre);
         }
+        if ($roundSlug) {
+            $query->where('latest_round_name', $roundSlug);
+        }
```
GauravGusain98, line 26: This should also be added below for calculating count query for each status

GauravGusain98 (changes requested):

Modules/CodeTrek/Services/CodeTrekService.php
jaskaransingh304: > This should also be added below for calculating the count query for each status

@GauravGusain98 I have added done, the only issue is the filter works when it is selected after selecting other filters. If I select round as level 1 and status inactive, then the roundSlug is not getting passed to the status parameter, I have tried doing that but as we don't have a fixed value for $roundSlug, it keeps changing. So can you please help with the other procedure? In case of any question, we can connect and discuss.

GauravGusain98: The filter is getting removed when I am using other filters

GauravGusain98: maybe you are not passing the value of level filter when the other filters form is getting submitted

jaskaransingh304: @GauravGusain98 , it is not working only when selected before other filter.. If round filter is selected after it..It is working...So in case to make it work for other filter , for example status - I would have to pass the value in status filter for $roundSlug...but the value of $roundSlug may change when it is selected...so that's what i was asking about...in this case i need to pass value in every filter for roundSlug

GauravGusain98: yes you need to pass the value of roundSlug in every filter

jaskaransingh304: > yes you need to pass the value of roundSlug in every filter

@GauravGusain98 , It is passed to every filter.

GauravGusain98 (approved):
LGTM

PR #3241, opened 2023-07-04
changes made for if user is null

MohitGururani:
### Description

Here's a summary of the update:
1.File: https://github.com/ColoredCow/portal/blob/master/resources/views/hr/employees/show.blade.php
Changes:
Added an @if condition to check if $employee->user is null
If it's not null, a hidden input field with the corresponding value is added.

How this PR helps:
Adding this condition helps to tackle the error  "Trying to get property 'user' of non-object". if `$employee->user` is null,this will give an error because user is null and we can't be able to access them.

2.File:https://github.com/ColoredCow/portal/blob/master/Modules/Project/Resources/views/index.blade.php
Changes:
Added an @if condition to check if $employee property exists. If it does, the condition evaluates to true.
 If it doesn't exist or is null, the condition evaluates to false, and the code within the @if block will be skipped.
 
 How this PR helps:
 Adding this condition helps in tackle the error  "Trying to get property 'employee' of non-object" .If the condition is false (i.e., $teamMember->user is null or doesn't have an employee property)

### DIFF
--- a/Modules/Project/Resources/views/index.blade.php
+++ b/Modules/Project/Resources/views/index.blade.php
@@ -94,7 +94,7 @@
                                     <div class="d-flex justify-content-between">
                                         <div>
                                             {{ $client->name }}
-                                            {{ sprintf("(%03s)",$client->client_id) }}
+                                            {{ sprintf('(%03s)', $client->client_id) }}
                                         </div>
                                         <div class="">
                                             {{ __('Total Hours Booked: ') . $client->current_hours_in_projects }}</div>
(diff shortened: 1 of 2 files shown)


Ayush8923:
@MohitGururani Please add the issue link and the description in the PR description. and then assign the PR for review.

MohitGururani:
@Ayush8923 this is self assigned. As this problem occurs when the user is not in the employee table so it gives null value.To deal with it I added some if check.

Issue #3215, opened 2023-06-27
Incorrect Tab Count when Applying All Centres Filter


LavishRawat:
**Describe the task**
The task involves addressing the issue related to the count of codetrek applicants in the codetrek module. Currently, the count is displayed correctly in the ` active `, ` inactive `, and ` completed ` tabs. However, when `All Centres` filter is applied, the count does not appear to be accurate.

**Expected behavior**
When the `All Centres` filter is applied in the codetrek module, the system should correctly filter out the list of users based on the selected `Centre`. However, there is an issue with the count of applicants not being displayed accurately.
This means that the count should dynamically adjust based on the filtered results, providing an accurate representation of the number of applicants from the specific `Centre`.

To Reproduce

1. Go to Codetrek.
2. Applt the Centres filter.
3. You could see the Count mismatch.

<img width="1410" alt="Screenshot 2023-06-27 at 6 38 24 PM" src="https://github.com/ColoredCow/portal/assets/108713935/ac0ee0ec-e4fe-4edc-bc07-80a1dce23a06">

LavishRawat:
### Action items

**Basics Concept - 1 pointer:**
- [ ] Read about Models in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Views in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Routes in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Controller in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read about Services files in Laravel. Share your understanding of why we use them in the task comment in your own language
- [ ] Read how we can print a value in Laravel using dd()
- [ ] Try printing `hello world` using the `dd` function


**Form Submission - 1 pointer:**
- [ ] Read about HTML forms and how we submit forms. Also, share the use of the below provide items
- [ ] Action attribute in the form
- [ ] Method attribute in the form and different types of Methods
- [ ] Name attribute in any HTML tag and its importance while form submission

AnirudhAP2k:
Laravel is a MVC based framework. here :-
1. M stands for 'Model'. It is used in backend to provide data or requested data to controller.
2. V stands for 'View'. It is used in frontend. It is presented in a form of blade template. It is an HTML that we can see in the browser.
3. C stands for 'Controller'. It is used as an interface or medium between view and model. The view requests the data from controller which in turn access the model methods using Get() method.
[...]

PR #3247, opened 2023-07-06
tab count correction in codetrek module  using centre filter

AnirudhAP2k:
Target #3215 
### Description
I updated the data coming to $applicantCountData variable according to status and then assigned the codeTrek applicant's data to a $applicantsCountData variable by passing a conditional statement based on the value of center filter in $center variable.

### Feedback Cycle
Cycle Count- 4

### Checklist:


LavishRawat:
@AnirudhAP2k 
**Testing Feedbacks**

1) The filter only applies to the selected tab, while the other tab shows the total count, but as soon as we switch to the other Tab it gives the correct count for the applied filter while the previous tab shows the total count.

   image
   
after switching to the `Inactive Tab`, the count changes to 2 and the active tab then shows the total count of applicants.

image

GauravGusain98 (changes requested):

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -46,14 +44,25 @@ public function getCodeTrekApplicants($data = [])
...
 
+        $applicant = CodeTrekApplicant::all();
+
+        if ($centre) {
+            $statusCounts = [
+                'active' => $applicant->where('centre_id', $centre)->where('status', 'active')->count(),
+                'inactive' => $applicant->where('centre_id', $centre)->where('status', 'inactive')->count(),
+                'completed' => $applicant->where('centre_id', $centre)->where('status', 'completed')->count(),
+            ];
+        }
```
GauravGusain98, line 60: For count we already have written the logic in line 34 where the count data is being stored in $applicantsData variable. You don't need to again write query to get the count. 

In line 34 logic currently, its only filtering based on search and not on center so please update the logic such that it also filter based on center

GauravGusain98 (changes requested):

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -33,27 +31,34 @@ public function getCodeTrekApplicants($data = [])
...
+        $query = CodeTrekApplicant::whereIn('status', ['active', 'inactive', 'completed'])
+            ->with('mentor')
+            ->when($search, function ($query) use ($search) {
+                return $query->whereRaw("CONCAT(first_name, ' ', last_name) LIKE '%$search%'");
+            })
+            ->groupBy('status')
+            ->selectRaw('count(status) as total, status');
+
+        if ($centre == null) {
+            $applicantsData = $query->get();
```
Ayush8923, line 43: can you please let me know, what's the use of this line?

Ayush8923 (dismissed):
Added some comments. Please check.

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -33,27 +31,34 @@ public function getCodeTrekApplicants($data = [])
...
             ->appends(request()->except('page'));
 
-        $applicantsData = CodeTrekApplicant::whereIn('status', ['active', 'inactive', 'completed'])->with('mentor')
-        ->when($search, function ($query) use ($search) {
-            return $query->whereRaw("CONCAT(first_name, ' ', last_name) LIKE '%$search%'");
-        })
-        ->groupBy('status')
-        ->selectRaw('count(status) as total, status')
-        ->get();
+        $query = CodeTrekApplicant::whereIn('status', ['active', 'inactive', 'completed'])
```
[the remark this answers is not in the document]
AnirudhAP2k: Sir, this query variable was already defined in master branch before this task was given. I only altered the data passing to query variable in line 34.
And also query is performing same function as a database query of storing all the data of applicants.

Modules/CodeTrek/Services/CodeTrekService.php
AnirudhAP2k: In this statement, it is inside an 'IF' conditional statement such that if we put "ALL CENTRE" filter, the value assigned to $centre variable will be of 'null', so this condition will pass and it will assign all the codeTrek Applicant's data to to $appliantData variable from $query, which acts as a database query for Applicants having all data.

Modules/CodeTrek/Services/CodeTrekService.php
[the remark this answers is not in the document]
GauravGusain98: No problem @AnirudhAP2k. It's always good to correct the old code as well. Just give the variable a good meaningful name

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -33,27 +31,34 @@ public function getCodeTrekApplicants($data = [])
...
+            ->when($search, function ($applicantCountData) use ($search) {
+                return $applicantCountData->whereRaw("CONCAT(first_name, ' ', last_name) LIKE '%$search%'");
+            })
+            ->groupBy('status')
+            ->selectRaw('count(status) as total, status');
+
+        if ($centre) {
+            $applicantsData = $applicantCountData->where('centre_id', $centre)->get();
+        } else {
+            $applicantsData = $applicantCountData->get();
```
GauravGusain98, line 45: Why you changed the variable name here. It should be same `applicantCountData`

GauravGusain98 (changes requested):

Modules/CodeTrek/Services/CodeTrekService.php
AnirudhAP2k: Sir, applicantData is a different variable performing different function of storing filtered data on the basis of centre and status coming from $applicantCountData. 
Here applicantData and applicantCountData are two different variables performing different functions.

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -33,27 +31,34 @@ public function getCodeTrekApplicants($data = [])
...
+
+        foreach ($applicantCountData as $data) {
             $statusCounts[$data->status] = $data->total;
         }
 
         return [
             'applicants' => $applicants,
-            'applicantsData' => $applicantsData,
-            'statusCounts' => $statusCounts
+            'applicantsData' => $applicantCountData,
```
GauravGusain98, line 60: can you check if we are using this `applicantsData` in frontend or not. Because I think we might be only using `statusCounts` variable there as it is derived from the `applicantsData` only

GauravGusain98: In case the `applicantsData` is not being used in frontend, please remove it from the return statement else its fine

AnirudhAP2k: Okay done sir.

Modules/CodeTrek/Services/CodeTrekService.php
```diff
@@ -5,8 +5,6 @@
 use Modules\CodeTrek\Entities\CodeTrekApplicant;
 use Modules\CodeTrek\Entities\CodeTrekCandidateFeedback;
 use Modules\CodeTrek\Entities\CodeTrekApplicantRoundDetail;
-use Illuminate\Support\Facades\Mail;
-use  Modules\CodeTrek\Emails\CodetrekMailApplicant;
```
Ayush8923, line 9: I don't think, we need to remove these 2 lines. Commented out the mail functionality for now. But we need this codetrek mail functionality feature in the future. So let me know, is there any specific reason you removed these lines.

Ayush8923 (commented):
Added one question, else looks good to me 👍.

Modules/CodeTrek/Services/CodeTrekService.php
AnirudhAP2k: I removed those because it was showing error during CI checks, as classes  were unused during execution.

GauravGusain98 (approved):
LGTM

PR #3249, opened 2023-07-06
Show only current employees in an organization

Bahugunajii:
### Description
- Show only current employees of the organization in the dropdown.

### Checklist:

### DIFF
--- a/Modules/HR/Http/Controllers/EmployeeController.php
+++ b/Modules/HR/Http/Controllers/EmployeeController.php
@@ -102,7 +102,9 @@ class EmployeeController extends Controller
         $assessments = Assessment::where('reviewee_id', $employee->id)
             ->orderBy('created_at', 'desc')
             ->get();
-        $employees = Employee::all();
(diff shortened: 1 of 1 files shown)


Modules/HR/Http/Controllers/EmployeeController.php
```diff
@@ -102,7 +102,9 @@ public function reviewDetails(Employee $employee)
         $assessments = Assessment::where('reviewee_id', $employee->id)
             ->orderBy('created_at', 'desc')
             ->get();
-        $employees = Employee::all();
+        $employees = Employee::whereHas('user', function ($query) {
+            $query->whereNull('deleted_at');
+        })->get();
```
Ayush8923, line 107: I think, can we write like this?
```suggestion
        $employees = Employee::whereNull('deleted_at')->get();
```

Ayush8923 (changes requested):

Ayush8923 (approved):

PR #3344, opened 2023-08-19
Fix/sync master and release

GauravGusain98:
### Changes 

- Sync master and release


Modules/Report/Services/CodeTrek/ReportDataService.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Modules\Report\Services\CodeTrek;
+
+use Modules\CodeTrek\Entities\CodeTrekApplicant;
+
+class ReportDataService
+{
+    public function getDataForDailyCodeTrekApplications($type, $filters)
+    {
+        if ($type == 'codetrek-application') {
```
rathorevaibhav, line 11: shouldn't this have a default return type also? As the controller is returning the return value from this function

Modules/Report/Services/Finance/ReportDataService.php
```diff
@@ -62,8 +37,8 @@ private function revenueTrendForClient($filters)
 
         $filters['start_date'] = empty($filters['start_date']) ? $defaultStartDate : $filters['start_date'];
         $filters['end_date'] = empty($filters['end_date']) ? $defaultEndDate : $filters['end_date'];
-
-        $reportData = $this->service->getRevenueReportDataForClient($filters, $client);
+        $revenueReportService = new RevenueReportService;
+        $reportData = $revenueReportService->getRevenueReportDataForClient($filters, $client);
```
rathorevaibhav, line 41: I think it would be easier if we make these static functions and then can directly call these like:
```php
$reportData = RevenueReportService::getRevenueReportDataForClient($filters, $client);
```

Modules/Report/Services/Finance/ReportDataService.php
```diff
@@ -85,8 +60,8 @@ private function revenueTrend($filters)
             'previous_period_end_date' => $defaultPreviousEndDate,
         ];
         $filters = array_merge($defaultFilters, request()->all());
-
-        $reportData = $this->service->getRevenueGroupedByClient($filters);
+        $revenueReportService = new RevenueReportService;
+        $reportData = $revenueReportService->getRevenueGroupedByClient($filters);
```
rathorevaibhav, line 64: same feedback for this. Can be used as a static function.

GauravGusain98:
image

rathorevaibhav:
@GauravGusain98 in your error screenshot, have you defined the function as static before calling it statically?

GauravGusain98:
> @GauravGusain98 in your error screenshot, have you defined the function as static before calling it statically?

If I define it as static then I am getting this error
image

rathorevaibhav:
@GauravGusain98 static functions are called like this:
```php
RevenueReportService::getRevenueGroupedByClient($filters);

// and not like
$this->getRevenueGroupedByClient($filters);
```

rathorevaibhav (approved):

PR #3355, opened 2023-09-06
[Fix]:- HR module recruitment page performance

Ayush8923:
**Description**
- Optimize the HR recruitment index function code and remove the unnecessary code.
- In this, we need to test this in the UAT env. I checked this in the latest database dump of uat in local it was working fine.  Initially, it was taking time 14 seconds to load the page.

**What do we need to do next**
- We need to clean up more code in this. Because there is some same code in the application service and controller, but we are not using this service right now, all the code is added in the controller.

### DIFF
--- a/Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
+++ b/Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
@@ -86,20 +86,20 @@ abstract class ApplicationController extends Controller
             'roundFilters' => request()->get('roundFilters'),
         ];
         $loggedInUserId = auth()->id();
-        $applications = Application::join('hr_application_round', function ($join) {
+        $applications = Application::select('hr_applications.*')
+        ->join('hr_application_round', function ($join) {
             $join->on('hr_application_round.hr_application_id', '=', 'hr_applications.id')
                 ->where('hr_application_round.is_latest', true);
-        })->with(['applicant', 'job', 'tags', 'latestApplicationRound']);
-
-        $applications = $applications->whereHas('latestApplicationRound')
-            ->applyFilter($filters)
-            ->orderByRaw("FIELD(hr_application_round.scheduled_person_id, {$loggedInUserId} ) DESC")
-            ->orderByRaw('ISNULL(hr_application_round.scheduled_date) ASC')
-            ->orderByRaw('hr_application_round.scheduled_date ASC')
-            ->select('hr_applications.*')
-            ->latest()
-            ->paginate(config('constants.pagination_size'))
-            ->appends(request()->except('page'));
+        })
+        ->with(['applicant', 'job', 'tags', 'latestApplicationRound'])
+        ->whereHas('latestApplicationRound')
+        ->applyFilter($filters)
+        ->orderByRaw("FIELD(hr_application_round.scheduled_person_id, {$loggedInUserId} ) DESC")
+        ->orderByRaw('ISNULL(hr_application_round.scheduled_date) ASC')
+        ->orderByRaw('hr_application_round.scheduled_date ASC')
+        ->latest()
+        ->paginate(config('constants.pagination_size'))
+        ->appends(request()->except('page'));
         $countFilters = array_except($filters, ['status', 'round']);
         $attr = [
             'applications' => $applications,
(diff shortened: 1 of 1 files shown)


Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -136,28 +137,39 @@ public function index()
...
-        $attr['tags'] = Tag::orderBy('name')->get();
+        $attr['jobs'] = Cache::remember('jobs', now()->addMinutes(60), function () {
+            return Job::all();
+        });
+        $attr['universities'] = Cache::remember('universities', now()->addMinutes(60), function () {
+            return University::all();
+        });
+        $attr['tags'] = Cache::remember('tags', now()->addMinutes(60), function () {
+            return Tag::orderBy('name')->get();
+        });
```
Ayush8923, line 148: Before, When this function is rendered then every time it's fetched the data from the database. Now, here we are using the Laravel Caching mechanism, If this data is in the cache memory then it will fetch the data from the cache. Every data has its own unique identity and its cache expiration time is 60 minutes. So here we do not need to hit the query in the database all the time.

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -136,28 +137,39 @@ public function index()
...
         })->orderby('name', 'asc')->get();
 
-        $attr['openApplicationsCountForJobs'] = [];
+        $openApplicationCountForJob = Application::whereIn('hr_job_id', $applications->pluck('hr_job_id'))
+            ->isOpen()
+            ->selectRaw('hr_job_id, COUNT(*) as count')
+            ->groupBy('hr_job_id')
+            ->get()
+            ->keyBy('hr_job_id')
+            ->toArray();
```
Ayush8923, line 161: Before, every time the loop ran, we would hit the database with a query, which could cause some performance issues. But now we are hitting the query in the database only once and fetching the data we need.

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -136,28 +137,39 @@ public function index()
...
 
         $attr['applicantId'] = [];
-        $applications = Application::get('hr_applicant_id');
-        foreach ($applications as $application) {
-            $applicantID = ApplicantMeta::where('hr_applicant_id', $application->hr_applicant_id)->first();
-            if ($applicantID) {
-                $attr['applicantId'] = $applicantID;
-            }
+        $applications = Application::pluck('hr_applicant_id');
+        $applicantData = ApplicantMeta::whereIn('hr_applicant_id', $applications)->get();
```
Ayush8923, line 169: Earlier we used to hit query in databse corresponding to each applicant id in each loop. Due to which there could be some problem in performance and time complexity also increased. But now we are passing collection of applicant id directly and hitting query in the database only once.

GauravGusain98 (approved):
LGTM

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -136,28 +137,39 @@ public function index()
                 ->count();
         }
 
-        $attr['jobs'] = Job::all();
-        $attr['universities'] = University::all();
-        $attr['tags'] = Tag::orderBy('name')->get();
+        $attr['jobs'] = Cache::remember('jobs', now()->addMinutes(60), function () {
```
pankaj-ag, line 140: @Ayush8923 This will restrict the system updates for one hour. So let's say you add a tag now and it will show up after 1 hour which does not make sense to me. 

Also this implementation is good but I feel that there are lot of scope to fix the existing code and then apply caching for these things. If we apply caching right now it will just suppress the error and time. 

So for now remove this implementation and try to fix all the queries we are making to DB and just fix them. 








This seems like a fine approach but we will use it as a last thing.

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
pankaj-ag: This is good. 👍🏼

pankaj-ag (changes requested):

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
Ayush8923: Make sense 👍

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
```diff
@@ -85,20 +85,20 @@ public function index()
...
+        })
+        ->with(['applicant', 'job', 'tags', 'latestApplicationRound'])
+        ->whereHas('latestApplicationRound')
+        ->applyFilter($filters)
+        ->orderByRaw("FIELD(hr_application_round.scheduled_person_id, {$loggedInUserId} ) DESC")
+        ->orderByRaw('ISNULL(hr_application_round.scheduled_date) ASC')
+        ->orderByRaw('hr_application_round.scheduled_date ASC')
+        ->latest()
+        ->paginate(config('constants.pagination_size'))
+        ->appends(request()->except('page'));
```
Ayush8923, line 101: Before, we hit the different queries in the database, which could cause some performance issues and the time complexity increase. now, here we combined both queries and hit the query only once time in the database.

Ayush8923:
@pankaj-ag I'm still working on some more cleanup and will make those changes in another PR.

Modules/HR/Http/Controllers/Recruitment/ApplicationController.php
pankaj-ag: It's bit tough to check how it's going to impact. So please test it carefully

pankaj-ag (approved):

PR #3361, opened 2023-09-13
[Fix]:- Optimization HR Module & Code Clenaup

Ayush8923:
**Description**
- The `ApplicationController` and the `ApplicationService` from which the Recruitment data is rendered, there was a little bit of the same code in both these files, but we used only the `ApplicationController` to fetch the data. In this, Code duplication was happening on this. We have moved all the changes in the `ApplicationController` to the `ApplicationService` file which will be totally used to fetch the data.
- Also, Made some code cleanups to optimize the HR recruitment page.


pankaj-ag (approved):
On a high level it looks good but please test the code properly in case we miss some edge case

Issue #3613, opened 2024-05-01
Configurable "Deadline Alert" duration for Project Contract End Date.


sameer-bot1:
**Problem Statement:**
The "deadline alert" for the project contract's end date is set to 30 days in the code itself. it should be configurable within the portal.

**Proposed Solution-**

- At the moment, we're not taking any "deadline alert" duration.
- We're currently configuring it within the portal's Setting using a different field, like 'employee earning threshold'. We can allow input for configuring the deadline alert for the project contract's end date.
- we can use Setting table to store "deadline alert" duration in variable.

PR #3614, opened 2024-05-01
Project contract end date alert config

sameer-bot1:
Targets #3613

## Problem statement

The "deadline alert" for the project contract's end date is set to 30 days in the code itself. it should be configurable within the portal.


app/Http/Controllers/Settings/ConfigurationController.php
sameer-bot1, line 1: This file used to be named ProfitabilityController.php. I renamed it to ConfigurationController because now it handles configuring multiple values instead.

app/Services/ConfigurationService.php
sameer-bot1, line 1: I created ConfigurationService  here created a method retrieves settings from the database related to employee earning thresholds and contract end date thresholds and updates an existing record

Modules/Report/Http/Controllers/ContractReportController.php
```diff
@@ -18,9 +19,11 @@ public function __construct(ContractReportService $service)
     public function index()
     {
         $clientsData = $this->service->getAllClientsData();
+        $alertDiff = Setting::where('setting_key', 'contract_endDate_threshold')->value('setting_value');
```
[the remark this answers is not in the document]
pokhiii: What's the idea behind choosing the name `$alertDiff`? Why not simply use: `$contractEndDateThreshold`

app/Services/ConfigurationService.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace App\Services;
+
+use App\Models\Setting;
+
+class ConfigurationService
+{
+    public function index()
+    {
+        $ConfigsValue = Setting::query()
```
pokhiii, line 11: naming convention

app/Services/ConfigurationService.php
pokhiii, line 1: Any particular reason for calling it ConfigurationService and not SettingService?

resources/views/settings/configuration-threshold/index.blade.php
sameer-bot1, line 1: Previously, the location of index.blade.php was "resources/views/settings/employee-earning-threshold/index.blade.php". I've now changed it to "resources/views/settings/configuration-threshold/index.blade.php"

app/Services/ConfigurationService.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace App\Services;
+
+use App\Models\Setting;
+
+class ConfigurationService
+{
+    public function index()
```
pokhiii, line 9: The method `index` make more sense in the context of a controller. For a setting it can be more specific like `getList`, etc. depending on what you're trying to do.

app/Services/ConfigurationService.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+        $ConfigsValue = Setting::query()
+            ->where('setting_key', 'employee_earning_threshold')
+            ->orWhere('setting_key', 'contract_endDate_threshold')
```
pokhiii, line 13: What special about these two settings: `employee_earning_threshold` and `contract_endDate_threshold` that you have to hardcode it here?

Ideally, the service should be generic enough to be re-used and provide common operations that can be done on a particular component/resource of the system.


It is not clear to me why the `index` method is return employee earning and contract end date threshold

routes/web.php
```diff
@@ -99,8 +99,8 @@
 
     Route::prefix('settings')->namespace('Settings')->group(function () {
         Route::get('/', 'SettingController@index')->name('settings.index');
-        Route::get('/employee-earning-threshold', 'ProfitabilityController@index')->name('settings.employee-earning-threshold');
-        Route::post('/employee-earning-threshold', 'ProfitabilityController@update')->name('settings.update-employee-earning-threshold');
+        Route::get('/configuration-threshold', 'ConfigurationController@index')->name('settings.configuration-threshold');
+        Route::post('/configuration-threshold', 'ConfigurationController@update')->name('settings.update-configuration-threshold');
```
pokhiii, line 103: If we go by this pattern, we will have to introduce a new route everytime a new setting is introduced. So routes will start growing as the rows in the settings table increase. That's not very scalable approach.

You should think of how setting routes can be used to get retrieve those values.

pokhiii (changes requested):

Modules/Report/Http/Controllers/ContractReportController.php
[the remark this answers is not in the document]
sameer-bot1: > Also, what's the convention behind choosing the setting key `contract_endDate_threshold` it's a combination of snake case and cameCase.
> 
> `contract_end_date_threshold` is what i think we should have

Ok, Sure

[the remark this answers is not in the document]
sameer-bot1: > What's the idea behind choosing the name `$alertDiff`? Why not simply use: `$contractEndDateThreshold`

there is no idea behind it, I am changing it to $contractEndDateThreshold

app/Services/ConfigurationService.php
sameer-bot1: > Any particular reason for calling it ConfigurationService and not SettingService?

in Setting we have different options like bank Detail, Salary, Country Detail, Config Variable. we are using Service for Config Variable.

app/Services/ConfigurationService.php
sameer-bot1: ok, sure

> The method `index` make more sense in the context of a controller. For a service it can be more specific like `getList`, etc. depending on what you're trying to do.

ok, sure

app/Services/ConfigurationService.php
sameer-bot1: yes, i renamed as **ThresholdValues**

routes/web.php
sameer-bot1: no, We just need two routes. One to display the variable and the other to update all the variables.

app/Services/ConfigurationService.php
sameer-bot1: Setting contains various data from different modules, but not all of it is needed for the configurationService. We only need specific data from the Setting, like `employee_earning_threshold` and `contract_endDate_threshold`..
And I renamed index method as getThresholdValue.

Modules/Report/Http/Controllers/ContractReportController.php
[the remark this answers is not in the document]
sameer-bot1: > Don't we have a `SettingService`? Something like would be really nice:

Right now, we don't have SettingService.

app/Services/ConfigurationService.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace App\Services;
+
+use App\Models\Setting;
+
+class ConfigurationService
+{
+    public function getThresholdValue()
```
pokhiii, line 9:
```php
$setting_keys = [
    'employee_earning_threshold',
    'contract_end_date_threshold',
]
```

```suggestion
    public function getSettings($settingKeys: array)
```

app/Services/ConfigurationService.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+        $ThresholdValues = Setting::query()
+            ->where('setting_key', 'employee_earning_threshold')
+            ->orWhere('setting_key', 'contract_end_date_threshold')
```
pokhiii, line 13: We can use [`whereIn`](https://laravel.com/docs/11.x/queries#additional-where-clauses) here
```suggestion
            ->whereIn('setting_key', $settingKeys)
```

app/Services/ConfigurationService.php
```diff
@@ -0,0 +1,27 @@
...
+            ->get();
+
+        return $ThresholdValues;
```
pokhiii, line 16:
```suggestion
        return $settings;
```

app/Http/Controllers/Settings/ConfigurationController.php
```diff
@@ -0,0 +1,36 @@
...
+    use AuthorizesRequests;
+    protected $service;
+
+    public function __construct(ConfigurationService $service)
+    {
+        $this->service = $service;
+    }
+    public function index()
+    {
+        $ThresholdValues = $this->service->getThresholdValue();
```
pokhiii, line 21: 1. Replace `getThresholdValue` with a more generic `getSettings` method
2. We need to provide the list of settings the controller wants to fetch from the service. For that, the controller will read the configuration values. The configurations for settings can be defined module-wise in `config/constants.php`


```php
[
    // ...
    'module-settings' => [
        'config-variable' => [
            'employee_earning_threshold',
            'contract_end_date_threshold',
        ]
    ],
]
```

3. The values can be read using the `config` helper: https://laravel.com/docs/11.x/configuration#accessing-configuration-values

app/Services/ConfigurationService.php
sameer-bot1: > 

ok, I changed it to `$setting`

app/Services/ConfigurationService.php
sameer-bot1: > ```
> $setting_keys = [
>     'employee_earning_threshold',
>     'contract_end_date_threshold',
> ]
> ```

Yes I changed it

app/Services/ConfigurationService.php
sameer-bot1: > We can use [`whereIn`](https://laravel.com/docs/11.x/queries#additional-where-clauses) here

Ok, Sure

app/Http/Controllers/Settings/ConfigurationController.php
```diff
@@ -0,0 +1,37 @@
...
+
+use App\Http\Controllers\Controller;
+use App\Services\SettingService;
+use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
+use Illuminate\Http\Request;
+
+class ConfigurationController extends Controller
+{
+    use AuthorizesRequests;
+    protected $service;
```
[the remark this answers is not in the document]
sameer-bot1: > leave a line above

Ok

app/Http/Controllers/Settings/ConfigurationController.php
```diff
@@ -0,0 +1,37 @@
...
+    public function __construct(SettingService $service)
+    {
+        $this->service = $service;
+    }
+    public function index()
+    {
+        $settingKeys = config('constants.module-settings.config-variable');
+        $getSettings = $this->service->getThresholdValue($settingKeys);
+
+        return view('settings.configuration-threshold.index', [
```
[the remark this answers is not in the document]
sameer-bot1: > Can we keep the name consistent everywhere:

Ok, sure

app/Http/Controllers/Settings/ConfigurationController.php
```diff
@@ -0,0 +1,37 @@
...
+            'getSettings' => $getSettings,
+        ]);
+    }
+    public function update(Request $request)
+    {
+        $this->service->updateSetting([
```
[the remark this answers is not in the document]
sameer-bot1: > Since we can save/update multiple settings

Ok, sure

app/Services/SettingService.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    public function getThresholdValue($setting_keys)
+    {
+        $getSettings = Setting::query()
+            ->whereIn('setting_key', $setting_keys)
+            ->get();
+
+        return $getSettings;
+    }
+    public function updateSetting($settings)
```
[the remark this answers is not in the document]
sameer-bot1: sure

config/constants.php
```diff
@@ -315,4 +315,9 @@
             'class' => 'badge badge-success',
         ],
     ],
+    'module-settings' => [
+        'config-variable' => [
+            'employee_earning_threshold', 'contract_end_date_threshold',
```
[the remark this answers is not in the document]
sameer-bot1: ok, sure

app/Http/Controllers/Settings/ConfigurationController.php
```diff
@@ -0,0 +1,37 @@
...
+    public function index()
+    {
+        $settingKeys = config('constants.module-settings.config-variable');
+        $getSettings = $this->service->getThresholdValue($settingKeys);
+
+        return view('settings.configuration-threshold.index', [
+            'getSettings' => $getSettings,
+        ]);
+    }
+    public function update(Request $request)
```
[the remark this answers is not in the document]
sameer-bot1: sure

routes/web.php
```diff
@@ -99,8 +99,8 @@
 
     Route::prefix('settings')->namespace('Settings')->group(function () {
         Route::get('/', 'SettingController@index')->name('settings.index');
-        Route::get('/employee-earning-threshold', 'ProfitabilityController@index')->name('settings.employee-earning-threshold');
-        Route::post('/employee-earning-threshold', 'ProfitabilityController@update')->name('settings.update-employee-earning-threshold');
+        Route::get('/config-variables', 'ConfigurationController@index')->name('settings.config-variables');
+        Route::post('/config-variables', 'ConfigurationController@update')->name('settings.update-config-variables');
```
pokhiii, line 103:
```suggestion
        Route::resource('config-variables', 'ConfigurationController')->only(['index', 'update']);
```

// config-variables.index
// config-variables.update

pokhiii (approved):

PR #3656, opened 2024-06-07
Feature/cron-job-for-appplication-rejection

srivastava-jyoti:
Targets: In this pr I have worked if a application is opened for more than 10 days and is in introductory call phase and has not scheduled call they will be rejected and mail will be triggered for rejection. And once they reapplied their application gets opened and a re-applied tag will be enabled.


app/Console/Kernel.php
```diff
@@ -46,12 +48,13 @@ protected function schedule(Schedule $schedule)
     {
         // $schedule->command('application:no-show')->dailyAt('21:00');
         $schedule->command('application:send-interview-reminders')->dailyAt('08:00');
-        $schedule->command('sync:effortsheet')->weekdays()->timezone(config('constants.timezone.indian'))->hourlyAt(25);
+        $schedule->command('sync:effortsheet')->timezone(config('constants.timezone.indian'))->hourlyAt(25);
```
srivastava-jyoti, line 51: Syncing effort sheet everyday instead of on weekdays

PR #3667, opened 2024-06-21
Optimize invoice edit page

GauravGusain98:
### Changes
1. Remove unwanted variable that was being passed to invoice edit page. It was being used in old file when we have submodules but later the code have been removed where this variable was being used but somehow the variable are missed in the cleanups. Due to this code a huge number of queries were being made by the system resulting in slowness in page loading
2. Deleted the old legacy file which is not being used now


app/Http/Controllers/Finance/InvoiceController.php
```diff
@@ -86,17 +86,11 @@ public function store(InvoiceRequest $request)
...
     public function edit(Invoice $invoice)
     {
-        $clients = Client::getInvoicableClients(collect($invoice->projectStageBillings)->pluck('id')->toArray());
-        $invoice->load('projectStageBillings', 'projectStageBillings.projectStage', 'projectStageBillings.projectStage.project');
-
-        return view('finance.invoice.edit')->with([
-            'invoice' => $invoice,
-            'clients' => $clients,
-        ]);
+        // Show the form for editing the specified resource.
```
GauravGusain98, line 93: Old controller function. Not needed now

P4NK4J (approved):

PR #3671, opened 2024-06-27
Fix - employee entry not getting created when user sign in to portal

GauravGusain98:
### Problem 
In portal we have 2 tables `users` and `employees`, When someone log into portal using their Gsuite account, entry gets created in these 2 tables. But recently 2 cases occurred where the entry gets created in user table but not in the employee table.

### Diagnosis
- We found that earlier there was an userObserver file which is used to create entry in Employee table whenever a user entry gets created. 
- This is the PR in which this file got deleted 
https://github.com/ColoredCow/portal/pull/3315


GauravGusain98:
@pankaj-ag Can you please check the code changes since I am not sure why this file was deleted earlier.

Satendra-SR (approved):
Looks good

PR #3693, opened 2024-08-01
[Feat]:- Upgrade the PHP Version

Ayush8923:
## Description of the changes

- Upgrade the PHP version `7.4` to `8.2`.

## Checklist:

- [x] I have added comments on my code changes where required.


Ayush8923:
The coding standards check is failing because we have upgraded the PHP version. We will also need to make some changes in the code, such as updating the way parameters are passed in functions and parameter declaration.

PR #3740, opened 2024-10-28
[Hotfix]:- Remove the Mandatory Field

Ayush8923:
### Changes
1. Make the nullable field in the `Prospects` table.
2. Update the `ActiveUsers` function so that the function is return all the active users with sorting.
3. Remove the mandatory validation field from the `Request`.
4. If the data is not present then show the `N/A` and in the `dashboard` show the `-` instead of empty.

## Summary by CodeRabbit

- **New Features**
	- Enhanced user interface with improved handling of null values in prospect details.
	- Conditional toast notifications for user feedback on actions.
  
- **Bug Fixes**
	- Simplified validation rules, reducing required fields for prospect creation and editing.
	- Improved robustness in data handling, ensuring missing fields are set to null.

- **Documentation**
	- Updated error messages for clarity in validation feedback.

- **Style**
	- Removed visual indicators for mandatory fields in forms.

### DIFF
--- a/Modules/User/Entities/User.php
+++ b/Modules/User/Entities/User.php
@@ -240,9 +240,8 @@ class User extends Authenticatable
 
     public function getActiveUsersAttribute()
     {
-        $users = self::all()->sortBy('name');
+        $users = self::query()->orderBy('name')->get();
         $activeUsers = [];
-
         foreach ($users as $user) {
             if (! $user->isActiveEmployee) {
                 continue;
--- a/Modules/User/Entities/User.php
+++ b/Modules/User/Entities/User.php
@@ -241,15 +241,20 @@ class User extends Authenticatable
     public function getActiveUsersAttribute()
     {
         $users = self::all();
-        $activeUser = [];
+        $activeUsers = [];
+
         foreach ($users as $user) {
             if (! $user->isActiveEmployee) {
                 continue;
             }
-            $activeUser[] = $user;
+            $activeUsers[] = $user;
         }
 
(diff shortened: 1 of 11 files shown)


Modules/User/Entities/User.php
```diff
@@ -241,15 +241,20 @@ public function prospects()
     public function getActiveUsersAttribute()
     {
         $users = self::all();
```
pankaj-ag, line 243: Let's sort the users in the query itself. $users = self::query()->orderBy('name') instead of applying sorting later

pankaj-ag (changes requested):

Modules/User/Entities/User.php
Ayush8923: Update the changes. please check.

Modules/User/Entities/User.php
```diff
@@ -240,16 +240,17 @@ public function prospects()
 
     public function getActiveUsersAttribute()
     {
-        $users = self::all();
-        $activeUser = [];
+        $users = self::all()->sortBy('name');
```
pankaj-ag, line 243: This will also fetch the data first and then sort it. It's same thing you wrote previously

pankaj-ag (changes requested):

Modules/User/Entities/User.php
Ayush8923: Made the changes according to your suggestion.

pankaj-ag (approved):

PR #3763, opened 2025-04-30
Updated salary type on Payroll and Staff Type

P4NK4J:
Upon updating the payroll and staff type for an employee, the corresponding salary type was not getting updated. Hence, the UI displayed incorrect results. 

This PR removes the contractor-fee tag from the database when a contractor is changed to an employee in terms of payroll type

## Summary by CodeRabbit

- **New Features**
  - The current salary type for an employee is now automatically updated when staff type or payroll type changes are made in user settings.


Modules/User/Http/Controllers/UserSettingsController.php
```diff
@@ -22,11 +22,13 @@ public function update(Request $request)
...
         return redirect()->back()->with('status', 'Saved Successfully!');
     }
+
     public function addStaffType(Request $request)
     {
         $chosenEmployee = User::find($request->id)->employee;
         $chosenEmployee->staff_type = $request->typeOfStaff;
         $chosenEmployee->payroll_type = $request->payrollType;
         $chosenEmployee->save();
+        $chosenEmployee->updateCurrentSalaryType($chosenEmployee->payroll_type);
```
rathorevaibhav, line 32: do we need to pass `payroll_type`? as I guess it can be accessed within `updateCurrentSalaryType` using `$this->payroll_type`?

rathorevaibhav (approved):

PR #3854, opened 2026-06-02
fix(project): remove unused crashing fixed-budget-project command

rathorevaibhav:
## What & why

`project:fixed-budget-project` has been throwing a fatal on **every scheduled run** — it calls `ProjectService::getMailForFixedBudgetProjectKeyAccountManagers()`, which a prior merge-conflict resolution removed. The crash exists on both `main` and `develop` and has gone unnoticed, so the "fixed budget project about to end" reminder emails are not relied upon.

Rather than restore the feature, this removes the command and all code created for it. It also ends the recurring `production.ERROR` log noise.

## Removed
- `Modules/Project/Console/FixedBudgetProject.php` — the command
- `Modules/Project/Emails/FixedBudgetProjectMail.php` — the mailable
- `Modules/Project/Resources/views/mail/fixed-budget-project.blade.php` — the mail view
- `app/Console/Kernel.php` — the import, the `$commands` entry, and the schedule line

## Kept
- `ProjectService::getMailDetailsForProjectKeyAccountManagers()` — a **different** method used by the separate `project:ended-project` command.

## Verification
- `php artisan list` boots cleanly (exit 0); `project:fixed-budget-project` is no longer listed; other `project:*` commands intact.
- No remaining references to the command, mailable, or view anywhere in the codebase.
- PHP-CS-Fixer clean on the changed file.

Supersedes #3853 (which restored the command). A companion removal PR targets `develop` so the command can't reappear at the next release.


rathorevaibhav:
Closing as redundant. Consolidating to a single removal on `develop` (#3855), which will reach `main` via the normal production release PR — no separate hotfix needed.

PR #3861, opened 2026-06-10
Phase 0: upgrade safety baseline

Vipul-Ydv:
## Summary

Pre-upgrade cleanup before touching any Laravel version. Makes the codebase safe to upgrade by removing abandoned package dependencies, fixing CI, and establishing a test baseline.

- **Replace google-vision** — \`jordikroon/google-vision\` is abandoned (PHP 7 only, blocks PHP 8 upgrade). Replaced the single call site in \`BookServices::getISBN()\` with a direct Google Vision REST API call via Laravel's HTTP client. Same behaviour, same API key, no new dependency.
- **Replace corcel** — \`jgrossi/corcel\` has no Laravel 13 release (hard blocker for Phase 6). Decoupled all 4 call sites by introducing plain Eloquent models for the WordPress DB tables (\`app/Models/WordPress/Wp*.php\`) on the existing \`wordpress\` connection. \`WORDPRESS_ENABLED\` gate preserved throughout. Decision recorded: replace with plain Eloquent, not a fork.
- **Remove corcel + google-vision from composer.json** — Removed both dead packages (and 3 transitive dependencies). \`config/corcel.php\` deleted. Unblocked by temporarily setting \`config.audit.block-insecure=false\` during removal (reverted after).
- **Migrate root factories** — converted 3 live root factories to class-based (Laravel 8 style); deleted 3 dead ones targeting models that no longer exist.
- **Unify CI workflows** — bumped \`actions/checkout\` v2 → v4, \`codecov-action\` v3 → v4, added \`--no-interaction\` to composer installs.
- **Pin Node 24** — added \`.nvmrc\` + \`actions/setup-node@v4\` in all three workflows.
- **Add feature tests** — 16 new tests covering flows with zero coverage: Google OAuth login, Invoice access, Prospect CRUD, Salary access, EffortTracking tasks. Baseline: 136 tests, 230 assertions, all green on PHP 8.2.

## Deferred Phase 0 items (not in this PR)

These are on the Phase 0 exit checklist in \`docs/upgrade_plan.md\` but are deferred. They need a home before Phase 2 begins.

| Item | Reason deferred | Owner |
|------|----------------|-------|
| Remove \`laravel/legacy-factories\` + \`fzaninotto/faker\` | Need to audit all test files for direct \`Faker\...\` imports before removing | Phase 1 |
| Remove \`laravelcollective/html\`, \`consoletvs/charts\`, \`bordoni/phpass\` | Need usage audit + confirm zero references in Blade/JS | Phase 1 |
| Unify CI on PHP 8.1 (prove suite green on L8 + PHP 8.1) | Needs a standalone CI matrix change + local verification pass | Phase 1 |
| PDF stack decision (barryvdh/laravel-snappy, niklasravnsborg/laravel-pdf) | Needs team decision on which library to standardise on | Phase 1 |
| Staging DB seed script | Needs server access (Phase 1 infrastructure work) | Phase 1 |

## Test plan

- [x] CI passes on this PR (Cypress: 4/4, Unit: all green)
- [x] \`vendor/bin/phpunit\` — 136 tests, all green on PHP 8.2
- [x] \`jgrossi/corcel\` and \`jordikroon/google-vision\` removed from \`composer.json\` and \`composer.lock\`
- [ ] Confirm \`WORDPRESS_ENABLED=false\` default is untouched (no WP calls in normal flow)

## Summary by CodeRabbit

* **Chores**
  * Updated GitHub Actions workflows to use latest action versions (checkout v4, setup-node v4, codecov-action v4).
  * Updated Node.js version to 24.
  * Removed external WordPress integration library dependency.

* **Tests**
  * Added comprehensive feature tests for task management, salary access, invoicing, prospect management, and Google OAuth authentication.
  * Added unit tests for ISBN extraction functionality.
  * Enhanced test coverage with factory validation tests.

* **Refactor**
  * Modernized model factories to class-based implementation.
  * Replaced external WordPress library with native model layer for improved maintainability and control.


app/Models/WordPress/WpUser.php
```diff
@@ -0,0 +1,20 @@
...
+use Illuminate\Database\Eloquent\Model;
+
+class WpUser extends Model
+{
+    public $timestamps = false;
+    protected $connection = 'wordpress';
+
+    protected $table = 'users';
+
+    protected $primaryKey = 'ID';
```
rathorevaibhav, line 14: **Security issue, please fix before merge.**

The old corcel `User` model had `protected $hidden = ['user_pass'];`, so the WordPress password hash was never included when the model was converted to JSON. This new model hides nothing.

This matters because `Modules/User/Entities/User.php` (line 41) appends `websiteUser` to every user. So when `WORDPRESS_ENABLED=true`, any response that returns user data will now include the WordPress password hash. Also, the accessor loads this model `with('meta')`, so all usermeta is included too. WordPress stores session tokens in usermeta, so that data is sensitive as well. The old corcel code did not load meta, so none of this was exposed before.

Suggested fix:

```suggestion
    protected $primaryKey = 'ID';

    protected $hidden = ['user_pass', 'user_activation_key', 'meta'];
```

Hiding `meta` only affects JSON output. Code like `getWebsiteUserMeta()` can still read `$user->meta` normally.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -2,90 +2,69 @@
...
+        $term = WpTerm::select('term_id')->where('name', $job->domain)->first();
+        if ($term && $savedPost) {
+            $relation = new WpTermRelationship();
+            $relation->object_id = $savedPost->ID;
             $relation->term_taxonomy_id = $term->term_id;
             $relation->save();
         }
-        $job->opportunity_id = $post->ID;
-        $job->link = Option::get('siteurl') . $post->post_type . '/' . $post->post_name . '/';
+        $job->opportunity_id = $savedPost->ID;
```
rathorevaibhav, line 40: Small inconsistency: line 31 checks `$savedPost` before using it, but this line uses `$savedPost->ID` without a check. If the post lookup returns null, this line crashes. The old code had the same risk, so this is not a regression. But since the check now exists above, it is cleaner to return early when `$savedPost` is null, or guard both places the same way.

app/Services/BookServices.php
```diff
@@ -49,18 +47,24 @@ public static function getBookDetails($isbn)
     public static function getISBN($file)
     {
         $apiKey = config('constants.google.vision-api-key');
-        $vision = new Vision($apiKey, [new Feature(Feature::TEXT_DETECTION, 100)]);
-        $response = $vision->request(new LocalImage($file->path()));
-        $faces = $response->getTextAnnotations();
+
+        $response = Http::post('https://vision.googleapis.com/v1/images:annotate?key=' . $apiKey, [
```
rathorevaibhav: Behavior change to be aware of: the old Vision SDK threw an exception when the API call failed. `Http::post()` does not throw on 4xx or 5xx responses. So an auth or quota error now quietly returns an empty string, which looks the same as "no ISBN found". Suggest adding `->throw()` or at least a log line so real API failures stay visible.

app/Models/WordPress/WpTermRelationship.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace App\Models\WordPress;
+
+use Illuminate\Database\Eloquent\Model;
+
+class WpTermRelationship extends Model
+{
+    public $incrementing = false;
```
rathorevaibhav, line 12: Heads up for future readers: this table has a composite primary key (object_id + term_taxonomy_id) and no `id` column. The current usage is safe because rows are only inserted as new models or updated through the query builder. But calling `save()` or `delete()` on a fetched model instance would misbehave, because Eloquent would look for an `id` column. A short comment in the class explaining this would help the next person.

app/Models/WordPress/WpOption.php
```diff
@@ -0,0 +1,22 @@
...
+class WpOption extends Model
+{
+    public $timestamps = false;
+    protected $connection = 'wordpress';
+
+    protected $table = 'options';
+
+    protected $primaryKey = 'option_id';
+
+    public static function get(string $key, $default = null)
```
rathorevaibhav, line 18: Note: corcel's `Option::get()` also unserialized the value, because WordPress stores some options as serialized PHP. For `siteurl` this is fine since it is a plain string. If this helper is reused for other options later, it may need an unserialize step. A short comment here would be enough.

rathorevaibhav (changes requested):
Thanks for this PR. The core work here is good. The corcel and google-vision replacements behave the same as the old packages, the factory cleanup matches the upgrade plan exactly, and CI is green. I checked the new WordPress models against corcel in detail (table prefix, meta save and lookup, email lookup) and they match.

I am requesting changes for one security issue (see the inline comment on `WpUser.php`) and the points below.

## 1. Please update the PR description, it does not match the code

- The description says this PR fixed the integration-testing MySQL credentials. The last commit ("fix(ci): revert integration-testing DB to homestead credentials") undid that change, so the final code still uses the homestead database. Please remove or correct that bullet.
- The description says 16 new tests covering Google OAuth login, Invoice access, Prospect CRUD, Salary access, and EffortTracking tasks. The PR actually has 12 new tests, and there are no OAuth, Invoice, or Prospect tests in it.

## 2. This is the only Phase 0 PR, so the missing Phase 0 items need a home

These items are on the Phase 0 exit checklist in `docs/upgrade_plan.md` but are not in this PR. Since no other Phase 0 PR is planned, please either add them here or list them in the description as deferred work with a clear owner, so Phase 0 does not silently stay half done:

- **Add the missing tests for Google OAuth login, Invoice access, and Prospect CRUD.** The plan calls the test safety net the most important task in the whole plan. These flows still have no coverage.
- **Remove `jgrossi/corcel` and `jordikroon/google-vision` from composer.json.** This is the most important leftover. The code no longer uses them, but as long as google-vision stays in composer.json, Composer cannot resolve the dependencies on PHP 8.1. That means Phase 1 stays blocked even after this PR. Tip: try `composer remove --no-audit` to get past the advisory block mentioned in your notes. Also delete the leftover `config/corcel.php` when the package goes.
- **Remove `laravel/legacy-factories` and `fzaninotto/faker`.** After this PR they are fully unused. They are on the same Phase 0 checklist but the notes do not mention them.
- **Remove the three unused packages:** `laravelcollective/html`, `consoletvs/charts`, `bordoni/phpass`.
- **Unify CI on one PHP version.** The workflows still run 7.4 / 8.2 / 7.4. The plan also asks for the suite to be proven green on PHP 8.1 while still on Laravel 8.
- **The PDF stack decision and the staging DB seed** are also Phase 0 items with no home yet.

## 3. Branch model, for the next phases

The plan says `feature/laravel-upgrade` is a long-lived branch and each phase PR merges into it. This PR merges that branch into develop instead. For Phase 0 that is fine, because everything here is safe on Laravel 8. Two things to keep in mind:

- This merge uses up the branch name. It will need to be recreated for Phase 1 and later.
- Phase 2 and later (the real Laravel version bumps) must not target develop. develop flows to production before the full upgrade chain is proven on staging.

## What is done well

- All corcel code references are gone on this branch, and the de facto "decouple" choice resolves the corcel decision the plan asks for in Phase 0. This removes the Laravel 12 cap and keeps Laravel 13 reachable. Worth recording that decision in the description.
- The factory migration matches the plan exactly: 3 converted, 3 dead ones deleted, BookCategory split into its own factory, and the en_IN faker locale still works through `$this->faker`.
- Using a direct REST call for Google Vision instead of the heavy official SDK is a good call. It keeps the same API key auth and adds no new dependency, and the new tests fake the HTTP layer properly.

rathorevaibhav:
**About the CodeRabbit comments: some are real and need fixing, some should be skipped. Please do not apply all of them blindly.**

I checked each one against the code. Here is the breakdown.

## Please fix these in this PR

1. **`WpPostMeta` mass assignment (the one CodeRabbit marked Critical). This is a real crash, please fix.**
   I verified it by running the code. Laravel blocks mass assignment when a model has no `$fillable`, and the new `WpPostMeta` has none. So `WpPost::saveMeta()` throws `MassAssignmentException`. The old corcel meta model had `$fillable = ['meta_key', 'meta_value', 'post_id']`, which is why the old code worked. In practice this means creating or updating a job with `WORDPRESS_ENABLED=true` will crash. Tests did not catch it because WordPress is disabled in the test environment.
   Fix: add `protected $fillable = ['post_id', 'meta_key', 'meta_value'];` to `WpPostMeta`. The extra "unsaved post" guard in that comment is optional, since the code already saves the post first.

2. **`BookServices`: add a timeout and error handling.** Same point as my review comment. Add `Http::timeout(5)` and handle failure with a log or `->throw()`, so a Vision API error does not silently look like "no ISBN found".

3. **`WebsiteUserService`: add `auth()->check()` before reading the email.** The old code had the same problem, but the fix is one line and makes the service safe in artisan commands and queue jobs.

4. **`JobObserver` status mapping: add the `?? 'draft'` fallback.** Cheap and harmless. It avoids a null `post_status` when a status value is missing from the config map.

## Please skip these in this PR

The next three describe behavior the old corcel code also had. This PR is meant to behave exactly like the old code, and we cannot test these changes against the real WordPress database right now. Better to open one follow-up issue that lists them:

- **`term_taxonomy_id` set from `term_id`.** The old code did the same. On this WordPress version the two values are usually equal, which is why it works in practice. Changing it now is a behavior change with no way to verify.
- **Missing slash in the job link URL.** The old code built the URL the same way. Check the real `siteurl` option value on the WordPress site before changing this, otherwise we may create double slashes.
- **Deleting posts by title.** The suggested hr_id lookup is a real improvement, but it changes behavior, including deleting meta rows the old code left behind. Follow-up.

- **Codecov token:** skip for now, we will handle it separately.

app/Models/WordPress/WpUser.php
Vipul-Ydv: Fixed — added `protected $hidden = ['user_pass', 'user_activation_key', 'meta'];` to `WpUser`. Good catch, didn't notice corcel had `$hidden` on the old model. The meta hidden only affects JSON serialisation, so `getWebsiteUserMeta()` and the accessor still work normally.

Modules/HR/Observers/Recruitment/JobObserver.php
Vipul-Ydv: Fixed — added an early `return` when `$savedPost` is null right after the lookup, so both the `$term` block and the `$savedPost->ID` assignment are now guarded consistently.

app/Services/BookServices.php
Vipul-Ydv: Fixed — added `Http::timeout(5)` and a `$response->failed()` check that logs a warning with the status and body before returning early. A Vision API auth or quota error will now appear in the logs instead of silently looking like "no ISBN found".

app/Models/WordPress/WpTermRelationship.php
Vipul-Ydv: Added a comment explaining the composite PK situation and the safe/unsafe usage patterns.

app/Models/WordPress/WpOption.php
Vipul-Ydv: Added a comment noting that `siteurl` is a plain string so it works as-is, and that an `unserialize()` step may be needed if this helper is reused for other options later.

.github/workflows/unit-testing.yml
```diff
@@ -68,7 +73,7 @@ jobs:
           vendor/bin/phpunit --coverage-clover coverage.xml
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
```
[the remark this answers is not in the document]
Vipul-Ydv: Acknowledged — deferring to a separate task. Adding the token requires confirming whether a CODECOV_TOKEN secret is configured in the repo settings first, and that's outside the scope of this Phase 0 PR.

app/Models/WordPress/WpPost.php
```diff
@@ -0,0 +1,54 @@
...
+        });
+    }
+
+    public function saveMeta(string $key, $value): void
+    {
+        $this->meta()->updateOrCreate(
+            ['meta_key' => $key],
+            ['meta_value' => $value]
+        );
+    }
```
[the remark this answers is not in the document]
Vipul-Ydv: Fixed the mass assignment issue — added `protected $fillable = ['post_id', 'meta_key', 'meta_value'];` to `WpPostMeta`.

Skipping the unsaved-post guard per the reviewer's guidance — the code always calls `$post->save()` before `saveMeta()`, so `$this->ID` will always exist at that point.

app/Services/BookServices.php
```diff
@@ -49,18 +47,24 @@ public static function getBookDetails($isbn)
...
+            'requests' => [[
+                'image' => ['content' => base64_encode(file_get_contents($file->path()))],
+                'features' => [['type' => 'TEXT_DETECTION', 'maxResults' => 100]],
+            ]],
+        ]);
+
+        $annotations = $response->json('responses.0.textAnnotations') ?? [];
```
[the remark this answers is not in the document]
Vipul-Ydv: Fixed — added `Http::timeout(5)` and a `$response->failed()` check that logs a warning with the status code and response body before returning early. API errors will now be visible in logs instead of silently returning an empty string.

app/Services/WebsiteUserService.php
```diff
@@ -11,8 +11,7 @@ class WebsiteUserService
     public function __construct()
     {
         if (config('database.connections.wordpress.enabled')) {
-            $userProvider = new AuthUserProvider();
-            $this->user = $userProvider->retrieveByCredentials(['email' => auth()->user()->email]);
+            $this->user = WpUser::with('meta')->where('user_email', auth()->user()->email)->first();
         }
```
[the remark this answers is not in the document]
Vipul-Ydv: Fixed — added `&& auth()->check()` to the condition, so the email is only accessed when a user is actually authenticated. Safe to call from artisan commands and queue workers now.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -2,90 +2,69 @@
...
+        $post->post_status = config('hr.opportunities-status-wp-mapping')[$job->status];
+        $post->save();
+        $post->saveMeta('hr_id', $job->id);
+        $post->saveMeta(config('hr.slugs.job-form.key'), config('hr.slugs.job-form.value'));
+        $savedPost = WpPost::hasMeta('hr_id', $job->id)->first();
+        $term = WpTerm::select('term_id')->where('name', $job->domain)->first();
+        if ($term && $savedPost) {
+            $relation = new WpTermRelationship();
+            $relation->object_id = $savedPost->ID;
             $relation->term_taxonomy_id = $term->term_id;
```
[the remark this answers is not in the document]
Vipul-Ydv: Deferring — the old corcel code had the same mapping, and on this WordPress instance `term_id` and `term_taxonomy_id` are equal in practice (simple taxonomy setup), which is why it works in production. Fixing this correctly requires a `WpTermTaxonomy` model and a lookup join, which is a behaviour change we can't safely verify without the live WordPress DB. Will track as a follow-up issue.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -2,90 +2,69 @@
...
             $relation->save();
         }
-        $job->opportunity_id = $post->ID;
-        $job->link = Option::get('siteurl') . $post->post_type . '/' . $post->post_name . '/';
+        $job->opportunity_id = $savedPost->ID;
+        $job->link = WpOption::get('siteurl') . $savedPost->post_type . '/' . $savedPost->post_name . '/';
```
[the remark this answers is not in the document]
Vipul-Ydv: Deferring — the old corcel code built the URL identically (`Option::get('siteurl') . $post->post_type . '/' . $post->post_name . '/'`), so this is not a regression. Whether a slash is needed depends on whether the `siteurl` option value in the actual WordPress DB has a trailing slash or not. Changing this without checking could introduce a double-slash. Will verify against the live WP instance and fix as a follow-up.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -2,90 +2,69 @@
...
-            $corcel->post_status = $job->status ? (config('hr.opportunities-status-wp-mapping')[$job_status]) : 'draft';
-            $corcel->post_name = str_replace(' ', '-', strtolower($job->title));
-            $corcel->update();
-            $term = Term::select('term_id')->where(['name' => $job->domain])->first();
+        $savedPost = WpPost::hasMeta('hr_id', $job->id)->first();
+        if ($savedPost) {
+            $savedPost->post_title = $job->title;
+            $savedPost->post_content = $job->description;
+            $savedPost->post_type = config('hr.post-type.career');
+            $savedPost->post_status = $job->status ? config('hr.opportunities-status-wp-mapping')[$job->status] : 'draft';
```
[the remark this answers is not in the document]
Vipul-Ydv: Fixed — added `?? 'draft'` fallback in both `created()` and `updated()` so a missing config key defaults to draft instead of throwing an undefined key error.

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -2,90 +2,69 @@
...
-     * @param  \Modules\HR\Entities\Job $job
-     * @return void
-     */
     public function deleted(Job $job)
     {
         if (! config('database.connections.wordpress.enabled')) {
             return;
         }
-        Corcel::where(['post_type' => 'career', 'post_title' => $job['title']])->delete();
+        WpPost::where('post_type', 'career')->where('post_title', $job->title)->delete();
```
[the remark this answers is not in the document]
Vipul-Ydv: Deferring — the old corcel code deleted by title identically. The `hr_id` meta lookup is a genuine improvement but it changes cascade behaviour (the suggested fix also deletes associated meta rows, which the old code did not). This needs verification against the live WP DB to confirm no data integrity issues. Will track as a follow-up issue.

database/factories/ApplicantFactory.php
rathorevaibhav, line 1: any reason to remove these? I don't see any replacement for these. What's the purpose behind deleting them? Same for other factories that are removed

Modules/HR/Observers/Recruitment/JobObserver.php
```diff
@@ -2,90 +2,72 @@
...
-            $corcel->post_status = $job->status ? (config('hr.opportunities-status-wp-mapping')[$job_status]) : 'draft';
-            $corcel->post_name = str_replace(' ', '-', strtolower($job->title));
-            $corcel->update();
-            $term = Term::select('term_id')->where(['name' => $job->domain])->first();
+        $savedPost = WpPost::hasMeta('hr_id', $job->id)->first();
+        if ($savedPost) {
+            $savedPost->post_title = $job->title;
+            $savedPost->post_content = $job->description;
+            $savedPost->post_type = config('hr.post-type.career');
+            $savedPost->post_status = $job->status ? config('hr.opportunities-status-wp-mapping')[$job->status] : 'draft';
```
rathorevaibhav: Tiny leftover from the CodeRabbit feedback: this is the line its comment actually pointed at, and the array access here is still unguarded. `created()` got the `?? 'draft'` fallback but `updated()` did not.

```suggestion
            $savedPost->post_status = $job->status ? (config('hr.opportunities-status-wp-mapping')[$job->status] ?? 'draft') : 'draft';
```

Not blocking my approval. Fine to take now or in a later cleanup.

rathorevaibhav (approved):
All the requested changes are in, and I re-checked the whole PR from scratch. Approving.

## What I verified

**Security and crash fixes, all confirmed in the code:**
- `WpUser` now hides `user_pass`, `user_activation_key`, and `meta`. The password hash and usermeta no longer leak when users are serialized.
- `WpPostMeta` now has `$fillable = ['post_id', 'meta_key', 'meta_value']`. This fixes the `MassAssignmentException` crash in `saveMeta()`.
- `JobObserver` returns early when the saved post is not found, so no more null crash risk.
- `BookServices` now has a 5 second timeout and logs a warning when the Vision API fails.
- `WebsiteUserService` checks `auth()->check()` before reading the user email.

**Composer cleanup, confirmed:**
- `jgrossi/corcel` and `jordikroon/google-vision` are gone from composer.json and composer.lock, along with their 3 transitive packages (`thunderer/shortcode` and the 2 zend ones). `config/corcel.php` is deleted. No corcel references remain anywhere in the code.
- `fzaninotto/faker` correctly moved from production packages to dev packages in the lock, because corcel was the only production package that needed it.
- The temporary `audit.block-insecure` setting was reverted as stated. composer.json is clean.

**Tests, confirmed:**
- The OAuth, Invoice, and Prospect tests are now real and they run in the suite (phpunit.xml picks up `Modules/*/Tests` with a wildcard). I checked that the routes, the `finance_invoices.view` permission, and the `org_name` request field they use all exist.
- CI is fully green on the latest commit: unit tests, integration tests, Cypress (4/4), StyleCI, and coding standards.

**Also good:** the risky CodeRabbit suggestions (term_taxonomy_id, the URL slash, deleting posts by title) were correctly NOT applied. The old behavior is preserved, which is the right call for this PR. Those three stay as follow-up items.

## Two tiny non-blocking notes

1. The `?? 'draft'` fallback was added in `created()` but not in `updated()` (see inline comment). One line, can go in a later cleanup if you prefer.
2. The PR description says "16 new tests" but the count is now 22 test methods across 7 files, and the "136 tests" baseline number is also stale. Worth a quick number update since this body is the Phase 0 record.

And one optional thought for later, no action needed now: `WpPost::booted()` fills `post_date_gmt` and `post_modified_gmt` with local Asia/Kolkata time, not GMT. The old code left them as zeros, so this is not a regression, but if WordPress date display ever looks 5.5 hours off for synced jobs, this is why. `now('UTC')` for the two `_gmt` fields would make it exact.

Good work on this round. The fixes are exactly what was asked, nothing extra, nothing missed.

database/factories/ApplicantFactory.php
Vipul-Ydv: `App\Models\HR\Applicant` doesn't exist anywhere in the codebase — the actual model is `Modules\HR\Entities\Applicant`, which already has its own factory at `Modules/HR/Database/Factories/HrApplicantsFactory.php`. The deleted root factory was pointing at a non-existent class, so it was dead code that couldn't have been called. Same check was done for the other deleted factories before removing them.

Modules/HR/Observers/Recruitment/JobObserver.php
Vipul-Ydv: Good catch — fixed. `created()` had the `?? 'draft'` guard but `updated()` missed it. Applied the exact suggested change.

PR #3871, opened 2026-06-21
feat(phase-2): upgrade to Laravel 9

Vipul-Ydv:
## Summary

- Upgrades Laravel framework from 8 to 9 (`laravel/framework ^9.0`)
- Updates ecosystem packages: `laravel/ui ^4.0`, `nwidart/laravel-modules ^9.0`, `spatie/laravel-permission ^5.0`, `doctrine/dbal ^3.0`, `larastan/larastan ^2.0`, `nunomaduro/collision ^6.0`, `laravel-notification-channels/google-chat ^3.0`
- Replaces removed packages: `fzaninotto/faker` → `fakerphp/faker`, `facade/ignition` → `spatie/laravel-ignition ^1.0`, removes `fideloper/proxy` (now in framework)
- Moves `lang/` files from `resources/lang/` to root `lang/` (Laravel 9 convention)
- Rewrites `config/mail.php` for L9 mailer structure (`MAIL_DRIVER` → `MAIL_MAILER`)
- Updates `config/filesystems.php` (`FILESYSTEM_DRIVER` → `FILESYSTEM_DISK`)
- Fixes `TrustProxies` middleware to use L9 framework base class and individual header constants (removes deprecated `HEADER_X_FORWARDED_ALL`)
- Updates `.env.example` and `.env.testing` for renamed env vars
- Regenerates `composer.lock` against PHP 8.1 platform pin

## Test plan

- [x] `php artisan --version` returns `Laravel Framework 9.52.21`
- [x] All 136 PHPUnit tests passing (230 assertions)
- [ ] CI coding standards (PHP-CS-Fixer + ESLint)
- [ ] CI unit tests


Modules/Project/Database/Seeders/ProjectDatabaseSeeder.php
```diff
@@ -13,10 +13,10 @@
...
 class ProjectDatabaseSeeder extends Seeder
 {
-    const CLIENTS = 5;
-    const PROJECTS_PER_CLIENT = 2;
-    const MEMBERS_PER_PROJECT = 3;
-    const WORKDAYS = 10;
+    public const CLIENTS = 5;
+    public const PROJECTS_PER_CLIENT = 2;
+    public const MEMBERS_PER_PROJECT = 3;
+    public const WORKDAYS = 10;
```
rathorevaibhav, line 19: quick check - reason for making these public?

Vipul-Ydv: PHP CS Fixer's modifier_keywords rule (which replaced the deprecated visibility_required rule in .php-cs-fixer.php) requires explicit visibility on class constants — bare const fails the check. Making them public is the correct explicit declaration since that's what bare const means in PHP anyway.

rathorevaibhav: okay, but why other seeder files aren't updated with the same logic?

Also, what exactly is modifier_keywords rule and why is it required? I'm not sure why it is required at all.

Vipul-Ydv: No other seeder has class constants, so nothing else to update.

On modifier_keywords: we already had a rule that required explicit public/private/protected on all methods and properties. That rule got deprecated and its replacement (modifier_keywords) just extends the same logic to constants too. So const FOO = 5 now needs to be public const FOO = 5 — same idea, just consistent with how methods are already written.

rathorevaibhav (approved):

